Issue #1, opened 2024-04-16
Missing source


InsanityBringer:
After a few hours of working, I've gotten this compiling and partially working on Windows, but a chunk of the Windows specific code is missing. This is due to the template VS .gitignore including the following two rules:

```
[Ww][Ii][Nn]32/
[Ww][Ii][Nn]32/
```

This excludes the win32 directory. To compile it, I've rolled in the win32 sources as included in the editor source release and made a dummy version of the win32Controller class, but I'd like to be able to use something with consistent licensing. 

The .vsproj files are also missing, but I am unaware if the intended build system is CMake and the .sln is an old dev artifact. CMake seems to work fine.

kevinbentley:
Thanks. I [added the source from the tree](https://github.com/kevinbentley/Descent3/commit/ee8250371f6ee397e69b0af4d6a5a3a02d78cc6f) I had since I'm not sure how much it diverged from the editor code. We can always merge them together if it turns out something's missing.

And yes, we've been using CMake.

PR #6, opened 2024-04-16
Minimal changes for compilation (including linking) under Windows

Arcnor:
I was waiting for #1 to be done before submitting this, but this is the very minimum set of changes for the game to compile & link under Windows, at least with MSVC 2022 (32 bits mode only).

Surprising that the code works with almost no changes! Great job :).


kevinbentley:
Thank you!

PR #13, opened 2024-04-16
Formatting code with clang-format

### LINKED ISSUE #3: Formatting code with clang-format
Hi there.

After some digging in code I would suggest also run `clang-format` on all source files since there a lot code styles mixing and CRLF line endings.

winterheart:
Implementing clang-format on project to unify code style. As per #3 LLVM code style was selected, with some changes:

* ColumnLimit: 120 (column length increased to 120 symbols)
* SortIncludes: Never (don't sort includes as this may cause compilation failures)

Additionally converted most of files from non-ASCII encoding into UTF-8 and replaced CRLF line endings.

There also first GitHub Actions pipeline to improve QoL of future Descent 3 development. Each pull request or push into main branch will trigger clang-format check.

Fixes #3.

### DIFF
--- a/.github/workflows/clang-format-check.yml
+++ b/.github/workflows/clang-format-check.yml
@@ -1,0 +1,18 @@
+name: clang-format Check
+on:
+  pull_request:
+    branches:
+      - main
+  push:
+    branches:
+      - main
+jobs:
+  formatting-check:
+    name: Formatting Check
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


simonsan:
I'm not sure, but as it's 750.000 lines of changed code, that formatting should be made by a code owner. And only the .clang-format file should be commited via a PR. With the recent xz-utils thing, we would otherwise need to review 750k lines of code.

Copying from issue.

Issue #10, opened 2024-04-16
AngelScript compile error under MSVC


SiriusTR:
When building the project under MSVC (VS2022, using CMake/Ninja, x86 debug settings) I see this error under the AngelScript directory:
`scriptstring.cpp(596): error C2440: 'static_cast': cannot convert from 'void (__thiscall std::basic_string<char,std::char_traits<char>,std::allocator<char>>::* )(const unsigned int,const _Elem)' to 'void (__thiscall std::basic_string<char,std::char_traits<char>,std::allocator<char>>::* )(size_t)'`
This appears to be because MSVC's implementation of the STL implements both prototypes of basic_string::resize using a single method with an optional parameter... which I guess works, but it doesn't allow you to static_cast to the single-parameter version.

I notice that currently I can just remove AngelScript from the CMakeLists and then it's all fine, so I suppose it's not really in use yet. But I assume the intention is to eventually move there... would an appropriate solution be to update the version of AngelScript? Looks like this one is from 2009, so maybe a more current version has, erm, workarounds for the MSVC idiosyncrasy :)

kevinbentley:
I think it's just best to remove it for now. I don't know if we'll go the AngelScript direction, we never got very far in integrating it.

jcoby:
@kevinbentley PR added. It doesn't appear to be integrated with any of the other code and should be able to be removed cleanly.

Upgrading to the latest AngelScript (2.36.1) does fix the compile issues but it also requires C++11 if you'd rather go that direction.

PR #14, opened 2024-04-16
enable CI for building

Arcnor:
Very simple CI config, could be used to maybe check that #13 builds properly before merging. ~That said, AngelScript fails to build as mentioned in #10 (although for some strange reason I cannot reproduce it locally!) so the build is red, but still only a single failure should be expected for that PR.~

~It's easy to remove from the CMakeLists though, but I didn't want to generate conflicts with that PR just in case.~

Update: Build is green after rebasing: https://github.com/Arcnor/Descent3/actions/runs/8712259559

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,33 @@
+name: Descent 3 Build
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
+
+jobs:
+  build:
+    strategy:
+      fail-fast: false
+
+      matrix:
+        os: [windows-latest]
+        build_type: [Debug, Release]
+
+    runs-on: ${{ matrix.os }}
+
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,33 @@
...
+      fail-fast: false
+
+      matrix:
+        os: [windows-latest]
+        build_type: [Debug, Release]
+
+    runs-on: ${{ matrix.os }}
+
+    steps:
+    - uses: actions/checkout@v3
```
kpalmvik, line 21: I would suggest using the [modern `checkout?v4`](https://github.com/actions/checkout) based on a supported Node version, instead of the legacy v3.

That is unless there is a particular reason that I’m missing completely 🤗

Arcnor: Oh, no reason at all, just copied what I had around in some other project 🙂. I'll change it to v4

PR #18, opened 2024-04-16
Fix cf_SetSearchPath bad memory access (tested on macOS)

danra:


JeodC:
Can you rebase? Thanks!

Lgt2x:
Tested, it also fixes Linux-clang build. Thanks !

Issue #9, opened 2024-04-16
Collab with the enhanced macOS/Linux port?


jcotton42:
A few years back Ryan Gordon made some updates to the Linux and macOS versions of D3, but never brought those changes to the Windows version https://www.patreon.com/posts/project-descent-33611585.

It might be worth reaching out and seeing about integrating those changes. I don't know of a way to contact them outside of Patreon though.

kevinbentley:
Good catch. I'm pretty sure we have some conflicting changes and fixes in there. We had some interaction with Ryan back when Jeff and I worked on the 1.5 fixes, but I'd love to work with him to resolve everything and integrate it into the open source code.

icculus:
I need Interplay's permission to release my changes, as they own them; signing papers about this point was a condition of publishing the Mac/Linux builds on Steam. But I would definitely like to collaborate, as I have a bunch of stuff already solved, like 64-bit support, updated Linux and Mac stuff, Ogg Theora movies, etc.

Does Hervé Caen know about this repository? And is he going to be extremely upset if I bring it to his attention by asking him if I can publish my work too?

kevinbentley:
Parallax owns the copyright for Descent 3 and Matt Toschlog authorized the release of this code. I'm not sure what the legal situation with Interplay is, I know they only had the rights to publish and sell Descent 3, not ownership of the code. I don't know if that changed somewhere along the lines, but the last thing I've seen published on the matter is here: 
[https://www.gog.com/forum/general/descent_12_descent_3_removed_from_sale/post1420](https://www.gog.com/forum/general/descent_12_descent_3_removed_from_sale/post142), which states that Interplay has never owned the _copyright_, only the trademark. Then there's the whole dispute over their payment of royalties.

icculus:
What a mess.  :/

Okay, I'll email Hervé later today and see if he'll let me release my changes. It couldn't hurt to ask, right?

kevinbentley:
Also, bear in mind that this source does **not** include any of the game files, so a person would still need to purchase the game to play anything built with this code.

bryanperris:
Since some of the original copyrighted mission files provided their own custom compiled library, will the source behind those DLLs become open sourced?  If not, how will this project handle being backwards compatible?

kevinbentley:
The source to those should all be in the 'netgames' folder. For example:

https://github.com/kevinbentley/Descent3/tree/main/netgames/ctf

Is that what you mean?

[...]

Issue #4, opened 2024-04-16
Minimal C++ standard convention


winterheart:
Would C++14 ok here? Or maybe even C++17? Which standard proposed to use?

kevinbentley:
IMO, C++17 would be reasonable and I can't think of any reason why not to use it. But I'd like to see some others weigh in on this. Post what you think here please!

dsommers:
In the OpenVPN team, we've moved to C++17 ... Our OpenVPN 3 based code builds nicely on all our officially supported Linux distros (Debian, Ubuntu, RHEL8+, Fedora).  The community does builds on even more distros.

For RHEL-8, you would benefit from adding the gcc-toolset-10 (or newer) compiler.  After installing it, it's easy to enable it - just run `scl enable gcc-toolset-10 bash` and that terminal has the environment set up to use that compiler.

I would say C++17 is perfectly sane and reasonable to use these days.

Arcnor:
C++17 is reasonable, and even supported by the current and past generation of consoles (in case that's relevant either now or in the future) so that one has my vote as well.

JeodC:
> C++17 is reasonable, and even supported by the current and past generation of consoles (in case that's relevant either now or in the future) so that one has my vote as well.

Source ports making their way to homebrews like game consoles and even Linux-based retro handhelds and android phones is so common it's practically guaranteed.

kevinbentley:
Ok, let's officially say we support c++17. I'll update the readme soon, or if someone wants to submit a PR, feel free!

PR #20, opened 2024-04-16
Set C++ to 17, and a few CMake fixes for Linux

th1000s:


CMakeLists.txt
```diff
@@ -14,14 +14,19 @@ PROJECT(Descent3)
...
+    SET(EXTRA_CXX_FLAGS "-Wno-address-of-temporary")
+  ELSE()
+    SET(BITS "-m32")
+    SET(EXTRA_CXX_FLAGS "-fpermissive")
+  ENDIF()
+
   SET(CMAKE_CXX_COMPILER "g++")
-  SET(CMAKE_CXX_FLAGS "-O0 -g -Wno-write-strings -Wno-multichar -Wno-address-of-temporary")
-  SET(CMAKE_C_FLAGS "-O0 -g")
+  SET(CMAKE_CXX_FLAGS "-O0 -g -std=c++14 -Wno-write-strings -Wno-multichar ${BITS} ${EXTRA_CXX_FLAGS}")
```
winterheart, line 26: You need this done like this
```
set(CMAKE_CXX_STANDARD 14)
set(CMAKE_CXX_STANDARD_REQUIRED ON)
set(CMAKE_CXX_EXTENSIONS OFF)
```
But there some issues in with CMake compatibility level. Also, as agreed in #4, there should be C++17 standard support.

CMakeLists.txt
```diff
@@ -14,14 +14,19 @@ PROJECT(Descent3)
 IF (UNIX)
 SET (D3_GAMEDIR "~/Descent3/")
 
+  IF (APPLE)
```
winterheart, line 17: I doubt that in Apple there may 64bit code generated. Also, -m32 should be added via compiler toolchain, not directly in CMake project file.

winterheart (commented):
Overall, there need to CMake project get updated, as there some old tricks made in 2.x era. I'd suggest to rewrite CMakeLists.txt to 3.11 compatibility level to get rid some of these quirks.

Arcnor:
I'd actually wait to see the outcome of #9 before merging this and similar PRs, given that Ryan Gordon had a much more "modern" port for Linux and macOS, just my 2c.

CMakeLists.txt
th1000s: Done, plus a few lines of code adapted. And indeed, that is just functional cmake to get the code to compile.

CMakeLists.txt
th1000s: Yes, `-m32` did not quite work, removed entirely.

Lgt2x:
Merging, thanks for your contribution !

PR #21, opened 2024-04-16
Linux build fixes: remove unused osMutex code and typedefs

th1000s:
- Remove superfluous typedefs
- gcc: disable -Wnarrowing via pragma in snd8to16.h
- Remove linux/ipx.h include
- Remove unused osMutex code. It was never Acquire'd, only Created and Destroyed; and it was a stub on Linux and macOS.


Arcnor:
Hi, can you explain what the changes are? I see you removing the mutex code and some other things, but an explanation as to why will be nice :).

PR #22, opened 2024-04-16
Change case sensitivity

JeodC:
Converts all directories and files to lowercase except for top level files. May have conflicts with some recent commits--resolve manually. This was done via python script:

```python
import os

def convert_to_lowercase(directory):
    for root, dirs, files in os.walk(directory):
        for file in files:
            original_path = os.path.join(root, file)
            lowercase_path = original_path.lower()
            if original_path != lowercase_path:
                os.rename(original_path, lowercase_path)
                print(f"Renamed '{original_path}' to '{lowercase_path}'")
        for dir in dirs:
            original_path = os.path.join(root, dir)
            lowercase_path = original_path.lower()
            if original_path != lowercase_path:
                os.rename(original_path, lowercase_path)
                print(f"Renamed '{original_path}' to '{lowercase_path}'")

if __name__ == "__main__":
    script_dir = os.path.dirname(os.path.abspath(__file__))
    convert_to_lowercase(script_dir)
```

Ignore this PR if you want, it's a preference thing. Looking through the `scripts` directory was particularly annoying as there are `Levelxxx.cpp` and `levelxxx.cpp` files. ;)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,16 +12,24 @@ jobs:
       fail-fast: false
 
       matrix:
-        os: [windows-latest]
+        os: [{runner: windows-latest, preset: win}, {runner: macos-latest, preset: mac}]
         build_type: [Debug, Release]
 
-    runs-on: ${{ matrix.os }}
+    runs-on: ${{ matrix.os.runner }}
 
     steps:
     - uses: actions/checkout@v4
 
+    - name: Install macOS dependencies
+      if: ${{ matrix.os.preset == 'mac' }}
+      run: |
+        # Note this is x86_64 Homebrew
+        /usr/local/bin/brew install sdl12-compat sdl2_image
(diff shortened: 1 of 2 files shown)


JeodC:
Looks like #19 was committed after I forked, so would need re-committing ig.

JeodC:
I have added the missing commit to `.clang-format` and removed what appeared to be a tmp file `#.clang-format#`.

winterheart:
CMakeLists.txt also renamed which shouldn't be. I think main idea is rename project files, if so, there should be file mask on *.cpp, *.CPP, *.c, *.C, *.h, *.H.

JeodC:
> CMakeLists.txt also renamed which shouldn't be. I think main idea is rename project files, if so, there should be file mask on *.cpp, *.CPP, *.c, *.C, *.h, *.H.

Yes, this was quick n' dirty. If you'd like to do this along with your clang formatting, go ahead.

JeodC:
This is now fixed.

PR #23, opened 2024-04-16
Convert min/max macros to std::min/max

th1000s:
Added explicit `<float>` or `<int>` template argument where needed.
Add `-DNOMINMAX` define for Windows.

---
Commit message was:

These only cause trouble in stdlib headers, such as:

```
/usr/include/c++/12/limits:1828:13: error: expected unqualified-id before ‘noexcept’
 1828 |       max() _GLIBCXX_USE_NOEXCEPT { return __LDBL_MAX__; }
      |             ^~~~~~~~~~~~~~~~~~~~~
Descent3/d3music/musicapi.cpp: In function ‘void D3MusicSetVolume(float)’:
Descent3/d3music/musicapi.cpp:167:57: error: ‘std::numeric_limits<float>::min’ cannot be used as a function
  167 |   const float kEpsilon = std::numeric_limits<float>::min();
      |                          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~
```

plus there is the whole NOMINMAX situation on Windows.

And don't introduce `<algorithm>` just for that.

Automated via:

```shell
$ sed -i 's/_\?_\?min *(\(.*\),/D3_min(\1,/' `git grep -l min`
$ sed -i 's/_\?_\?max *(\(.*\),/D3_max(\1,/' `git grep -l max`
```

plus remove duplicates from lib/Macros.h + lib/linux/linux_fix.h and fix netgames/dmfc/dmfcstats.cpp (2x min on one line) manually.

----

Or, any other name, maybe just `min_` / `max_`?


kevinbentley:
Is there a reason not to use std::min? numeric_limits<float>::min is the minimum value a float can store, whereas std::min is the lesser of the two numbers.

winterheart:
Aggree, would be better use std::min / std::max and get rid min / max macros altogether. In code there also `_min`, `__min`, `_max`, `__max` variations.

th1000s:
> `numeric_limits::min()`

That might just be the pre-processor replacing stuff where it should not.  

So far I only saw `<string>` from the stdlib, so I thought I'd avoid `<algorithm>` which famously increases compile times.

However, moving to `std::max()`  triggers a *lot* of `no matching function for call to ‘max(int, float)’` (or `float`, `ushort`, `int`, `ubyte` in various combinations). For the time being renaming the macro is the quicker way to get Linux to compile.

winterheart:
Well, there need a casting into desired type. Since there anyway a lot of renaming, why don't go little further and use standardized functions?

th1000s:
Renaming the macro was mostly done to get the code to compile, small steps :)

Fully switching to `std::max()` is less automated and requires different casts -- `max(int, float)` should not become `max(static_cast<double>(a), static_cast<double>(b))`. And even with the correct types and just `float(a)` C-style casts, that makes the call sites a bit unwieldy.

Instead, maybe replace the macro with a custom `d3_max()` template using `std::common_type`?

kevinbentley:
You can specialize std::min like this:
```
int a = 123;
float b = 100.01;
float x = std::min<float>(a,b);
```
IMO, that's the best way to do it.

th1000s:
It is now templated where needed. `<float>` was only added (in a somewhat automated way) where at least one component was a `float` / `double` according to gcc, I also converted a few `1.0` to `1.0f` or added `<int>` manually.

Lgt2x:
@th1000s can you please rebase on master branch ?

PR #29, opened 2024-04-17
Mac CI fixes

danra:


Arcnor:
Hey, thanks for the PR!

There have been other changes in the meantime, so at least part of your PR won't work as is anymore (brew is now used on `main` due to another Mac PR being merged). Regarding the artifact name, that's great, but mac artifacts are still not being pulled either way, bigger changes are needed there.

And finally, is there a particular reason you want to have `ninja` specifically for macOS?

danra:
> There have been other changes in the meantime

Rebased

> Regarding the artifact name, that's great, but mac artifacts are still not being pulled either way, bigger changes are needed there.

How so? Worked fine yesterday in the branch CI https://github.com/kevinbentley/Descent3/actions/runs/8715829209

> And finally, is there a particular reason you want to have ninja specifically for macOS?

Just that it's faster than Xcode (which could also work)

Arcnor:
> > Regarding the artifact name, that's great, but mac artifacts are still not being pulled either way, bigger changes are needed there.
> 
> How so? Worked fine yesterday in the branch CI https://github.com/kevinbentley/Descent3/actions/runs/8715829209
> 
Not sure what changed there then, but even in the case of that previous build I don't think the artifacts are complete, it's missing the dylib for the scripts for example, although I suspect those should be static instead of shared, but we'll see.

> > And finally, is there a particular reason you want to have ninja specifically for macOS?
> 
> Just that it's faster than Xcode (which could also work)
Well, the current build is failing because of it, but besides that, the build seems to be fast enough as it is (it might be using `make` instead of `xcodebuild`, not sure). Can you remove the Ninja step for now, just to have a few less moving parts? We can add that at a later date I think.

danra:
> Well, the current build is failing because of it, but besides that, the build seems to be fast enough as it is (it might be using `make` instead of `xcodebuild`, not sure). Can you remove the Ninja step for now, just to have a few less moving parts? We can add that at a later date I think.

We need *some* multi-config generator (without changing the CI further). Make isn't one, and is the default. So we need to set a different a generator anyway. The only remaining delta between xcode and Ninja Multi-Config generators is the extra brew dependency, doesn't seem significant.

Arcnor:
> > Well, the current build is failing because of it, but besides that, the build seems to be fast enough as it is (it might be using `make` instead of `xcodebuild`, not sure). Can you remove the Ninja step for now, just to have a few less moving parts? We can add that at a later date I think.
> 
> We need *some* multi-config generator (without changing the CI further). Make isn't one, and is the default. So we need to set a different a generator anyway. The only remaining delta between xcode and Ninja Multi-Config generators is the extra brew dependency, doesn't seem significant.

I'm not sure we *need* a multi-config generator at all. The build runs in parallel in different agents, so having multi config is useless in this scenario.

danra:
> I'm not sure we need a multi-config generator at all. The build runs in parallel in different agents, so having multi config is useless in this scenario.

You're right, it's actually probably better to configure once (using a multi-build config), and build both configs, avoiding a matrix for Debug/Release in the first place.

Anyway, uploading the artifacts is set up (was set up also before I added macOS CI) to work with folder structure matching multi-config.

This works so let's merge it in. Can further tweak later.

PR #31, opened 2024-04-17
Committing local fixes and changes I had.

kevinbentley:


SiriusTR:
Regarding the Windows build failures - these seem to come from CMake not finding SDL - I notice we have an extra step to install it for Mac already in .github/workflows/build.yml so maybe this needs to happen on Windows too now. Unfortunately, I'm not familiar with the GitHub Actions API so I don't know how you'd go about doing that.

libmve/mvelibl.cpp
SiriusTR, line 1: This change is a bit odd - as discussed on Discord it looks like it might cause a crash.

Arcnor:
If we don't mind having `vcpkg` as a package manager, it's just a line or two to get SDL or basically any other supported package.

Otherwise there are GH actions to get SDL.

Arcnor:
It seems the SDL dependency is not needed on Windows yet (and the `sdl2_image` is not needed anywhere AFAIK), so it can just be removed from the Windows section on the CMakeLists and this PR will compile properly.

That said I've been working on adding vcpkg support and it already works fine, but currently there is no point until SDL2 is a requirement. I'll bring that up again after Ryan's Linux/Mac port is added, or that's rejected and a SDL2 port is started.

PR #39, opened 2024-04-17
Added MIT license.

xet7:
Added MIT license.


JeodC:
The initial MIT license was the default by github. Kevin wasn't sure it was the correct one to use so took it down.

PR #45, opened 2024-04-17
Enable Linux CI build

ei-grad:


ei-grad:
@kevinbentley can you approve CI trigger?

ei-grad:
Nevermind, I just pushed it into mine fork main branch and got some failures to fix :-).

Arcnor:
Hi, as mentioned in your other issue, there are a ton of PRs in flight, many doing what you're doing here already. Will it be possible to maybe contribute to those other PRs, or build on top of them, instead of having the same changes?

ei-grad:
@Arcnor This one builds and runs, and has Github Actions to check. If there is a PR with a similar progress, can you refer it here please?

Arcnor:
@ei-grad we have 20, 21, 23 and 33, besides this one. It's great, but also a bit overwhelming, specially with the size of the PR. The CI part will definitely help to test both this and the rest, so maybe you can open a PR for that separately, so we can merge before anything else?

th1000s:
Apart from `chmod +x hogUtils-i686` and the audio libs fix in cmake (I ended up removing more than just `esd`)  I also had to remove the 2 calls to `move(int, int)` in `linux/lnxcon.cpp`, otherwise I get linker errors. This function seems to be defined nowhere, right?

Lgt2x:
#23 is merged, you can rebase :)

Descent3/CMakeLists.txt
```diff
@@ -285,9 +285,8 @@ SET(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /SAFESEH:NO /SUBSYSTEM:WIN
 ENDIF()
 
 IF (UNIX AND NOT APPLE)
-SET (PLATFORM_LIBS linux dd_lnxsound ddvid_lnx lnxcontroller ddio_lnx ${SDL_LIBRARY} caca asound audio esd aa directfb m dl GLU)
+SET (PLATFORM_LIBS linux dd_lnxsound ddvid_lnx lnxcontroller ddio_lnx ${SDL_LIBRARY} caca asound audio aa directfb m dl GLU ncurses pulse-simple)
 SET (PLATFORM_CPPS loki_utils.c lnxmain.cpp)
-SET(CMAKE_EXE_LINKER_FLAGS "/usr/lib/libpulse-simple.so.0")
```
Lgt2x, line 290: Oh wow that's dirty

ei-grad:
> remove the 2 calls to  move(int, int) in linux/lnxcon.cpp, otherwise I get linker errors. This function seems to be defined nowhere, right?

It is from `curses.h`.

ei-grad:
Build fails on copying hog files to `D3_GAMEDIR`:
```
 - Creating Hog File... [Y2K.so] (99% done)
 - Done Creating Hog File.
Error copying file "/home/runner/work/Descent3/Descent3/scripts/data/linuxfullhog/d3-linux.hog" to "/home/runner/Descent3/".
gmake[2]: *** [scripts/CMakeFiles/FullLinuxHog.dir/build.make:129: scripts/CMakeFiles/FullLinuxHog] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:3809: scripts/CMakeFiles/FullLinuxHog.dir/all] Error 2
gmake: *** [Makefile:136: all] Error 2
```

Would it be ok to change its default value to something in build directory?

Arcnor:
It's probably better if `D3_GAMEDIR` is actually passed in as a variable to CMake, and removed from CMakeLists.

Lgt2x:
> It's probably better if `D3_GAMEDIR` is actually passed in as a variable to CMake, and removed from CMakeLists.

`set(D3_GAMEDIR "" CACHE PATH "Descent 3 game directory installation")` at the top of CMakeLists and setting the variable for each CI OS should be enough

ei-grad:
Lets deal with it in a separate PR? Currently it was failing just because ~/Descent3 directory was not created.

ei-grad:
Workflow is passing - https://github.com/ei-grad/Descent3/actions/runs/8728718390

.github/workflows/build.yml
```diff
@@ -2,17 +2,27 @@ name: Descent 3 Build
...
 jobs:
   build:
+
+    # In forks - all pushes and pull requests
+    # In main repo - only pushes and pull requests to the main branch
+    if: ( github.repository != 'kevinbentley/Descent3' )
+     || ( github.repository == 'kevinbentley/Descent3'
+          && ( github.ref == 'refs/heads/main'
+            || github.event_name == 'pull_request' ) )
+
```
winterheart, line 17: It's better to rework on top to
```
on:
  push:
    branches:
      - main
  pull_request:
    branches:
       - main
```
Don't limit CI on own forks.

Descent3/CMakeLists.txt
```diff
@@ -285,9 +285,8 @@ SET(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /SAFESEH:NO /SUBSYSTEM:WIN
 ENDIF()
 
 IF (UNIX AND NOT APPLE)
-SET (PLATFORM_LIBS linux dd_lnxsound ddvid_lnx lnxcontroller ddio_lnx ${SDL_LIBRARY} caca asound audio esd aa directfb m dl GLU)
+SET (PLATFORM_LIBS linux dd_lnxsound ddvid_lnx lnxcontroller ddio_lnx ${SDL_LIBRARY} caca asound audio aa directfb m dl GLU ncurses pulse-simple)
```
winterheart, line 288: Is there really needed external libraries like caca asound audio aa directfb? Seems this is redundant.

.github/workflows/build.yml
ei-grad: I don't understand, this change is actually needed to run CI on own forks in all branches, not limited to main. `on` mechanics doesn't allow to switch logic. With `on` we only can make CI to run on all branches in origin repository too, though. My intention was just to make it run CI in linux-build branch on my fork, and don't change behavior in upstream. It is not needed actually, anymore, so I can remove this change from PR.

winterheart: Yes, please remove it.

PR #46, opened 2024-04-17
Fix incorrect compilation flag append on Linux

danra:


Arcnor:
Hi, I think we actually want to remove the whole thing, this kind of flag should be set outside and not from the CMakeLists file.

PR #50, opened 2024-04-17
Rename artifacts

JeodC:
Artifacts are currently all named build-Debug and build-Release. Include the platform in the name.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,7 +10,6 @@ jobs:
   build:
     strategy:
       fail-fast: false
-
       matrix:
         os: [{runner: windows-latest, preset: win}, {runner: macos-latest, preset: mac}]
         build_type: [Debug, Release]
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -35,5 +35,5 @@ jobs:
     - name: 'Upload Artifacts'
       uses: actions/upload-artifact@v4
       with:
-        name: build_${{matrix.build_type}}
+        name: build_${{matrix.platform}}_${{matrix.build_type}}
```
Jayman2000, line 38: It looks like it didn’t work. [The CI logs say](https://github.com/kevinbentley/Descent3/actions/runs/8729169956/job/23950607961?pr=50#step:6:18):

```
Artifact build__Debug has been successfully uploaded! Final size is 5014487 bytes. Artifact ID is 1423817003
```

Shouldn’t it say “build_windows_Debug” or something like that?

Jayman2000 (changes requested):

.github/workflows/build.yml
JeodC: Yeah, let me try that again.

PR #52, opened 2024-04-17
Update procedurals.h

MaddTheSane:
Quiet a warning about header guard not matching the defined macro.


JeodC (commented):
This looks like it changes nothing? Or am I blind?

Descent3/procedurals.h
JeodC, line 1: Is it me or does the diff change nothing?

MaddTheSane:
It swaps two characters, from "RU" to "UR"

JeodC:
Well I'll be.

PR #54, opened 2024-04-18
Fixes and enhancements to CMake build system

winterheart:
Updated compatibility level to 3.19. Set C++17 globally for all platforms. Removed hardcoded compiler and optimisation flags. Adjusted dependencies and libraries linking.

Reworked script building and hog creation.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,39 +1,35 @@
-# CMake compatibility issues: don't modify this, please!
-CMAKE_MINIMUM_REQUIRED( VERSION 2.4.6 )
+CMAKE_MINIMUM_REQUIRED(VERSION 3.19)
 
-MARK_AS_ADVANCED(CMAKE_BACKWARDS_COMPATIBILITY)
 # allow more human readable "if then else" constructs
 SET( CMAKE_ALLOW_LOOSE_LOOP_CONSTRUCTS TRUE )
 
+PROJECT(Descent3 VERSION 1.5.500)
 
-PROJECT(Descent3)
+SET(CMAKE_CXX_STANDARD 17)
+SET(CMAKE_CXX_STANDARD_REQUIRED ON)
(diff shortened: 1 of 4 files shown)


scripts/CMakeLists.txt
```diff
@@ -65,29 +65,23 @@ SET (SCRIPTS
...
-
-    ENDIF()
-
+	ADD_LIBRARY(${SCRIPT} SHARED ${CPPS} "${SCRIPT}.cpp")
+	set_target_properties(${SCRIPT} PROPERTIES PREFIX "")
+	IF(UNIX)
+		ADD_CUSTOM_COMMAND(
+			TARGET ${SCRIPT}
+			POST_BUILD
+			COMMAND ${CMAKE_COMMAND} -E copy $<TARGET_FILE:${SCRIPT}> "${CMAKE_SOURCE_DIR}/scripts/data/linuxfullhog/"
```
Arcnor, line 76: I think we'll have to revisit this, copying into the source directory is not very clean, this should probably be the binary one, then some extra code to deal with whatever needs to happen to those files later (install them?)

CMakeLists.txt
```diff
@@ -1,39 +1,35 @@
...
 
 IF (UNIX)
-SET (D3_GAMEDIR "~/Descent3/")
-
-  SET(CMAKE_CXX_STANDARD 17)
-  SET(CMAKE_CXX_STANDARD_REQUIRED ON)
-  SET(CMAKE_CXX_EXTENSIONS OFF)
-  SET(CMAKE_POSITION_INDEPENDENT_CODE ON)
-
+  SET (D3_GAMEDIR "~/Descent3/")
```
Lgt2x, line 14: D3_GAMEDIR should be a cache variable set by the user

scripts/CMakeLists.txt
winterheart: Yeah, I was thinking on it. Will rework this some day.

PR #56, opened 2024-04-18
Add inverted flight sim mode

zardoru:
Mouselook doesn't quite preserve inertia, and feels jarring to use as a non inverted-Y user. This PR fixes that, preserving the old mouselook mode and adding a new "Inverted Flight Sim" mode. The video illustrates the differences between current mouselook and the new mode.

Problem is that by adding it as a new option it needs to do something about the strings being included with the game data. Oops.

https://github.com/kevinbentley/Descent3/assets/391382/bc8a1e67-8b64-4380-a54d-a8f02556caca

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,7 @@ bld/
 [Oo]bj/
 [Ll]og/
 [Ll]ogs/
+out/
 
 # Visual Studio 2015/2017 cache/options directory
 .vs/
--- a/Descent3/ctlconfig.cpp
+++ b/Descent3/ctlconfig.cpp
@@ -1246,8 +1246,9 @@ void joystick_settings_dialog() {
   sheet->NewGroup(TXT_MOUSECONTROL, 210, y);
   msectl = sheet->AddFirstLongRadioButton(TXT_MOUSEFLIGHTSIM);
   sheet->AddLongRadioButton(TXT_MOUSELOOK);
-  *msectl = Current_pilot.mouselook_control ? 1 : 0;
-  y += 50;
+  sheet->AddLongRadioButton(TXT_MOUSEFLIGHTSIM_INVERTED);
+  *msectl = Current_pilot.mouselook_control;
+  y += 75;
   sheet->NewGroup(NULL, 210, y);
   sheet->AddLongButton(TXT_CALIBJOYSTICK, UID_JOYCFG);
   // force feedback stuff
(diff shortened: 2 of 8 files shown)


JeodC:
Quote from @InsanityBringer via Discord:
> all strings that aren't a hardcoded string literal are from the game data
it's not out of the realm of possibility to throw in a little program that makes hog3 files, and use that with the build system to build another extra hog that gets used by the engine
where new strings can be put
but atm strings are referenced by their index so it would get confusing fast. a key/value system might help
.. and this is presuming we get people who want to localize strings. it's not out of the realm of possibility, gzdoom supports many langs now but.. that's gzdoom
I think descent is just slightly less popular than doom overall. slightly

This will need to be addressed down the road.

Jayman2000:
What’s the difference between using this new inverted flight sim mode and [using the original flight sim mode with an inverted Y-axis](https://support.gog.com/hc/en-us/articles/4407602848913-Descent-3-Mercenary-How-to-invert-controls?product=gog)?

zardoru:
> What’s the difference between using this new inverted flight sim mode and [using the original flight sim mode with an inverted Y-axis](https://support.gog.com/hc/en-us/articles/4407602848913-Descent-3-Mercenary-How-to-invert-controls?product=gog)?

I sincerely didn't even know you could work around that like that. Guess the PR's redundant?

Jayman2000:
> I sincerely didn't even know you could work around that like that.

Yeah, I tolerated this problem for like ten years before learning that you could invert the Y-axis like that. That option is too hidden.

zardoru:
I do think there's a benefit to having an option like this, since it is significantly more obvious than going to the mouse and inverting the pitch controls, but I understand if this gets closed.

Jayman2000:
> I do think there's a benefit to having an option like this, since it is significantly more obvious than going to the mouse and inverting the pitch controls, but I understand if it's better to close this.

I agree. You’ve identified a really significant problem with Descent 3. The option for inverting the Y-axis should not be so hard to find. It’s just that your implementation doesn’t make sense because it’s redundant and adds unnecessary complexity.

I just opened #62 to make sure that this problem gets fixed, even if this PR isn’t the one that ends up fixing it.

JeodC:
> > I do think there's a benefit to having an option like this, since it is significantly more obvious than going to the mouse and inverting the pitch controls, but I understand if it's better to close this.
> 
> 
> 
> I agree. You’ve identified a really significant problem with Descent 3. The option for inverting the Y-axis should not be so hard to find. It’s just that your implementation doesn’t make sense because it’s redundant and adds unnecessary complexity.
> 
> 
> 
> I just opened #62 to make sure that this problem gets fixed, even if this PR isn’t the one that ends up fixing it.

Thanks. It may be a while, but it's on the board at least.

PR #58, opened 2024-04-18
Use __aarch64__ instead of __arm64__ for Linux builds

mstephenson6:
- Fixes a build error on M1 Linux, `unknown mnemonic int -- int $3`, in mveasm.cpp


Arcnor:
This also affects macOS, I'm assuming it works because `__aarch64__` is supposed to be defined there too, but if somebody can test (@jcoby ?) that'd be great.

PR #59, opened 2024-04-18
Fix HOG file entry parser in 64 bit architectures

Lgt2x:
timestamp is 4 bytes in the file, but is read as an 8-bytes long when built on 64 bit architectures, screwing up the read offset


Arcnor:
We should probably make this a bit more obvious, with `uint32_t` from `cstdint` for example.

PR #60, opened 2024-04-18
enable both Intel & M1 macOS runners

Arcnor:
GH docs seem to be lying to us, and `macos-latest` is actually giving us `macos-12` according to our previous runs, so do that on purpose but also do `macos-14` so we have M1 (in theory)


Lgt2x:
Thanks!

PR #69, opened 2024-04-18
Add Linux and D3 specific .gitignore entries

GravisZro:
Ignore these:
`build/` (CMake build directory)
`*.so` (Linux)
`*.hog` (D3)


winterheart:
We have plans to move generation of *.so and *.hog files into build directory, but for now that will ok.

PR #70, opened 2024-04-18
Fix Linux clang build

Lgt2x:
- UIEdit: set char to 0 instead of NULL
 - Also apply -Wno-address-of-temporary to Linux


Lgt2x:
don't merge just yet, it seems to crash in-game when gcc does not

Lgt2x:
> don't merge just yet, it seems to crash in-game when gcc does not

We're good, it just needed #18

PR #72, opened 2024-04-18
Ensure STDCALL/STDCALLPTR is only invoked on x86

GravisZro:
Prevent the code from trying to use stdcall for anything except (32-bit) x86 builds because it's the only platform it's applicable for.

This clears up a lot of annoying warning messages.

Alteration to CMakeList.txt is only whitespace clean up.

### DIFF
--- a/lib/module.h
+++ b/lib/module.h
@@ -76,17 +76,24 @@
 #define CPPEXTERN extern
 #endif
 
+#ifdef _MSC_VER // MSVC Build
+# define STDCALL __stdcall
+# define STDCALLPTR *STDCALL
+#else // Non-MS compilers
+# if defined(__i386__) || defined(_X86_) || defined(__THW_INTEL__)
+#  define STDCALL __attribute__((stdcall))
+# else
+#  define STDCALL
+# endif
+# define STDCALLPTR STDCALL *
+#endif
+
+# define DLLFUNCCALL STDCALL
+# define DLLFUNCCALLPTR STDCALLPTR
+
 #ifdef WIN32
 //=========================Windows Definition============================
 #include "windows.h"
-
-#ifdef _MSC_VER // Visual C++ Build
-#define DLLFUNCCALL __stdcall
-#define DLLFUNCCALLPTR *DLLFUNCCALL
-#else // Non-Visual C++ Build
-#define DLLFUNCCALL __attribute__((stdcall))
-#define DLLFUNCCALLPTR DLLFUNCCALL *
-#endif
 
 #define MODPROCADDRESS FARPROC
 #define DLLFUNCEXPORT __declspec(dllexport)
@@ -101,8 +108,6 @@ typedef struct {
 #include <dlfcn.h>
 
 #define MODPROCADDRESS void *
-#define DLLFUNCCALL __attribute__((stdcall))
-#define DLLFUNCCALLPTR DLLFUNCCALL *
 #define DLLFUNCEXPORT
 #define DLLFUNCIMPORT
 #define DLLEXPORT CPPEXTERN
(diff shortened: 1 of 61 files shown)


CMakeLists.txt
```diff
@@ -1,15 +1,23 @@
-CMAKE_MINIMUM_REQUIRED(VERSION 3.19)
```
th1000s, line 1: Maybe place these cmake changes into a separate commit?

CMakeLists.txt
```diff
@@ -1,15 +1,23 @@
-CMAKE_MINIMUM_REQUIRED(VERSION 3.19)
+# CMake compatibility issues: don't modify this, please!
+CMAKE_MINIMUM_REQUIRED( VERSION 2.4.6 )
```
Lgt2x, line 2: Why do you want to go back ?

JeodC: Looks like it failed to build otherwise?
https://github.com/kevinbentley/Descent3/actions/runs/8742874968/job/23992131506

CMakeLists.txt
```diff
@@ -10,6 +10,7 @@ SET(CMAKE_CXX_STANDARD_REQUIRED ON)
 SET(CMAKE_CXX_EXTENSIONS OFF)
 SET(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
+include_directories(.)
```
winterheart, line 13: You need use ${CMAKE_SOURCE_DIR} instead that.

lib/module.h
```diff
@@ -58,18 +58,40 @@
...
+        defined(__i486__) || \
+        defined(__i586__) || \
+        defined(__i686__) || \
+        defined(_X86_) || \
+        defined(__X86__) || \
+        defined(__THW_INTEL__) || \
+        defined(__I86__) || \
+        defined(__IA32__)
+# define X86_32
+#endif
```
winterheart, line 74: Dunno. Looks like big ugly hack there.

scripts/AIGame.cpp
```diff
@@ -8,13 +8,7 @@
...
 #include "DallasFuncs.cpp"
 
-#ifdef _MSC_VER // Visual C++ Build
-#define STDCALL __stdcall
-#define STDCALLPTR *STDCALL
-#else // Non-Visual C++ Build
-#define STDCALL __attribute__((stdcall))
-#define STDCALLPTR STDCALL *
-#endif
+#include <lib/module.h>
```
winterheart, line 11: Use `#include "header.h"` (quotes) instead of `#include <header.h>`.

winterheart:
Why just not remove STDCALLs entirely? Is there low-level magic that may prevent it runs in 32bit mode?

GravisZro:
@winterheart x86 calls had several different forms and without specifying how an exposed library function should be structured your compiler may end up structuring or calling that library function incorrectly.

GravisZro:
@winterheart 
You can learn more about why stdcall is needed here: https://en.wikipedia.org/wiki/X86_calling_conventions

Lgt2x:
I'll let others review this one, I'm not very comfortable with x86/win32 subjects :)

GravisZro:
Basing the architecture on `CMAKE_SIZEOF_VOID_P` is a bad idea because **ANY** 32-bit build will trigger this, be it x86, ARMv7, PPC.

CMake has a way to reporting the architecture but it's OS specific and things really get funky when cross compiling. It also makes the code entirely dependent on CMake build system which seems likely since the names for target architectures are OS dependent.

As for leaving the STDCALL code "as is", doing so will result in build failures for non-x86/x86_64 targets.

The solution I put forth works with all compilers and build systems. You called it a "big ugly hack" but it's not a hack at all because this kind of problem is precisely why compilers define these macros. It's the best solution as it supports a multitude of compilers and has no regard for the host OS, target OS, or build system being used. See also: https://sourceforge.net/p/predef/wiki/Architectures/

If you're still unsure then please just take the word of someone with more experience in targeting specific platforms and architectures that this is the proper solution.

GravisZro:
@winterheart, I have revised the code to exclude non-C++17 compilers and removed duplicate detection macros. I hope this is now sufficiently terse.

PR #74, opened 2024-04-18
Fix simple warnings

GravisZro:
Most of these warnings are due to the use of NULL instead of 0 or NULL instead of '\0'. Some are macro redefinitions. None of them are pointer storage related. Those will be in another PR.


Arcnor:
Can you use `nullptr` instead of `\0`?

GravisZro:
Neither `NULL` not `nullptr` are characters. `'\0'` is a character, so no, I cannot use `nullptr`.

Arcnor:
Oh, this is assigning null terminators to strings, I see. Not sure why there are so many instances of that, though.

winterheart:
Is there more secure way to do that with `memset`?

GravisZro:
@winterheart  Honestly, all the string building/manipulation code is obsolete garbage that needs to be replaced. At this point, I'm merely clearing up warning messages so that I can make a sane build **before** making major changes.

Lgt2x:
I still have some `implicit conversion of NULL constant to 'char'` in some places, but certainly fewer. Thanks!

PR #78, opened 2024-04-18
Change most sprintf to snprintf

MaddTheSane:
Clang/Xcode complains when using `sprintf` and `vsprintf` because they are insecure. This patch replaces a lot of `sprintf` calls to `snprintf` or, in rare cases, with `strcpy`.


GravisZro:
Awesome. I hope you are doing a PR for other "n" functions. `strcpy` is used a bunch instead of `strncpy`.

shermp:
> Awesome. I hope you are doing a PR for other "n" functions. `strcpy` is used a bunch instead of `strncpy`.

I think more care needs to be taken to replace `strcpy` with `strncpy`, as `strncpy` does not guarantee null termination.

Lgt2x:
Needs rebase as well

MaddTheSane:
rebased

linux/lnxcon.cpp
```diff
@@ -148,16 +148,9 @@ void con_Printf(const char *fmt, ...) {
...
   //	filter out messages
   va_start(args, fmt);
-  vsprintf(buf, fmt, args);
+  vsnprintf(buf, sizeof(buf), fmt, args);
   va_end(args);
 
-  int len = strlen(buf);
-  if (len >= CON_MAX_STRINGLEN) {
-    // we overflowed our buffer!!!
-    // we need to do some sort of error here!!!!
```
Lgt2x, line 157: lol

Lgt2x (approved):

Lgt2x:
Good job, look fine

PR #80, opened 2024-04-18
Accurately find the linux artifacts

JeodC:
This workflow excludes the build_type matrix when searching for the linux artifact path.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -35,7 +35,7 @@ jobs:
           sudo dpkg --add-architecture i386
           sudo apt update
           sudo apt install -y --no-install-recommends \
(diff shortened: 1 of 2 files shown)


JeodC:
I think Linux does not use Debug and Release, correct? So building twice for Linux is redundant?

th1000s:
Release / Debug in cmake also adds `-g` or `-O` flags on Linux, even `-O3` at the moment. Some warnings are only triggered on a certain optimization level.

But it might be interesting to build with `clang` on Linux as well.

winterheart:
```
            "generator": "Ninja Multi-Config",
```
For linux in CMakePresets.json unifies all builds, better using this.

JeodC:
This should be ok to squash and merge now.

PR #81, opened 2024-04-18
Add clang-format cmake targets and formatting script

th1000s:
fmt: formats all modified files (what `git status` lists)
fmtall: all files known to git
checkfmt: checks that all files are formatted correctly

Warn if the version does not match the one specified by the GitHub action.

---
(the removed version looked at every file, and also truncated them because of `$file > $file`)


Arcnor:
The problem is that the formatter causes issues (there was a PR to fix invalid code generated by the formatter) so I think this should be rethought a bit, maybe only apply to newly added lines or similar...

PR #85, opened 2024-04-19
Fix endian read-write for doubles.

MaddTheSane:


Lgt2x:
Nice catch, thanks!

Lgt2x:
I guess you have conflicts to solve before we can merge

PR #86, opened 2024-04-19
Remove D3_GAMEDIR hard-coded CMake variable

Lgt2x:
Proper standalone installation will come later


GravisZro:
Since not everyone is skilled with cmake, it should have a fallback value if not set via commandline.

Lgt2x:
> Since not everyone is skilled with cmake, it should have a fallback value if not set via commandline.

The plan has changed, we're postponing reworking the install procedure until we do proper packaging. Until them, we'll provide instructions to the user to copy the generated executable to the right location

Lgt2x:
We should be ready for review now. This just gets rid of the default D3_GAMEDIR

Arcnor (approved):
Only thing that I might change is not removing the newline at the end of every CMakeList file, but other than that just merge whenever you want, looks fine 🙂

Lgt2x:
> Only thing that I might change is not removing the newline at the end of every CMakeList file, but other than that just merge whenever you want, looks fine 🙂

Not sure what my editor did there, but it's fixed now

Issue #88, opened 2024-04-19
[MacOS] Missing Briefing.fnt


JeodC:
Reported by user meh and arcnor.

MacOS (at least m1) encounters an error regarding Briefing.fnt on run attempt. This did not occur at [1edaa3b](https://github.com/kevinbentley/Descent3/commit/1edaa3b628b6945f93f7fde453967e244fabb45b).

Bisecting is in progress.

JeodC:
First bad commit at #76 in [7bfea72](https://github.com/kevinbentley/Descent3/pull/76/commits/7bfea728b843ba532129604b7933cb0c9a7415e5)

PR #91, opened 2024-04-19
Re-apply cfile-update with fix regression

winterheart:
After merging #76 there was regression found in #88. This PR re-applies changes with additional fix.


CMakeLists.txt
```diff
@@ -69,7 +69,7 @@ ENDIF ()
 
 MESSAGE("Install will copy files to ${D3_GAMEDIR}")
 
-include_directories("lib" "Descent3" ${PLATFORM_INCLUDES})
+include_directories("${CMAKE_SOURCE_DIR}" "lib" "Descent3" ${PLATFORM_INCLUDES})
```
Lgt2x, line 72: This is not a great idea to include the root source dir. If you made this change to find cfile.h, then include `cfile` directory directly

winterheart:
Here additional regressions I've found in 32bit mode...

winterheart:
PR ready to merge, though additional testing on Windows would be useful.

CMakeLists.txt
Arcnor: I agree here, but not sure why this comment was marked as resolved?

cfile/CMakeLists.txt
```diff
@@ -1,7 +1,6 @@
...
-		CFILE.cpp
-		hog.cpp
-		InfFile.cpp)
+	cfile.cpp
+	hogfile.cpp
+	inffile.cpp)
 
-add_library(cfile STATIC ${HEADERS} ${CPPS})
\ No newline at end of file
+add_library(cfile STATIC ${CPPS})
```
Arcnor: To avoid adding the root of the project to find the "cfile/cfile.h" include, you can do:

1. Revert all the includes to be `#include <cfile.h>` instead of `#include "cfile/cfile.h"`
2. Here have `target_include_directories(cfile PUBLIC .)`. That will make any dependent of the `cfile` library find its includes.

Arcnor (changes requested):

cfile/CMakeLists.txt
winterheart: Main idea about it is splitting code into manageable submodules that can be accessed from root of project. Maybe ${CMAKE_SOURCE_DIR} is not good for it, but ${PROJECT_SOURCE_DIR} would better?

Arcnor: No, I'd still depend on the library to provide the includes instead of a blank `include_directories` (which I'd remove completely as well in favor of `target_include_directories` for each, but that's a separate discussion), I think that's the correct semantic for that (you depend on the assets of the `cfile` library, so it makes sense you add a dependency on it)

winterheart: I've added BUILD_INTERFACE to cfile module. Other modules that depends on it, can reuse includes on linking.
There some files formally not belonging any packages (lib directory), as workaround there temporary include_directories(cfile) on root of project. After migrating all modules this can be removed.

cfile/hogfile.cpp
```diff
@@ -408,30 +292,32 @@ int CreateNewHogFile(const char *hogname, int nfiles, const char **filenames, vo
...
   fclose(hog_fp);
   delete[] table;
-  // Setup the update message and send it
-  char msg[256];
-  snprintf(msg, sizeof(msg), "Done Creating Hog File.");
-  if (UpdateFunction != NULL)
+  if (UpdateFunction != nullptr) {
+    // Setup the update message and send it
+    char msg[256];
+    snprintf(msg, sizeof(msg), "Done Creating Hog File.");
```
Arcnor, line 319: I know this is old code, but... why `snprintf`? This can just simply be `const char* msg = "Done Creating Hog File."` unless I'm missing something :D.

winterheart: This message was used for callback function, so there no need to allocate/deallocate char * array on caller side. Or, maybe I'm missing something too :).

2dlib/CMakeLists.txt
```diff
@@ -10,4 +10,6 @@ set(CPPS
   viewport.cpp)
 
 add_library(2dlib STATIC ${HEADERS} ${CPPS})
-
+target_link_libraries(2dlib
+  cfile
```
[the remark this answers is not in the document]
winterheart: Done.

cfile/hogfile.cpp
Arcnor: What I'm saying is, this message can be a non-allocated string, just a compile time one instead. But no need to change that, it's just silly but doesn't break anything.

Arcnor (approved):

PR #92, opened 2024-04-19
Add Git-Head to version string

### LINKED ISSUE #90: Use Git commit head at title screen
Current builds use `VER 1.5` at the lower right of the main menu. Can this be changed to show major and minor versions along with git commit heads for better version tracking?

This appears to be handled in [Descent3/program.h](https://github.com/kevinbentley/Descent3/blob/main/Descent3/program.h#L155).

JeodC:
Attempts to resolve #90, may or may not have been done correctly.

### DIFF
--- a/Descent3/d3serial.cpp
+++ b/Descent3/d3serial.cpp
@@ -177,6 +177,13 @@ int SerialCheck(void) {
 
 #ifdef DEMO
   char ver[10];
+  /*
+  // Not sure this is needed for demo since open source, or might break things if applied
+  char git_head[41];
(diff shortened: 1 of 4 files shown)


Lgt2x:
I don't think we want the source code to be aware of the version control system

JeodC:
> This implementation requires git installed and available in PATH in runtime. Better to use cmake's generate files and include needed data at compile time. Here example how implement this - https://www.marcusfolkesson.se/blog/git-version-in-cmake/

Thanks. Nightly reading to do.

PR #96, opened 2024-04-19
Add D3 utilities source (from kevinbentley/add-missing)

Lgt2x:
Many untested Windows utility sources


Lgt2x:
@kevinbentley I added the license header to every .h/.cpp file, are we good to merge this ?

kevinbentley:
I think so

PR #97, opened 2024-04-19
Quiet Clang Static Analyzer warnings.

MaddTheSane:
Most of the warnings were caused by uninitialized values. 
Some were in plug-ins that didn't have a break in a switch, causing the memory to be deleted twice.


MaddTheSane:
This does not include marking the `Error` function with `__attribute__((__noreturn__))` (or `__dead2` as defined by Darwin's cdefs.h file) (this quiets some other warnings in regards to, I.E., null sent to memory/string functions, because it noted the check for NULL, but also noticed nothing that redirected the logic after the check (although, checking the function, `Error` can return…)).

PR #98, opened 2024-04-20
Dockerfile for hogutils

mstephenson6:
hogUtils-i686 is a dynamically linked 32-bit x86 binary, this is difficult on an aarch64 system, so just wrap it in a debian container. We set QEMU_GUEST_BASE so it's friendly with "qemu-user" mode emulation.

- build in this directory, example:
```
$ podman build . -t localhost/hogutils
```

- then run in one of the scripts/data/linux*hog directories, example:
```
$ podman run -it -v .:/work:Z localhost/hogutils --makehog d3linuxfullhog.txt
```

- in the same directory, verify you see new.hog

### DIFF
--- a/scripts/hogutils/Dockerfile
+++ b/scripts/hogutils/Dockerfile
@@ -1,0 +1,24 @@
+# Dockerfile for hogutils
+#
+# hogUtils-i686 is a dynamically linked 32-bit x86 binary, this is difficult on an aarch64 system,
+# so just wrap it in a debian container. We set QEMU_GUEST_BASE so it's friendly with "qemu-user" mode emulation.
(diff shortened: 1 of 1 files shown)


Lgt2x:
Hi, thanks for the PR. Having to run a 32bit executable during the build step is definitely a pain point. Currently, in CI, we install the required 32bit libraries loaded at runtime to run it. There is an ongoing effort in the community to re-create hog utilities and avoid using this opaque binary, while we don't have access to its source.

Meanwhile, I don't think the added complexity of QEMU and Docker is beneficial to the project, compared to installing the 32bit shared libraries on the system.

PR #101, opened 2024-04-20
Split debug module

winterheart:
Move debug functions to own module.

This PR is part of efforts to isolate subsystems of project into own modules. As discussed earlier, MONO stuff is now mostly useless as requires special hardware for it. Hide this code behind `#ifdef MONO` before removal.


GravisZro:
Umm... so what is this "Monochrome" device? Searching the web hasn't turned up much useful info much less images. Is it somehow related to this? https://archive.org/details/DBMONO_ZIP
I'm also confused why this would need it's own hardware. Please share because the mystery is killing me.

winterheart:
According to @kevinbentley , 

>  That was for writing to a monochrome monitor on our dev boxes 
> i.e. hurcules graphic card... https://en.wikipedia.org/wiki/IBM_Monochrome_Display_Adapter

CMakeLists.txt
```diff
@@ -71,7 +71,12 @@ if(WIN32)
   endif()
 endif()
 
-include_directories("lib" "Descent3" ${PLATFORM_INCLUDES})
+include_directories(
+	${PROJECT_SOURCE_DIR}
```
Lgt2x, line 75: add `ddebug` to the included directories to avoid changing every header (same idea as https://github.com/DescentDevelopers/Descent3/pull/91#discussion_r1573806961)

Descent3/macmain.cpp
```diff
@@ -138,15 +138,15 @@ const char *French_strings[] = {
     "Votre processeur et votre systFme doivent prendre en charge Katmai pour exTcuter ce jeu."};
 
 const char *German_strings[] = {
-    "Descent3 unter Windows NT ben�tigt fnr die Ausfnhrung die NT-Version 4.0 oder h�her",
-    "Descent 3 ben�tigt fnr die Ausfnhrung Windows 9x, NT 4.0 oder h�her.",
+    "Descent3 unter Windows NT ben�tigt fnr die Ausfnhrung die NT-Version 4.0 oder h�her",
+    "Descent 3 ben�tigt fnr die Ausfnhrung Windows 9x, NT 4.0 oder h�her.",
```
Lgt2x, line 142: unintended encoding change

ddebug/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
...
+	>
+	$<$<PLATFORM_ID:Windows>:
+		windebug.cpp
+		winmono.cpp
+	>
+)
+
+add_library(ddebug STATIC ${CPPS})
+
+export(TARGETS ddebug NAMESPACE D3:: FILE ddebugConfig.cmake)
```
Lgt2x, line 14: Where is this export used ?

Lgt2x (changes requested):
`MONO` definition needs to be documented somewhere, so we don't wonder what is it later. You can also open an issue about its removal

InsanityBringer:
something to clarify, but the mono code does not actually use a MDA, it dumps to the debug console on Windows, not sure where on Linux, but KRB seems to think it is worth keeping.

winterheart:
There a lot of changes happens on debug module, need to rework original PR.

winterheart:
Closed in favor #204.

PR #109, opened 2024-04-20
Fix some clang warnings

Lgt2x:
Many more warnings to go through, but this PR should fix the following
-Wdangling-else
-Wparentheses
-Wparentheses-equality
-Wswitch
-Wcomment
-Wlogical-not-parentheses
-Wtautological-pointer-compare

### DIFF
--- a/Descent3/BriefingParse.cpp
+++ b/Descent3/BriefingParse.cpp
@@ -820,13 +820,15 @@ int CBriefParse::ParseBriefing(char *filename) {
 
 done_parsing:;
 
-  if (abort)
+  if (abort) {
     mprintf((0, "Parse aborted\n"));
-  else if (!parse_error)
-    if (reading_text)
+  } else if (!parse_error) {
+    if (reading_text) {
       ParseError("Missing '$endtext'");
-    else
+    } else {
       mprintf((0, "Parse sucessful\n"));
+    }
+  }
 
   // Close the file
   cfclose(ifile);
--- a/Descent3/gamesequence.cpp
+++ b/Descent3/gamesequence.cpp
@@ -1161,6 +1161,14 @@ bool GameSequencer() {
 
   // interpret current function mode.
   switch (GetFunctionMode()) {
+  case INIT_MODE:
+  case GAME_MODE:
+  case EDITOR_MODE:
+  case EDITOR_GAME_MODE:
+  case MENU_MODE:
+  case QUIT_MODE:
+  case CREDITS_MODE:
+    break;
   case RESTORE_GAME_MODE:
   case LOADDEMO_MODE:
   case GAMEGAUGE_MODE:
@@ -1173,6 +1181,8 @@ bool GameSequencer() {
     old_game_state = Game_state;
 
     switch (Game_state) {
+    case GAMESTATE_IDLE:
+      break;
     case GAMESTATE_NEW:
       StartNewGame();
       SetGameState(GAMESTATE_LVLSTART);
--- a/scripts/aigame2.cpp
+++ b/scripts/aigame2.cpp
@@ -946,7 +946,7 @@ void aiBlackStormTrooper::OnInterval(tOSIRISEventInfo *data) {
         int roll = (rand() % 100);
         if (roll <= 50) {
           find_targets(data->me_handle, false);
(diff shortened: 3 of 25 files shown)


Descent3/BriefingParse.cpp
```diff
@@ -805,10 +805,11 @@ done_parsing:;
   if (abort)
     mprintf((0, "Parse aborted\n"));
   else if (!parse_error)
-    if (reading_text)
+    if (reading_text) {
       ParseError("Missing '$endtext'");
-    else
+    } else {
       mprintf((0, "Parse sucessful\n"));
+    }
```
th1000s, line 812: I still get a `-Wdangling-else` warning here.

th1000s, line 812: I still get a `-Wdangling-else` warning here.

Descent3/gamesequence.cpp
```diff
@@ -1161,6 +1161,14 @@ bool GameSequencer() {
...
   // interpret current function mode.
   switch (GetFunctionMode()) {
+  case INIT_MODE:
+  case GAME_MODE:
+  case EDITOR_MODE:
+  case EDITOR_GAME_MODE:
+  case MENU_MODE:
+  case QUIT_MODE:
+  case CREDITS_MODE:
+    break;
```
winterheart, line 1171: For simplicity, it can be replaced with "default: break" on tail of switch.

bitmap/iff.cpp
```diff
@@ -336,7 +336,7 @@ int bm_iff_parse_delta(CFILE *ifile, int len, iff_bitmap_header *bmheader) {
     }
 
     if (cnt == -1) {
-      if (!bmheader->w & 1)
+      if ((!bmheader->w) & 1)
```
winterheart, line 339: This may be dangerous. Need actual testing, both `!(bmheader->w & 1)` and `(!bmheader->w) & 1` may be intended by code logic.

scripts/aigame2.cpp
```diff
@@ -946,7 +946,7 @@ void aiBlackStormTrooper::OnInterval(tOSIRISEventInfo *data) {
         int roll = (rand() % 100);
         if (roll <= 50) {
           find_targets(data->me_handle, false);
-          memory->timer = Game_GetTime() + (memory->target_handle == OBJECT_HANDLE_NONE) ? 5.0f : 2.0f;
+          memory->timer = (Game_GetTime() + (memory->target_handle == OBJECT_HANDLE_NONE)) ? 5.0f : 2.0f;
```
winterheart, line 949: Here unclear what meat to be. I'd think that should be

```c
memory->timer = Game_GetTime() + ((memory->target_handle == OBJECT_HANDLE_NONE) ? 5.0f : 2.0f)
```

winterheart (changes requested):

bitmap/iff.cpp
Lgt2x: `((!bmheader->w) & 1)` is what the compiler does, I just make it explicit

scripts/aigame2.cpp
Lgt2x: `?` has lower priority than +, that's what clang told me when it triggered the warning. My change is equivalent to the previous version

bitmap/iff.cpp
Lgt2x: (again, what clang told me in the warning message)

scripts/aigame2.cpp
winterheart: Sentence `(Game_GetTime() + (memory->target_handle == OBJECT_HANDLE_NONE))` is always true as `Game_GetTime()` always positive. So `(T + X) ? 5.0f : 2.0f` is pointless. There intended to be `T + (X ? 5.0f : 2.0f)`

bitmap/iff.cpp
winterheart: Yes, but we can't tell what intended logic was be. Maybe there already placed bug which clang is pointing to.

scripts/aigame2.cpp
Lgt2x: Right, it has to be an oversight from the original code

bitmap/iff.cpp
Lgt2x: Dropping this modification for now. Looking at the rest of the code it's more likely `!(bmheader->w & 1)` but needs more testing

winterheart (approved):

PR #115, opened 2024-04-20
Add the original text files included with the game

omicronrex:
Some of the files are branded 1.4, but even inside the 1.5 patch files they are still numbered like that.


JeodC:
I am not sure if these files count as game assets and therefore should not be distributed. The wiki is now open and should be modifiable just like pull requests.

omicronrex:
Right! How should we add the contents of the files to the wiki then?

JeodC:
> Right! How should we add the contents of the files to the wiki then?

You can write content in [github markdown format](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax) and attach files to issues. Then we can review them and add to the wiki if approved.

If you need to edit existing wiki pages you can clone the wiki: https://github.com/DescentDevelopers/Descent3.wiki.git

Note that this method will not work with github desktop; you must use git clone. After, you can edit existing wiki pages:

image

JeodC:
I've created an issue template for wiki modifications. Please use that going forward.

PR #121, opened 2024-04-21
New windowed and fullscreen mode

InsanityBringer:
This adds a new fullscreen and windowed mode, with some benefits and caveats, and adds simple logic for mouse capture and release. This also removes the legacy resolution system, and uses a much larger selection of resolutions

Benefits of the new system are that it is possible to leave the window while a UI is up, but it is constrained when in game, and the size of the window is decoupled from the actual rendering resolution (for the 640x480 menus). In windowed mode, the window can be dragged around as usual due to this. This disables the code that makes the window vanish when alt tabbing, though the game will still defer and not execute during this time. 

Caveats right now are that the scaling system uses a virtual screen resolution enforced by setting the viewport and ortho matrices correctly, but this virtual screen resolution isn't used in game so the game's render resolution will always be that of the window. Framebuffer objects could be brought in to fix that if desired.

At the moment, this code only works on Windows. To restore compatibility with other platforms, in the oeApplication implementation for that platform, the virtual functions set_sizepos and set_flags will need to be implemented, and an implementation of opengl_UpdateWindow will need to be implemented in HardwareOpenGL.cpp. For the mouse capture system to work correctly, equivalents of the WM_SETFOCUS and WM_KILLFOCUS messages implemented in winmain.cpp will need to be implemented.

### DIFF
--- a/Descent3/game.cpp
+++ b/Descent3/game.cpp
@@ -958,8 +958,8 @@ void SetScreenMode(int sm, bool force_res_change) {
     int scr_width, scr_height, scr_bitdepth;
 
     if (sm == SM_GAME) {
-      scr_width = Video_res_list[Game_video_resolution].width;
-      scr_height = Video_res_list[Game_video_resolution].height;
+      scr_width = Game_window_res_width;
+      scr_height = Game_window_res_height;
       scr_bitdepth = Render_preferred_bitdepth;
 #ifdef MACINTOSH
       SwitchDSpContex(Game_video_resolution);
@@ -978,6 +978,9 @@ void SetScreenMode(int sm, bool force_res_change) {
       Render_preferred_state.width = scr_width;
       Render_preferred_state.height = scr_height;
       Render_preferred_state.bit_depth = scr_bitdepth;
+      Render_preferred_state.window_width = Game_window_res_width;
+      Render_preferred_state.window_height = Game_window_res_height;
+      Render_preferred_state.fullscreen = Game_fullscreen;
 
       rend_initted = rend_Init(PreferredRenderer, Descent, &Render_preferred_state);
       rend_width = rend_height = 0;
@@ -1008,10 +1011,16 @@ void SetScreenMode(int sm, bool force_res_change) {
         scr_height = 480;
       }
 
-      if (rend_width != scr_width || rend_height != scr_height) {
+	  if (rend_width != scr_width || rend_height != scr_height ||
+          Game_window_res_width != Render_preferred_state.window_width ||
+          Game_window_res_height != Render_preferred_state.window_height ||
+          Game_fullscreen != Render_preferred_state.fullscreen) {
         Render_preferred_state.width = scr_width;
         Render_preferred_state.height = scr_height;
         Render_preferred_state.bit_depth = scr_bitdepth;
+        Render_preferred_state.window_width = Game_window_res_width;
+        Render_preferred_state.window_height = Game_window_res_height;
+        Render_preferred_state.fullscreen = Game_fullscreen;
 
         mprintf((0, "Setting rend_width=%d height=%d\n", scr_width, scr_height));
         int retval = rend_SetPreferredState(&Render_preferred_state);
@@ -1353,4 +1362,12 @@ void DoScreenshot() {
   bm_FreeBitmap(bm_handle);
 
   StartTime();
+}
+
+bool ShouldCaptureMouse() {
+  // If the UI cursor is visible, then the mouse shouldn't be captured.
+  if (ui_IsCursorVisible())
+    return false;
+
+  return true;
 }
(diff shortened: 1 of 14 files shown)


JeodC:
I created the `newwindow` branch for this pull request. It goes a long way for playability on modern machines and we do want people to actually play the game so they find bugs. This will also make it easier for others to contribute to making this work on other platforms.

Descent3/game.cpp
```diff
@@ -1008,10 +1011,16 @@ void SetScreenMode(int sm, bool force_res_change) {
         scr_height = 480;
       }
 
-      if (rend_width != scr_width || rend_height != scr_height) {
+	  if (rend_width != scr_width || rend_height != scr_height ||
+          Game_window_res_width != Render_preferred_state.window_width ||
+          Game_window_res_height != Render_preferred_state.window_height ||
+          Game_fullscreen != Render_preferred_state.fullscreen) {
```
Lgt2x, line 1017: This complex boolean should go into a dedicated function

Descent3/game.cpp
```diff
@@ -1354,3 +1363,11 @@ void DoScreenshot() {
 
   StartTime();
 }
+
+bool ShouldCaptureMouse() {
+  // If the UI cursor is visible, then the mouse shouldn't be captured.
+  if (ui_IsCursorVisible())
```
Lgt2x, line 1369: Side note, what controls ui_IsCursorVisible? the logic seems scatted around the codebase

Descent3/init.cpp
```diff
@@ -1457,6 +1460,8 @@ void LoadGameSettings() {
   Database->read("RS_bitdepth", &Render_preferred_bitdepth, sizeof(Render_preferred_bitdepth));
   Database->read_int("RS_resolution", &Game_video_resolution);
   Database->read_int("RS_bilear", &Render_preferred_state.filtering);
+  Database->read_int("RS_windowwidth", &Game_window_res_width);
+  Database->read_int("RS_windowheight", &Game_window_res_height);
```
Lgt2x, line 1464: No need to read "RS_fullscreen" now ?

Descent3/winmain.cpp
```diff
@@ -178,7 +191,7 @@ class oeD3Win32App : public oeWin32Application {
           shutdown = false;
         }
       }
-    } break;
+    } break;*/
```
Lgt2x, line 194: ?

Descent3/winmain.cpp
```diff
@@ -624,8 +637,16 @@ int PASCAL HandledWinMain(HINSTANCE hInst, HINSTANCE hPrevInst, LPSTR szCmdLine,
   if (Dedicated_server) {
     d3 = new oeD3Win32App(OEAPP_CONSOLE, (HInstance)hInst);
   } else {
+    // [ISB] Reading this setting needs to be done before creating any windows. 
+    int temp;
+    Database->read_int("RS_fullscreen", &temp);
+    Game_fullscreen = !!temp;
```
Lgt2x, line 643: (temp == 0)

ddio_win/winmouse.cpp
```diff
@@ -356,9 +379,45 @@ int RawInputHandler(HWND hWnd, unsigned int msg, unsigned int wParam, long lPara
...
+        if (!GetCursorPos(&mousept))
+          Int3();
+        if (!ScreenToClient(hWnd, &mousept))
+          Int3();
+        // Get the client rectangle of the window and map brect to it.
+        RECT clientrect;
+        if (!GetClientRect(hWnd, &clientrect))
+          Int3();
+
+        int brectwidth = DDIO_mouse_state.brect.right - DDIO_mouse_state.brect.left;
```
Lgt2x, line 394: camelCase ?

renderer/HardwareOpenGL.cpp
```diff
@@ -731,6 +735,83 @@ int opengl_Setup(oeApplication *app, int *width, int *height) {
 }
 #endif
 
+#if defined(WIN32)
```
Lgt2x, line 738: What's specific to Windows here ? Is it just that `set_flags` needs to be implemented for other platforms ?

Lgt2x (commented):
good job. For the record, what needs to be ported for cross-platform support ?

Descent3/game.cpp
DanielGibson: why, is that condition used multiple times?

renderer/HardwareOpenGL.cpp
InsanityBringer: That code has a couple of Win32 specific calls, specifically GetSystemMetrics and AdjustWindowRect. It may be possible to generalize this with some more API in the platform specific stuff, but it may not be worth it if the jump to SDL is made

Descent3/init.cpp
InsanityBringer: RS_fullscreen is read in winmain.cpp, it needs to be read before the initial window is created

Descent3/game.cpp
Lgt2x: it doesn't matter. It would make more explicit in which case this condition is triggered

PR #122, opened 2024-04-21
Own implementation of HogUtils - HogMaker

### LINKED ISSUE #61: Source for hogUtils-i686
`hogUtils-i686 (scripts/hogutils/hogUtils-i686)` is executed as part of the build system to build hog files.
However, it is a linux 32-bit executable, which means that extra steps are required for building on anything other  than 32-bit linux.
It would be nice to have the source for this, so it can be built as part of the build system, and thus be native to whatever system it is used on.

winterheart:
HogMaker can create HOG files as does HogUtils.

Rearranged custom target creation of HOGs and placing them in right place.

Closes #61.


scripts/CMakeLists.txt
```diff
@@ -65,47 +64,48 @@ set(SCRIPTS
 	TrainingMission
 	Y2K)
 
+add_custom_target(HogLinuxFull-copy
```
Lgt2x, line 67: if (UNIX)

tools/HogMaker/HogFormat.h
```diff
@@ -0,0 +1,70 @@
...
+FILE 1			[filelen(FILE 1)]
+...
+FILE NFILES-1		[filelen(NFILES -1)]
+*/
+
+struct HogHeader {
+  std::vector<char> tag = {'H', 'O', 'G', '2'};           // "HOG2"
+  uint32_t nfiles = 0;                                    // number of files in header
+  uint32_t file_data_offset = 68;                         // offset in file to filedata.
+  std::vector<char> reserved = std::vector<char>(56, -1); // filled with 0xff
```
Lgt2x, line 45: use `0xFF` in init

tools/CMakeLists.txt
```diff
@@ -0,0 +1,5 @@
+add_executable(
```
Lgt2x, line 1: tools/CMakeLists.txt should be `add_subdirectory(HogMaker)` and this code would go in tools/HogMaker/CMakeLists.txt

scripts/CMakeLists.txt
winterheart: HogLinuxFull-copy is dependency of SCRIPTS targets so it can't be hided behind UNIX guard. Copying of directory is not a big deal for cross-platform support. I think, next step would be recompile hog file for every platform including Windows.

tools/CMakeLists.txt
winterheart: That would be excessive intermediate CMakeLists.txt with only `add_subdirectory(HogMaker)`, there no point to create CMakeLists.txt on every directory level.

Lgt2x: I was thinking ahead of time to prepare for when we'll have more executables in `tools`, but we can keep it this way for now

PR #123, opened 2024-04-21
[CI] Use same job for both Debug and Release builds

danra:
Making use of CMake's multi-config generators which we're using. 
Simpler and more efficient, halves number of jobs.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -42,10 +42,10 @@ jobs:
         run: cmake --preset ${{ matrix.os.preset }}
 
       - name: Debug Build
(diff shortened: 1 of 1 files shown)


Lgt2x:
Good idea, but artifacts don't seem correct

JeodC:
Right, artifacts only need the Debug and Release folders.

image

JeodC:
No artifacts were found: https://github.com/DescentDevelopers/Descent3/actions/runs/8773591912

danra:
Yes I see :)
Prodding it until it works

danra:
Looks good now

JeodC:
Perfect, thank you.

Issue #110, opened 2024-04-20
Windowed mode doesn't respect resolution


JeodC:
On windows 11 at least, using windowed mode puts a 640x480 window in the top left corner of the display. This window does not respect video resolution changes, and cannot be dragged.

JeodC:
In `Descent3/win32/winapp.cpp:L231` we have the following code. However the build is Release mode, not Debug.
```cpp
#ifdef RELEASE
    SetRect(&rect, 0, 0, GetSystemMetrics(SM_CXSCREEN), GetSystemMetrics(SM_CYSCREEN));
#else
    SetRect(&rect, 0, 0, 640, 480);
#endif
```
The full clip of relevant code:
```cpp
if (flags & OEAPP_CONSOLE) {
    m_X = CW_USEDEFAULT;
    m_Y = CW_USEDEFAULT;
    m_W = 640;
    m_H = 480;
} else {
    // Retrieve selected resolution from game options
    int selectedWidth = GetSelectedWidthFromOptions(); // Example function to get selected width
    int selectedHeight = GetSelectedHeightFromOptions(); // Example function to get selected height

    // Set the initial window dimensions based on selected resolution
    m_X = 0;
    m_Y = 0;
    m_W = selectedWidth;
    m_H = selectedHeight;
}
```
Could it be a flag must be changed somewhere?

JeodC:
This is a debugging tool, not a full-fledged feature.

PR #131, opened 2024-04-21
Fixed Release build

pzychotic:
This adds the missing preprocessor define RELEASE in Release builds.

This is related to #110, but still has problems in windowed mode (start menu screen and videos not in correct resolution).


JeodC:
Screenshot002
Screenshot003

I don't know if it's my previous settings causing this, but it doesn't seem quite right.

pzychotic:
That's why I wrote "but still has problems in windowed mode (start menu screen and videos not in correct resolution)."
Ingame resolutions looks ok.

JeodC:
> That's why I wrote "but still has problems in windowed mode (start menu screen and videos not in correct resolution)." Ingame resolutions looks ok.
> 
> And there seems to be more inconsisties between #ifndef RELEASE and #ifdef _DEBUG that only work on Windows.

Ingame runs fine, yes, but the ingame ui does not work well, putting the game in a less playable state. I think this is better accomplished by #121.

pzychotic:
> > That's why I wrote "but still has problems in windowed mode (start menu screen and videos not in correct resolution)." Ingame resolutions looks ok.
> > And there seems to be more inconsisties between #ifndef RELEASE and #ifdef _DEBUG that only work on Windows.
> 
> Ingame runs fine, yes, but the ingame ui does not work well, putting the game in a less playable state in window mode. I think this is better accomplished by #121. The caveat here is that #121 only works on Windows, and I think this is supposed to function on all platforms?

This started out to just add the missing preprocessor define for RELEASE, which just happens to be related to the window mode.
But with all the additional problems and breaking Linux/Max builds, that I can't test locally right now, this might better be closed for now.
I haven't looked at #121 or why that only works for Windows.

JeodC:
> > > That's why I wrote "but still has problems in windowed mode (start menu screen and videos not in correct resolution)." Ingame resolutions looks ok.
> > > And there seems to be more inconsisties between #ifndef RELEASE and #ifdef _DEBUG that only work on Windows.
> > 
> > 
> > Ingame runs fine, yes, but the ingame ui does not work well, putting the game in a less playable state in window mode. I think this is better accomplished by #121. The caveat here is that #121 only works on Windows, and I think this is supposed to function on all platforms?
> 
> This started out to just add the missing preprocessor define for RELEASE, which just happens to be related to the window mode. But with all the additional problems and breaking Linux/Max builds, that I can't test locally right now, this might better be closed for now. I haven't looked at #121 or why that only works for Windows.

I appreciate the effort, it will no doubt help others know where to look. #121 only works on Windows because the submitter hasn't been able to add the needed code for other platforms.

pzychotic:
Can we reopen this with my latest changes to the branch?

I found the problem with the inconsistent Linux/Mac builds. _DEBUG was always active for them, even in Release build.
I testet this on my fork with a modified GitHub workflow file to allow executing the build actions for all branches. Which now all pass: https://github.com/pzychotic/Descent3/actions/runs/8773558703

I also made a workaround to keep the old behavior regarding #110.

JeodC:
> Can we reopen this with my latest changes to the branch?
> 
> I found the problem with the inconsistent Linux/Mac builds. _DEBUG was always active for them, even in Release build. I testet this on my fork with a modified GitHub workflow file to allow executing the build actions for all branches. Which now all pass: https://github.com/pzychotic/Descent3/actions/runs/8773558703
> 
> I also made a workaround to keep the old behavior regarding #110.

There's not really a need for that workaround at the moment. #110 was closed due to the proposed merge branch being removed. We are contemplating moving over to a gitflow strategy, which we think will help with multiple forks trying to achieve a similar goal.

pzychotic:
I'm not quite sure that I correctly understand your comment. Do you mean my workaround for window mode changes are not required if we take the rest of this PR?

Just in case my description of why we want this PR weren't quite clear:
Right now we create Release builds with a lot of debug code, because RELEASE was never set on any platform. And additionally on Linux/Mac _DEBUG was set even for Release build.
Correcting these build problems leads to the problem you posted above in the screenshots. So my workaround just keeps the current state with the fixed resolution of 640x480 in window mode.
I haven't had the chance to look deeper into it to correctly fix window mode.

JeodC:
> I'm not quite sure that I correctly understand your comment. Do you mean my workaround for window mode changes are not required if we take the rest of this PR?
> 
> 
> 
> Just in case my description of why we want this PR weren't quite clear:
> 
> Right now we create Release builds with a lot of debug code, because RELEASE was never set on any platform. And additionally on Linux/Mac _DEBUG was set even for Release build.
> 
> Correcting these build problems leads to the problem you posted above in the screenshots. So my workaround just keeps the current state with the fixed resolution of 640x480 in window mode.
> 
> I haven't had the chance to look deeper into it to correctly fix window mode.

Your workaround isn't required, right. It's my opinion that windowed mode was going to be a 1.5 feature that didn't come out of debugging. I see two options: disable it in release or complete the intention (perhaps still restrict its resolution but allow it to be dragged and function as a window instead of the viewpoint currently is).

pzychotic:
> > I'm not quite sure that I correctly understand your comment. Do you mean my workaround for window mode changes are not required if we take the rest of this PR?
> > Just in case my description of why we want this PR weren't quite clear:
> > Right now we create Release builds with a lot of debug code, because RELEASE was never set on any platform. And additionally on Linux/Mac _DEBUG was set even for Release build.
> > Correcting these build problems leads to the problem you posted above in the screenshots. So my workaround just keeps the current state with the fixed resolution of 640x480 in window mode.
> > I haven't had the chance to look deeper into it to correctly fix window mode.
> 
> Your workaround isn't required, right. It's my opinion that windowed mode was going to be a 1.5 feature that didn't come out of debugging. I see two options: disable it in release or complete the intention (perhaps still restrict its resolution but allow it to be dragged and function as a window instead of the viewpoint currently is). Further optimizations like resizing windows is a modernization that should come after 1.5 release.

Ok, I reverted the workaround and disabled window mode in Release builds for now.
If I find time on the weekend I might look deeper into the window mode stuff, outside of this PR.

JeodC:
Thanks, sorry for the wait.

PR #132, opened 2024-04-21
Add -fov command-line option

JeodC:
This adjusts the default FOV to 90 and adjusts the default zoom proportionally. A FOV of 90 is the common default on modern systems. Ideally a FOV slider would be built in the future.

EDIT: See comments further down, this has evolved into a CLI.


JeodC:
This appeared to look fine for me, but of course I would appreciate other tester input.

JeodC:
fov

Here is a side by side comparison. Changed defaults on the left. This feels more natural to me and is less straining on the eyes.

JeodC:
I changed this to add `-fov <k>` as a command-line option, but it doesn't work. I'm not sure why.

PR #133, opened 2024-04-21
Remove unnecessary build dependency

Jayman2000:
2d4d82d (Accurately find the linux artifacts (#80), 2024-04-19) made it so that make no longer gets used as a part of the build process, but it didn’t remove make from the list of build dependencies.


JeodC:
Thanks.

Issue #83, opened 2024-04-19
libacm crash in UnpackLinear


MaddTheSane:
When running on an Apple Silicon Mac, I've run into an issue with libacm, specifically the `InternalAudioDecoder::UnpackLinear` function.

https://github.com/kevinbentley/Descent3/blob/1edaa3b628b6945f93f7fde453967e244fabb45b/libacm/libacm.cpp#L574-L589

It seems like when `id` is 9 (and thus `middleIndex` gets set to *256*) and `offset` gets set to 255, it results in trying to get a negative address at `decoder.m_pMidBuffer`, which causes a crash.

JeodC:
Maybe do something like this?
```cpp
     if (offset < middleIndex) {
       //Set pBlock[pos] to a default value, return false, or log an error
     }
 ```
 
 This would go above
 ```cpp
      const uint32 pos = (i << level) + col; 
     pBlock[pos] = decoder.m_pMidBuffer[offset - middleIndex]; 
     ```

**Comment by kevinbentley** [comment]
I've been debugging this same issue under Linux. I've been playing whack a mole trying to catch all the negative values, but we need to get to the root of the problem. It might be worth looking at switching to this library, which must be pretty good because it's part of ffmpeg now. [https://github.com/markokr/libacm](https://github.com/markokr/libacm)

**Comment by midnite8177** [comment]
I bet GetBitsEOF has the bug
Probably worth trying out that library

**Comment by midnite8177** [comment]
Marko Kreen wrote the original libacm, so this must be related

**Comment by bryanperris** [comment]
What about creating unit tests?

**Comment by devint1** [comment]
The issue seems to be that the C++ compiler treats `offset - middleIndex` as `uint32` instead of `sint32`.

This seems to fix the issue for now: 

```cpp
// pBlock[pos] = decoder.m_pMidBuffer[offset - middleIndex]; 
pBlock[pos] = decoder.m_pMidBuffer[((sint32) offset) - middleIndex];
```

DanielGibson:
The libacm code in this repo is very different than the upstream code (which is plain C).
Might make sense to import and use the current libacm code https://github.com/markokr/libacm/? Probably has this and other bugs fixed

paines:
I just checked https://github.com/markokr/libacm . This one compiles and runs fine on M1. I used [HUMAN1.ACM](https://sembiance.com/fileFormatSamples/audio/interplayACM/HUMAN1.ACM) with acmtool to test it.

[...]

PR #134, opened 2024-04-21
Acm library refactor

kevinbentley:
Replace libacm with one that has an open source library.


JeodC:
This meant to resolve #83 right? Tested?

JeodC:
Windows release tested. Background music (such as title screen) no longer plays.

kevinbentley:
> Windows release tested. Background music (such as title screen) no longer plays.

Thanks, I'm looking into it.

lib/streamaudio.h
```diff
@@ -143,9 +143,15 @@
 #include "ssl_lib.h"
 #include "TaskSystem.h"
 
+extern "C" {
+#include "libacm.h"
+}
```
GravisZro, line 148: Headers are supposed to take care of themselves. Only write `#include "libacm.h"` here.
In libacm.h add this before declarations
```
#ifdef __cplusplus
extern "C" {
#endif

```
and then this after the declarations
```
#ifdef __cplusplus
}
#endif

```

DanielGibson:
Oops, as I announced in #83 yesterday, I did basically the same work, see #135

I also fixed that extern "C" issue and sent a pull request to libacm: https://github.com/markokr/libacm/pull/1

stream_audio/streamaudio.cpp
```diff
@@ -837,12 +894,25 @@ void *AudioStream::StreamCallback(int *size) {
...
-    // We have a compressed stream
-    return m_decoder->Read(m_buffer[buf].data, len);
+  
+  int big_endian = 0;
+  #if OUTRAGE_BIG_ENDIAN
+  big_endian = 1;
+  #endif
+  int len_requested = len;
+  //mprintf((0,"Reading stream: len=%d bytes_read=0,len_requested=%d\n",len,len_requested));
+  int bytes_read = acm_read(m_acm,m_buffer[buf].data,len_requested,big_endian,2,1);
```
DanielGibson, line 904: this might return a value < 0 on error, and 0 on EOF, I think those cases should be handled?

kevinbentley: Good call! Thanks!

lib/streamaudio.h
kevinbentley: Fair enough. I didn't want to modify the code from the other library so we could drop in a new version.

kevinbentley:
> I also fixed that extern "C" issue and sent a pull request to libacm: [markokr/libacm#1](https://github.com/markokr/libacm/pull/1)

If they take the PR, that's great. If they don't, I'm conflicted about what to do as far as the libacm.h include comment above. On one hand, we only included that header in one place, so I don't think it's too big of a deal. But I agree that it's not the best approach to wrap the include with the extern c stuff.

DanielGibson:
My PR for libacm has just been merged :)

PR #135, opened 2024-04-21
Use upstream libacm 1.3 for ACM audio decoding

DanielGibson:
Using the latest code from https://github.com/markokr/libacm

This should make future upgrade to newer libacm versions much easier.

fixes #83


Arcnor:
This is the same as #134, one of them need to go :)

DanielGibson:
well, review both and decide which one you wanna keep :shrug:

DanielGibson:
rebased to resolve merge conflicts

JeodC:
For this, in windows release, background music plays, but at an extremely slow pace. I can record it if needed.

JeodC:
Background music becomes corrupted and unstable on entering a mission or starting a movie.

DanielGibson:
did it work better before or with #134 ?

on linux, sound generally seems corrupted, also in the movies

kevinbentley:
> did it work better before or with #134 ?
> 
> on linux, sound generally seems corrupted, also in the movies

Movies use a different sound system. This will only effect things like the in game music, the background music at the main menu, etc. Effects like weapon firing use a different system.

So that being said, is the music you hear in the main menu corrupted? If so, could you grab a clip somehow (upload something to youtube from your phone if needed.  :)

kevinbentley:
> For this, in windows release, background music plays, but at an extremely slow pace. I can record it if needed.

I'm looking into that now.

JeodC:
> > did it work better before or with #134 ?
> > on linux, sound generally seems corrupted, also in the movies
> 
> Movies use a different sound system. This will only effect things like the in game music, the background music at the main menu, etc. Effects like weapon firing use a different system.
> 
> So that being said, is the music you hear in the main menu corrupted? If so, could you grab a clip somehow (upload something to youtube from your phone if needed. :)

Give me a few, gotta wait for the zoo to go outside. :)

JeodC:
https://github.com/DescentDevelopers/Descent3/assets/47716344/9a49e2d4-8874-4c3a-aa86-7c814aea58a0

This is with artifact windows-release from this pull.

libacm/adecode.cpp
```diff
@@ -0,0 +1,161 @@
...
+//   pBuffer: The buffer to receive the data into
+//    amount: How much data to read
+// Returns the number of bytes read - zero when we're at the end of the file
+uint32 InternalAudioDecoder::Read(void *pBuffer, uint32 amount) {
+  const int bigendianp = 0; // we want little endian samples
+  const int wordlen = 2;    // the only supported value
+  const int sgned = 1;      // we want signed samples
+  uint32 totalBytesRead = 0;
+
+  while (totalBytesRead < amount) {
```
kevinbentley, line 111: This code appears to read over and over into the same place in the buffer, rather than appending to the buffer with each read.

kevinbentley:
Sorry about adding to the confusion, I didn't realize this was a different implementation of what I did (move to libacm). I do like the approach @DanielGibson took with this (he created a new decoder class, I just bypassed it and merged the decoder capabilities into streamaudio.cpp). Both of ours appear to have bugs. I'm not sure which is better to use, but I'm going to keep debugging my code in Windows, and I'll let someone else decide which one to merge.  :)

libacm/adecode.cpp
DanielGibson: oh right - damn =)

(this is what I meant with sound on linux being generally broken, I couldn't tell if I did this right or not, just that it didn't crash anymore)

JeodC:
This is now working on windows.

kevinbentley:
This works under Linux too, so I closed mine and we'll go with this. Nice job!

PR #137, opened 2024-04-21
Use system zlib instead of bundled.

Nakhr11n:
It's unnecessary to keep zlib in the project files.
Better to link against system zlib, which is probably more up to date.

If we really want to link statically, I would suggest using [CMake ExternalProject](https://cmake.org/cmake/help/latest/module/ExternalProject.html) to do so.


Lgt2x:
CI scripts and build instructions will need updated to install zlib

winterheart:
Better to use vcpkg to install required dependencies. It has native cmake integration via toolchain files and can build zlib statically on request.

vcpkg-configuration.json
```diff
@@ -0,0 +1,14 @@
+{
```
Arcnor, line 1: Rather than having this file, you can simplify and use the built-in baseline on vcpkg.json, check what I had at https://github.com/Arcnor/Descent3/commit/b3b31ccd36ae0760b54f8574775236c57379cf51

CMakePresets.json
```diff
@@ -15,7 +15,10 @@
 				"lhs": "${hostSystemName}",
 				"rhs": "Windows"
 			},
-			"architecture": "Win32"
+			"architecture": "Win32",
+			"cacheVariables": {
+				"CMAKE_TOOLCHAIN_FILE": "C:/vcpkg/scripts/buildsystems/vcpkg.cmake"
+			}
```
winterheart, line 21: If vcpkg is installed somewhere else, will that preset work with manually defined CMAKE_TOOLCHAIN_FILE?

Lgt2x (approved):

Lgt2x:
I'm being nit-picky here but could you please update README build instructions for Windows users to include VCPKG please ? @Arcnor can confirm but after that I think we're ready to merge :)

CMakePresets.json
```diff
@@ -15,7 +15,10 @@
 				"lhs": "${hostSystemName}",
 				"rhs": "Windows"
 			},
-			"architecture": "Win32"
+			"architecture": "Win32",
+			"cacheVariables": {
+				"CMAKE_TOOLCHAIN_FILE": "$env{VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake"
```
Arcnor, line 20: Instead of this, `toolchainFile` should be used

Arcnor (changes requested):

Arcnor:
Yeah, it will be great, because this PR introduced extra steps to build the project. 

You can just mention the need to define `VCPKG_ROOT` before calling the CMake commands.

Arcnor (approved):

Arcnor:
Thanks @Nakhr11n and sorry for all the requests, if you fix the conflict we can just merge this :)

PR #138, opened 2024-04-21
Allow bypassing the launcher, but warn the user.

JeodC:
Editing this file transcoded it to UTF-8. The actual change is:
```cpp
if (!FindArg("-launched") && !FindArg("-dedicated") && !FindArg("-timetest")) {
    int result = MessageBox(NULL, "Bypassing the launcher might break some settings. Launch anyway?", PRODUCT_NAME " Error", MB_YESNO);
    if (result == IDYES) {
    } else {
        return 0;
    }
}
```
This allows the user to bypass the launcher. Useful for testing purposes and eliminates the need to name the binaries `main.exe` in Windows.

### DIFF
--- a/Descent3/winmain.cpp
+++ b/Descent3/winmain.cpp
@@ -31,71 +31,72 @@
 #include "dedicated_server.h"
 #include "resource.h"
 
-const char *English_strings[] = {"Descent 3 under Windows NT requires version 4.0 or greater of NT to run.",
-                                 "Descent 3 requires  Windows 9x, NT 4.0 or greater to run.",
-                                 "",
-                                 "You must install DirectX through the Descent 3 Launcher before continuing.",
-                                 "You must install at least Service Pack 3 to run Descent 3 under Windows NT 4.0.",
-                                 "Failed to retrieve DirectX version.",
-                                 "Descent 3 requires DirectX 3 or greater on this machine.",
-                                 "This version of Windows NT doesn't have DirectX 3 or greater installed.",
-                                 "Your processor and system must support Katmai to run this game."};
+const char *English_strings[] = {
+     "Descent 3 under Windows NT requires version 4.0 or greater of NT to run.",
+     "Descent 3 requires  Windows 9x, NT 4.0 or greater to run.",
+     "",
+     "You must install DirectX through the Descent 3 Launcher before continuing.",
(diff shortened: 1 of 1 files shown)


JeodC:
image

JeodC:
Verified functioning on Windows.

Jayman2000:
> Editing this file transcoded it to UTF-8.

I don’t think that the transcoding was successful. You’re change introduced stuff like this:

```c++
    "Descent3 unter Windows NT ben÷tigt fnr die Ausfnhrung die NT-Version 4.0 oder h÷her",
```

I don’t speak German, but I don’t think that there should be divided-by signs in the middle of words.

JeodC:
> > Editing this file transcoded it to UTF-8.
> 
> I don’t think that the transcoding was successful. You’re change introduced stuff like this:
> 
> ```c++
> const char *German_strings[] = {
>     "Descent3 unter Windows NT ben÷tigt fnr die Ausfnhrung die NT-Version 4.0 oder h÷her",
> ```
> 
> I don’t speak German, but I don’t think that there should be divided-by signs in the middle of words.

I used github web browser. I'll have to find a diff and see if I can revert.

JeodC:
Hm, looks like those are present on main as well.

https://github.com/DescentDevelopers/Descent3/blob/bypass-launcher/Descent3/winmain.cpp

Lgt2x:
> > > Editing this file transcoded it to UTF-8.
> > 
> > 
> > I don’t think that the transcoding was successful. You’re change introduced stuff like this:
> > ```c++
> > const char *German_strings[] = {
> >     "Descent3 unter Windows NT ben÷tigt fnr die Ausfnhrung die NT-Version 4.0 oder h÷her",
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I don’t speak German, but I don’t think that there should be divided-by signs in the middle of words.
> 
> I used github web browser. I'll have to find a diff and see if I can revert.

German characters should be `ö`

winterheart:
Localized messages written in encoding that accepts locales of Windows. Recent changes converted them to UTF8, and there may additional efforts may required to get working on de/fr/it locales. MessageBox() may or not may work with UTF8 strings.

Issue #8, opened 2024-04-16
Loading any demo play file for native linux results in crash


bryanperris:
How to reproduce:
1. Install the game from Steam on a linux based system (Steam Deck for me)
2. From the main menu, select demo and pick any demo file
3. Process will crash

Jayman2000:
I don’t think that this is the right place to report that bug. The Linux/macOS version of Descent 3 that’s on Steam was created by icculus. [Unfortunately, he hasn’t gotten permission in order to release his version of the code yet.](https://github.com/kevinbentley/Descent3/issues/9) The code in this repo doesn’t contain any of his changes. Issues with icculus’s version of the game should be reported [here](https://github.com/icculus/descent3-bugs).

Are you able to reproduce this bug if you build the source code that’s in this repo?

bryanperris:
If someone uploads a compiled D3 build for linux, I can test it out on my machine.  I guess anyone that manages to get the source built on linux and just quickly check if any demo file is able to play without crashing.  I filed issues as soon this code was realized based on my SteamDeck D3 experience, not knowing much of much of which source repo was used for the Steam release. This weekend I plan to just clone the repo when I get time, and attempt to compile it myself.

bryanperris:
It seems the crash issue does exist for demo play, I have caught something in GDB

`Thread 1 "Descent3" received signal SIGSEGV, Segmentation fault.
0x0000000000677157 in hlsSystem::ComputePlayInfo (this=0x1339040 <Sound_system>, sound_obj_index=0, virtual_pos=0x7fffffffbb90, virtual_vel=0x7fffffffbb84, 
    adjusted_volume=0x7fffffffbb80) at /ho`

bryanperris:
I attempted to play a demo file, and then bad demo file is displayed, here is a GDB log

Int3 in /home/bperris/base/dev/Descent3/Descent3/loadstate.cpp at line 769.(Descent 3 Debug Break)

System Error

It's probably safe to continue.  Continue?

Thread 1 "Descent3" received signal SIGTRAP, Trace/breakpoint trap.
trap_instruction () at /home/bperris/base/dev/Descent3/lib/debugbreak.h:48
48	}
Missing separate debuginfos, use: dnf debuginfo-install xorg-x11-drv-nvidia-libs-550.67-1.fc39.x86_64
(gdb) c
Continuing.

[...]

Issue #7, opened 2024-04-16
Save/Load corruption bug for native Linux


bryanperris:
How to reproduce:
1. Install the game from Steam on a Linux based system (Steam Deck for me)
2. Start a new game and make a save
3. Attempt to load the save, sometimes it will be fine.
4. If the game saves a bad save file, it crashes the game upon load.

kevinbentley:
Can you verify that this bug still exists once the source has been compiled? We fixed a bunch of bugs that never made it into release. I'll try to find the list we had.

bryanperris:
Do you have the official build instructions I should follow for Linux? Once its built, I can make a duplicate of the D3 Steam directory, and replace the binary executable with the GitHub version to test on the Steam Deck.

bryanperris:
I think for my reported tickets, there should be a Steam label just to track, I found this issue from Steam version and not from this repo.

bryanperris:
Agreed, I meant more that the label to mark 'bug' issues not being related to this repo, I have no doubt others will also post here bugs they found on Steam D3 version.

JeodC:
> Author

I can see how that could cause confusion. Label created and applied.

bryanperris:
This issue does exist

So in the terminal a lot of these messages
RECOMPILE YOUR SCRIPTS!!!
Game-Checksum FAIL!!! (2273889867l!=2273873307l)
RECOMPILE YOUR SCRIPTS!!!
Game-Checksum FAIL!!! (2273889867l!=2273873307l)
RECOMPILE YOUR SCRIPTS!!!
Game-Checksum FAIL!!! (2273889867l!=2273873307l)
RECOMPILE YOUR SCRIPTS!!!
Game-Checksum FAIL!!! (2273889867l!=2273873307l)
RECOMPILE YOUR SCRIPTS!!!

Then what I caught in GDB:
`Thread 1 "Descent3" received signal SIGSEGV, Segmentation fault.
0x000000000053ae1a in ObjUnlink (objnum=236) at /home/bperris/base/dev/Descent3/Descent3/object.cpp:1478
1478	      seg->objects = obj->next;
`

bryanperris:
@kevinbentley 
I have applied this change to object.cpp line 1475:
terrain_segment *seg = &Terrain_seg[cellnum & 0xFFFF];

[...]

PR #140, opened 2024-04-21
Object unlink sanity check

bryanperris:
This is a simply sanity check: if an object has already been unlinked from a room, do nothing else and just return.

This fixes the crash issue for issues:
https://github.com/DescentDevelopers/Descent3/issues/7
https://github.com/DescentDevelopers/Descent3/issues/8


JeodC:
Will this also fix #141?

PR #143, opened 2024-04-21
Fix possible heap buffer overflow

winterheart:
Strict size of name of entry with 36 bytes.
Thanks to MaddTheSane for pointing out.

### DIFF
--- a/tools/HogMaker/HogFormat.h
+++ b/tools/HogMaker/HogFormat.h
@@ -46,10 +46,10 @@ struct HogHeader {
 };
 
 struct HogFileEntry {
-  std::filesystem::path name; // file name (36 char max)
-  uint32_t flags;             // extra info
-  uint32_t len;               // length of file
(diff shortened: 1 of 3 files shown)


tools/HogMaker/HogFormat.h
```diff
@@ -46,10 +46,10 @@ struct HogHeader {
 };
 
 struct HogFileEntry {
-  std::filesystem::path name; // file name (36 char max)
-  uint32_t flags;             // extra info
-  uint32_t len;               // length of file
-  uint32_t timestamp;         // time of file.
+  std::vector<char> name = std::vector<char>(36, 0); // file name (36 char max)
```
th1000s, line 49: The vector can still be `push`'ed onto. If the size is always 36 (padded with `'\0'`), what about using `std::array<char, 36>`?

MaddTheSane:
The static analyzer didn't complain, so this will also work.

PR #144, opened 2024-04-22
Build the plug-ins as CMake modules.

MaddTheSane:
This causes the level scripts/plugins to be built, on macOS, as a mach bundle (doesn't have an install name) instead of a library.
This also includes a hack to make the created modules for macOS match the old (pre-patch) file extension, otherwise they would have the extension of ".so" by CMake.

### DIFF
--- a/netcon/lanclient/CMakeLists.txt
+++ b/netcon/lanclient/CMakeLists.txt
@@ -3,7 +3,7 @@ set(CPPS lanclient.cpp)
 
 set(NETGAME_MODULE "TCP_IP")
 
-add_library(${NETGAME_MODULE} SHARED ${CPPS} ${HEADERS})
+add_library(${NETGAME_MODULE} MODULE ${CPPS} ${HEADERS})
 set_target_properties(${NETGAME_MODULE} PROPERTIES PREFIX "")
 set_target_properties(${NETGAME_MODULE} PROPERTIES SUFFIX ".d3c")
 target_link_libraries(${NETGAME_MODULE} inetfile)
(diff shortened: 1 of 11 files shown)


Lgt2x:
I don't really get it, what's the actual benefit of the mach bundle instead of a library? What advantage to we get to use `MODULE` instead of `SHARED` ?

winterheart:
I think these changes are intended to allow regenerate d3-linux.hog for macOS.

MaddTheSane:
Mach bundles are designed to be loaded at runtime, as opposed to dynamic libraries, which are designed to be loaded at link time.

This isn't that important.

JeodC:
This is going to make components load at runtime instead of link time right? Will this break any compatibility with the older systems the game was originally built for? If so, this might be a modernization to push back for later.

Lgt2x:
@Arcnor @jcoby  has this PR been tested? Can we merge?

Arcnor:
I tested it and it fails on Mac Intel at least. I cannot say if it read because of it or something merged just before it, though

Lgt2x:
> I tested it and it fails on Mac Intel at least. I cannot say if it fails because of it or because of something merged just before it, though

please investigate and report back =)

JeodC:
This PR apparently changes dependencies to be required in the installdir. If this is merged, `zlib1.dll` will not be static and must be copied to the installdir.

JeodC:
From discord user thunderbird: `The current main branch requires zlib1.dll in the Descent3 install directory, this PR removes the need`.

Jayman2000 (approved):
I think that this change makes sense, and I was able to verify that it works on NixOS 23.11. The commit that this PR is based on is broken on my system, but when I rebased this PR on main, it worked fine.

MaddTheSane:
Just rebased.

JeodC:
> Just rebased.

Once the conflict is resolved I'll merge. Sorry it's taken so long.

MaddTheSane:
Rebased again.

Issue #129, opened 2024-04-21
Inconsistent line endings


Jayman2000:
Some files in this repo use LF line endings, and some files in this repo use CRLF line endings. [I created a script that will calculate how many files use LF line endings and how many files use CRLF line endings.](https://gist.github.com/Jayman2000/f8e1d010401c0f19125729546fac199b) Here’s what the script says when you run it on the most recent commit (07d988e194cfdaf87a86d798bb423af0ff25d096):

```
Total number of files that use LF line endings: 1,491 (86%)
Total number of files that use CRLF line endings: 62 (3%)
Total number of files: 1,724
```

It would be nice if all of the files in this repo used the same line endings. Are there any files that need to use LFs or need to use CRLFs? For example, I can see that `scripts/data/linuxfullhog/d3i.str` uses CRLFs. Would using LFs in that file break anything?

GravisZro:
This is fixed in PR #145 with a bunch of text file encodings as well which prevented them from being edited on my machine.

JeodC:
It is simple to add a `.gitattributes` file to take care of this, but scripts may depend on CRLF endings to function. It would need to be tested on the other platforms.

PR #145, opened 2024-04-22
Convert iso-8859-1 encoded files to UTF8 encoding

GravisZro:
All files that were iso-8859-1 have been converted to UTF8 using this script.
```
for filename in $(find)
do
  encoding=$(file -b --mime-encoding $filename)
  if [ "$encoding" = "iso-8859-1" ]; then
    iconv -f iso-8859-1 -t utf8 ${filename} -o ${filename}.utf8
    mv ${filename}.utf8 ${filename}
  fi
done
```


winterheart:
Converting non-English messages may break them in game. Need test it on runtime.

Descent3/macmain.cpp
```diff
@@ -138,15 +138,15 @@ const char *French_strings[] = {
     "Votre processeur et votre systFme doivent prendre en charge Katmai pour exTcuter ce jeu."};
 
 const char *German_strings[] = {
-    "Descent3 unter Windows NT ben�tigt fnr die Ausfnhrung die NT-Version 4.0 oder h�her",
-    "Descent 3 ben�tigt fnr die Ausfnhrung Windows 9x, NT 4.0 oder h�her.",
+    "Descent3 unter Windows NT ben÷tigt fnr die Ausfnhrung die NT-Version 4.0 oder h÷her",
```
frank-aurich, line 141: The German umlauts are still looking wonky.

Descent3/macmain.cpp
```diff
@@ -338,7 +338,7 @@ int main(int argc, char *argv[]) {
 
   Descent = d3;
 
-  // � Initialize Metrowerks Code Profiler
+  // ¥ Initialize Metrowerks Code Profiler
```
Jayman2000, line 341: Is this supposed to be a Yen sign?

Jayman2000 (changes requested):
- I recommend putting `fixes #145` into the description of this pull request [so that that issue automatically gets closed when this PR is merged](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue).
- I think that you should split this change into two different commits, one that converts things to UTF-8, and one that switches to LF line endings. We may find a bug in the future that is caused by switching to UTF-8, but not caused by switching to LFs. If we do end up finding such a bug, it will be easier to use [`git-bisect`](https://git-scm.com/docs/git-bisect) to figure out what’s going on if these two big changes are in separate commits.

Descent3/macmain.cpp
winterheart: Yes, it's how IDE (CodeWarrior?) on old Macs was denoted documentation in the code.

Jayman2000: Oh, OK. I would mark this conversation as resolved, but [apparently I can’t do that because I wasn’t the one who submitted this PR](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/commenting-on-a-pull-request#resolving-conversations).

JeodC:
> * I recommend putting `fixes #145` into the description of this pull request [so that that issue automatically gets closed when this PR is merged](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue).
> * I think that you should split this change into two different commits, one that converts things to UTF-8, and one that switches to LF line endings. We may find a bug in the future that is caused by switching to UTF-8, but not caused by switching to LFs. If we do end up finding such a bug, it will be easier to use [`git-bisect`](https://git-scm.com/docs/git-bisect) to figure out what’s going on if these two big changes are in separate commits.

Do you mean #129?

GravisZro:
> Converting non-English messages may break them in game. Need test it on runtime.

@winterheart How can I start D3 in a specific language?

JeodC:
> > Converting non-English messages may break them in game. Need test it on runtime.
> 
> @winterheart How can I start D3 in a specific language?

I think it uses your system language.

Jayman2000:
> > * I recommend putting `fixes #145` into the description of this pull request [so that that issue automatically gets closed when this PR is merged](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue).
> > * I think that you should split this change into two different commits, one that converts things to UTF-8, and one that switches to LF line endings. We may find a bug in the future that is caused by switching to UTF-8, but not caused by switching to LFs. If we do end up finding such a bug, it will be easier to use [`git-bisect`](https://git-scm.com/docs/git-bisect) to figure out what’s going on if these two big changes are in separate commits.
> 
> Do you mean #129?

Yes.

PR #146, opened 2024-04-22
cmake: modernize, using safer/clearer options

scivision:
CMAKE_BUILD_TYPE must be set CACHE before project() to take effect

general best practice rather than modifying CMake internal variables is to use add_compile_options with genex

add_definitions => add_compile_definitions


winterheart:
It's worth to use `CMAKE_SYSTEM_NAME` in platform checks instead of `WIN32`, `UNIX AND NOT APPLE` checks.

CMakeLists.txt
```diff
@@ -25,25 +30,29 @@ if(NOT MSVC)
...
-    elseif(D3_COMPILER_IS_CLANG)
-      add_compile_options (-fcolor-diagnostics)
+    if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.24)
+      set(CMAKE_COLOR_DIAGNOSTICS ON)
+    else ()
+      if(CMAKE_COMPILER_IS_GNUCC)
+        add_compile_options (-fdiagnostics-color=always)
+      elseif(D3_COMPILER_IS_CLANG)
+        add_compile_options (-fcolor-diagnostics)
+      endif ()
```
winterheart, line 40: We can just update `cmake_minumum_required` to 3.24 and simplify block to just `set(CMAKE_COLOR_DIAGNOSTICS ON)`.
I'd prefer remove this block entirely (and compiler detection above) as color output can be enabled just by `-DCMAKE_COLOR_DIAGNOSTICS=ON` on config time.

Lgt2x: thanks, it's better now. However the Windows CI build seems to fail now

CMakeLists.txt
```diff
@@ -52,36 +48,32 @@ if(UNIX)
...
+if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
   message("Building for Linux")
-  add_definitions( -D_DEBUG -D__LINUX__ -DLINUX -D_MAX_PATH=260 -D_MAX_FNAME=256 -D_REENRANT -D__32BIT__ -DHAVEALLOCA_H -D_USE_OGL_ACTIVE_TEXTURES)
+  add_compile_definitions(_DEBUG __LINUX__ LINUX _MAX_PATH=260 _MAX_FNAME=256 _REENRANT __32BIT__ HAVEALLOCA_H _USE_OGL_ACTIVE_TEXTURES)
   set(PLATFORM_INCLUDES "lib/linux" ${SDL_INCLUDE_DIR})
-endif()
-
-if(APPLE)
+elseif(CMAKE_SYSTEM_NAME STREQUAL "Darwin")
   message("Building for MAC OSX")
```
th1000s, line 56: I think these messages can be removed as well, it's not like someone is going to cross compile Linux <-> macOS or has forgotten what system they are on :)

CMakeLists.txt
```diff
@@ -108,15 +100,13 @@ add_subdirectory(cfile)
 add_subdirectory(czip)
 add_subdirectory(d3music)
 
-if(WIN32)
+if(CMAKE_SYTEM_NAME STREQUAL "Windows")
```
winterheart, line 103: Here typo that prevents build Windows, there should be `CMAKE_SYSTEM_NAME` (missed second "S" in SYSTEM).

winterheart (changes requested):

JeodC:
@scivision

CMakeLists.txt
```diff
@@ -1,49 +1,45 @@
...
-
-if(NOT MSVC) # GCC/clang or compatible, hopefully
-  option(FORCE_COLORED_OUTPUT "Always produce ANSI-colored compiler warnings/errors (GCC/Clang only; esp. useful with ninja)." OFF)
+# set default cmake build type to Debug (None Debug Release RelWithDebInfo MinSizeRel)
+if(NOT CMAKE_BUILD_TYPE AND NOT DEFINED ENV{CMAKE_BUILD_TYPE})
+  set(CMAKE_BUILD_TYPE "Debug" CACHE STRING "default build type")
 endif()
 
+project(Descent3
+LANGUAGES C CXX
```
Lgt2x, line 9: indent 2 tabs

PR #148, opened 2024-04-22
Bypass launcher

JeodC:
This pull allows bypassing the launcher to run Descent3 directly with default assumed settings. It also corrects some language string errors in `winmain.cpp`. This was created to make drop-in testing more accessible. It can easily be hidden behind the DEBUG build later.

```cpp
if (!FindArg("-launched") && !FindArg("-dedicated") && !FindArg("-timetest")) {
    int result = MessageBox(NULL, "Bypassing the launcher might break some settings. Do you want to continue?", PRODUCT_NAME " Error", MB_YESNO);
    if (result == IDYES) {
        // Check if ShellExecute with "runas" is available
        HINSTANCE hShell32 = LoadLibrary("Shell32.dll");
        if (hShell32 != NULL) {
            FARPROC pfnShellExecute = GetProcAddress(hShell32, "ShellExecuteA");
            if (pfnShellExecute != NULL) {
                // ShellExecute with "runas" is available, use it
                ShellExecute(NULL, "runas", GetCommandLine(), NULL, NULL, SW_SHOWNORMAL);
                FreeLibrary(hShell32);
                return 0;
            }
            FreeLibrary(hShell32);
        }
        // ShellExecute is not available, run the application with standard privileges
        return 0;
    } else {
        return 0;
    }
}
```


JeodC:
For some reason this isn't working, nothing happens after clicking Yes. Since we can use terminal just as easily, it's kind of moot.

PR #150, opened 2024-04-22
Use new issue forms

JeodC:
Convert the issue templates to the new [issue forms](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository). An example of how these look and work can be seen [here](https://github.com/JeodC/Descent3/issues/new/choose).


Lgt2x:
Wow that's a great new feature! I'll merge and give them a try

PR #151, opened 2024-04-22
Fix endline characters

### LINKED ISSUE #129: Inconsistent line endings

GravisZro:
Replace \r\n with \n.
```
#!/bin/sh
for filename in $(find)
do
  encoding=$(file -b --mime-encoding $filename)
  if [ "$encoding" = "us-ascii" ] ||
     [ "$encoding" = "iso-8859-1" ]; then
    echo fixing ${filename}
    sed 's/\r$//' ${filename} > ${filename}.lf
    mv ${filename}.lf ${filename}
done
```
Fixes #129


GravisZro:
The file encodings are not altered by this patch.

GravisZro:
@Lgt2x By properly generated do you mean the game or the editor?

Lgt2x:
the game when creating the HOGs

JeodC:
I just played through Level 5 with this build, I think it's fine.

PR #152, opened 2024-04-22
Fixes for missing movies

### LINKED ISSUE #149: [Runtime Bug/Crash] Movie files required
This is the minimum file structure needed to play Descent 3:

```
Descent 3/
├── custom/
│   ├── cache
│   ├── graphics
│   ├── settings
│   └── sounds
└── missions/
└── d3.hog
└── Descent3.exe
```

Without the `movies/mainmenu.mve` file, the main menu ui will overlap. This might be because the movie is constantly drawn overtop of inactive ui windows, so if it isn't present, there is nothing to hide the ui windows.

Without the other `.mve` files, the player will be prompted for a CD on attempting to start `Descent 3: Retribution`, `Training Mission`, or `Descent 3: Mercenary`.

Since the engine is open source, we should not require a CD--we should simply not play the movies.

InsanityBringer:
If the mainmenu video file isn't present, revert to using a static background image in the main menu.
If one of the built in mission videos can't be found, don't prompt for the Descent 3 CD-ROM.
Fixes #149

### DIFF
--- a/Descent3/Mission.cpp
+++ b/Descent3/Mission.cpp
@@ -1671,20 +1671,9 @@ void DoMissionMovie(char *movie) {
   return;
 #endif
   if (movie && *movie) {
-    char *moviepath;
-
-    if (Current_mission.filename &&
-        !(!stricmp(Current_mission.filename, "d3.mn3") || !stricmp(Current_mission.filename, "d3_2.mn3"))) {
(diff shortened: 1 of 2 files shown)


JeodC:
Tested with windows, working now. Thank you!

PR #153, opened 2024-04-22
HogMaker: don't modify data when printing it

th1000s:
- Create reserved bytes when constructing HogHeader
- Only byteswap a copy of the data
- Other small cleanups

---

I was about to leave comments to this effect on #143, but then it got merged :)

Verified to produce the same result (if timestamp is set to 0).


Lgt2x:
sorry for merging at the wrong time, thanks for the cleanup!

Issue #157, opened 2024-04-22
Enable console logging


JeodC:
Per Kevin Bentley, it may be wise to modify cmake to do `-DCMAKE_BUILD_TYPE=debug` or `-DLOGGER`. Log output to the console for better linux/mac troubleshooting. Perhaps provide instructions to the user in an issue template on how to utilize it.

jcoby:
logging was removed in #77 because the `mprintf_at` macro wasn't expanding to `Debug_ConsolePrintf(int n, int row, int col, char *format, ...)` and causing a segfault. #159 reenables the option, renames the function to `Debug_ConsolePrintfAt` and adds the LOGGER define as well.

winterheart:
Perhaps this issue may be considered as solved?

JeodC:
> Perhaps this issue may be considered as solved?

That's what I'm wondering. I think it's ok to merge.

PR #159, opened 2024-04-23
Enable console logging if MONO or LOGGER is defined

jcoby:
Re-enable console logging on linux and mac builds with the -DMONO option.

Rename overloaded Debug_ConsolePrintf function to
Debug_ConsolePrintfAt to fix macro expansion.

Adds -DLOGGER alias for -DMONO.

Implements #157


CMakeLists.txt
```diff
@@ -54,13 +54,13 @@ endif()
...
   message("Building for Linux")
-  add_definitions( -D_DEBUG -D__LINUX__ -DLINUX -D_MAX_PATH=260 -D_MAX_FNAME=256 -D_REENRANT -D__32BIT__ -DHAVEALLOCA_H -D_USE_OGL_ACTIVE_TEXTURES)
+  add_definitions( -D_DEBUG -D__LINUX__ -DLINUX -D_MAX_PATH=260 -D_MAX_FNAME=256 -D_REENRANT -D__32BIT__ -DHAVEALLOCA_H -D_USE_OGL_ACTIVE_TEXTURES -DLOGGER)
   set(PLATFORM_INCLUDES "lib/linux" ${SDL_INCLUDE_DIR})
 endif()
 
 if(APPLE)
   message("Building for MAC OSX")
-  add_definitions(-D_DEBUG -D__LINUX__ -DLINUX -D_MAX_PATH=260 -D_MAX_FNAME=256 -D_REENRANT -DMACOSX=1 -D_USE_OGL_ACTIVE_TEXTURES)
+  add_definitions(-D_DEBUG -D__LINUX__ -DLINUX -D_MAX_PATH=260 -D_MAX_FNAME=256 -D_REENRANT -DMACOSX=1 -D_USE_OGL_ACTIVE_TEXTURES -DLOGGER)
```
Lgt2x, line 63: we want this definition conditionally, if the user specified the CMake [Option](https://cmake.org/cmake/help/latest/command/option.html) LOGGER

lib/mono.h
```diff
@@ -66,12 +66,12 @@
 #include "debug.h"
 void nw_InitTCPLogging(char *ip, unsigned short port);
 void nw_TCPPrintf(int n, char *format, ...);
-#if (!defined(RELEASE)) && defined(MONO)
+#if (!defined(RELEASE)) && (defined(MONO) || defined(LOGGER))
```
Lgt2x, line 69: I don't think we need to keep both, `LOGGER` is enough

jcoby: I wanted to keep both so we could keep the MONO code for historical reasons if we add a logging library. But it's easy enough to resurrect MONO later.

CMakeLists.txt
jcoby: Added in latest commit

Lgt2x:
Tested, works great. Thanks!

PR #165, opened 2024-04-23
Implementing C++ byteswap functions

winterheart:
Implementing byteswap functions by using "backported" from C++23 std::byteswap. Adding unittests based on GoogleTests. To enable it add `-DBUILD_TESTING=ON` to cmake.

PR #168, opened 2024-04-23
Use compiler intrinsic byteswapping functions

GravisZro:
I'm not trying to be stubborn but the current C++ byteswap code is a lot slower than using compiler intrinsic functions.

Converts INTEL_* and MOTOROLA_* macros to use the proper compiler intrinsic functions.

**List of macro aliases:**
From POSIX for networking conversions:
* host to network (16-bit) : htons()
* host to network (32-bit) : htonl()
* network to host (16-bit) : ntohs()
* network to host (32-bit) : ntohl()

From GNU for endian conversions:
* host to little endian (16-bit) : htole16()
* host to little endian (32-bit) : htole32()
* host to little endian (64-bit) : htole64()
* little endian to host (16-bit) : le16toh()
* little endian to host (32-bit) : le32toh()
* little endian to host (64-bit) : le64toh()

### Comparison to existing code

You can see the [code and corresponding assembly output here](
https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:15,endLineNumber:17,positionColumn:15,positionLineNumber:17,selectionStartColumn:15,selectionStartLineNumber:17,startColumn:15,startLineNumber:17),source:'%23include+%3Ccstdint%3E%0A%23include+%3Ccstdlib%3E%0A%23include+%3Cendian.h%3E%0A%23include+%3Cbyteswap.h%3E%0A%0Anamespace+D3%0A%7B%0A//+std::byteswap+from+C%2B%2B23%0A++++template+%3Ctypename+T%3E+constexpr+T+byteswap(T+n)+%7B%0A++++++++T+m%3B%0A++++++++for+(size_t+i+%3D+0%3B+i+%3C+sizeof(T)%3B+i%2B%2B)%0A++++++++++++reinterpret_cast%3Cuint8_t+*%3E(%26m)%5Bi%5D+%3D+reinterpret_cast%3Cuint8_t+*%3E(%26n)%5Bsizeof(T)+-+1+-+i%5D%3B%0A++++++++return+m%3B%0A++++%7D%0A%7D%0A%0Aint+main(void)%0A%7B%0A++++uint16_t+j+%3D+0%3B%0A++++j+%3D+D3::byteswap(j)%3B%0A++++j+%3D+bswap_16(j)%3B%0A++++return+j%3B%0A%7D'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:33.333333333333336,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:gsnapshot,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'1',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!(),options:'',overrides:!((name:stdver,value:c%2B%2B17)),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+x86-64+gcc+(trunk)+(Editor+%231)',t:'0')),k:36.71649619722003,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'llvm-mos+nes-mmc3',editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'0'),l:'5',n:'0',o:'Output+of+x86-64+gcc+(trunk)+(Compiler+%231)',t:'0')),k:29.950170469446622,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4
)

**C++ byteswap implementation**
```asm
unsigned short D3::byteswap<unsigned short>(unsigned short):
        push    rbp
        mov     rbp, rsp
        mov     eax, edi
        mov     WORD PTR [rbp-20], ax
        mov     QWORD PTR [rbp-8], 0
        jmp     .L12
.L13:
        mov     eax, 1
        sub     rax, QWORD PTR [rbp-8]
        lea     rdx, [rbp-20]
        add     rax, rdx
        lea     rcx, [rbp-10]
        mov     rdx, QWORD PTR [rbp-8]
        add     rdx, rcx
        movzx   eax, BYTE PTR [rax]
        mov     BYTE PTR [rdx], al
        add     QWORD PTR [rbp-8], 1
.L12:
        cmp     QWORD PTR [rbp-8], 1
        jbe     .L13
        movzx   eax, WORD PTR [rbp-10]
        pop     rbp
        ret
```

**compiler intrinsic byteswap implementation**
```asm
__bswap_16:
        push    rbp
        mov     rbp, rsp
        mov     eax, edi
        mov     WORD PTR [rbp-4], ax
        movzx   eax, WORD PTR [rbp-4]
        rol     ax, 8
        pop     rbp
        ret
```


JeodC:
With #165 merged, is this now stale?

GravisZro:
> With #165 merged, is this now stale?
No, #165 is really slow (like 10x or more) which is what I point out the post under "C++ byteswap implementation".

GravisZro:
@winterheart 0 ns for execution time is impossible. I'm guessing the compiler optimized out the byteswapping entirely. I'm working on a test and I'll post the results soon.

winterheart:
It's just rounded to 3th sign. Here CSV output to demonstrate how small numbers are:

```csv
name,iterations,real_time,cpu_time,time_unit,bytes_per_second,items_per_second,label,error_occurred,error_message
"CppByteswap16",1000000000,1.96043e-07,2.44e-07,ns,,,,,
"CppByteswap32",1000000000,1.77999e-07,1.77e-07,ns,,,,,
"CppByteswap64",1000000000,1.95985e-07,1.96e-07,ns,,,,,
"CppByteswapFloat",1000000000,2.44007e-07,1.92e-07,ns,,,,,
"CppByteswapDouble",1000000000,1.72993e-07,1.85e-07,ns,,,,,
"BSwap16",1000000000,1.93017e-07,1.79e-07,ns,,,,,
"BSwap32",1000000000,1.73983e-07,2.41e-07,ns,,,,,
"BSwap64",1000000000,1.93017e-07,1.79e-07,ns,,,,,
```

GravisZro:
Here's the results of my test on x86_64 using the settings used to compile Descent 3. [The code is here.](https://gist.github.com/GravisZro/d397b1c7c09712d58581c48ae29615e4)

```
$ g++ -O2 swapper.cpp -o swapper.g++
$ sudo ./swapper.g++ 2> /dev/null
Performing 16777215 swaps per test.
D3::byteswap<uint16_t>() took 1731300172 cycles, an average of 103.194 cycles per call
D3::byteswap<uint32_t>() took 1754326965 cycles, an average of 104.566 cycles per call
D3::byteswap<uint64_t>() took 1691950672 cycles, an average of 100.848 cycles per call
bswap_16() took 77076338 cycles, an average of 4.59411 cycles per call
bswap_32() took 77121710 cycles, an average of 4.59681 cycles per call
bswap_64() took 77272393 cycles, an average of 4.60579 cycles per call
bswap_16 was 22.4622 times faster.
bswap_32 was 22.7475 times faster.
bswap_64 was 21.8959 times faster.
```

GravisZro:
> It's fast, clean, and don't need any maintenance in future - any system with C++11 capable compiler can run it.

I'm happy to make it a fallback function but it seems to be too slow to be the primary. Most processors have a byteswapping opcode, so it's best to trust the compiler on this matter and use a software implementation as a backup.

winterheart:
Here:
```cpp
  uint64_t start, stop;
  get_time(start);
  do {
    a = D3::byteswap(start^a); // why??
  } while(--count);
  get_time(stop);
```

`(start^a)` will be int64_t, so you comparing <int64_t>D3::byteswap() (which converts back and forth values twice) with bswap_16. And why you counts math calculations in benchmark? With this sample

```cpp
  get_time(start);
  do {
    a = D3::byteswap<uint16_t>(start^a); // Force using correct type
  } while(--count);
  get_time(stop);
```
and swaps_per_test = 1000000000

I get these results
```
Performing 16777215 swaps per test.
D3::byteswap<uint16_t>() took 24410510 cycles, an average of 1.45498 cycles per call
D3::byteswap<uint32_t>() took 24437341 cycles, an average of 1.45658 cycles per call
value 16: 32673
value 32: 1124674387
value 64: 1041466293910909
value 16: 49904
value 32: 453698981
D3::byteswap<uint64_t>() took 36597026 cycles, an average of 2.18135 cycles per call
bswap_16() took 24387177 cycles, an average of 1.45359 cycles per call
bswap_32() took 24459049 cycles, an average of 1.45787 cycles per call
value 32: 9042890989104306419
bswap_64() took 36578798 cycles, an average of 2.18027 cycles per call
bswap_16 was 1.00096 times faster.
bswap_32 was 0.999112 times faster.
bswap_64 was 1.0005 times faster.
```

It's not a much differences.

Arcnor:
It's funny that you blame @winterheart's test of being flawed because it's been optimized away, yet don't realize your test is flawed in that same way (and this is why creating microbenchmarks is difficult, and using something like Google's Caliper makes sense).

In a nutshell, @winterheart 's code is being converted into a single `bswap` instruction (with optimizations on) as compilers are smart enough to realize what's happening. However, your code is being converted into an unrolled loop with multiple `bswap` instructions, which is great if you're going to byteswap in a loop, but that's just not realistic.

You can check the generated assembly if you want, but as stated before, in the grand scheme of things this doesn't make any difference, while introducing a huge amount of changes, so while your contrubution is of course appreciated, I think we can close this.

GravisZro:
@winterheart what is the compilation command you are using? It makes all the difference because D3 is not compiled with -O3.

winterheart:
I've used these options:

```
[1/2] /usr/bin/g++   -O3 -DNDEBUG -std=c++17 -fPIE -fdiagnostics-color=always -MD -MT tests/CMakeFiles/bench.dir/bench.cpp.o -MF tests/CMakeFiles/bench.dir/bench.cpp.o.d -o tests/CMakeFiles/bench.dir/bench.cpp.o -c /home/winterheart/playground/Descent3/tests/bench.cpp
[2/2] : && /usr/bin/g++ -O3 -DNDEBUG  tests/CMakeFiles/bench.dir/bench.cpp.o -o tests/bench   && :
```

GravisZro:
@winterheart Like I wrote, D3 is compiled with -O2. We don't use -O3.

GravisZro:
@Arcnor I checked the assembly output and it didn't unroll the loop. Are you sure you using -O2 and not -O3? We don't use -O3 for compiling D3. Kinda seems like you are dismissing me without cause. I'm not happy about trying to improve the project an getting rebuffed by with bogus claims.

JeodC:
There is a fine line between improvements and excess. Let's prioritize practicality and avoid over-engineering. Considering the slim likelihood of Descent 3 running on a system requiring big endian byteswap, it may be more efficient to focus our efforts on areas that directly impact the majority of users. If there's a specific desire or need for big endian support, that could be explored separately, perhaps through a fork of the project.

GravisZro:
@JeodC I'm literally targeting a low-end BE system that doesn't have a lot of excess computing power. Frankly, anything I've done that doesn't align exactly with the core group has been dismissed out of hand or ignored entirely. A total lack of compromise (especially in favor of ideology) is not my idea of a good direction.

JeodC:
> @JeodC I'm literally targeting a low-end BE system that doesn't have a lot of excess computing power. Frankly, anything I've done that doesn't align exactly with the core group has been dismissed out of hand or ignored entirely. A total lack of compromise (especially in favor of ideology) is not my idea of a good direction.

It is not a favor of ideology. This organization's mission statement has been made clear, and you are not the only one with personal goals. We're working with contributors who want to add their own ideas into the mix like modern rendering options and imgui--but those things are not on the docket yet.

I came to the Descent community due to my wish to bring dxx-rebirth to SBC handhelds. Ever since I fulfilled that wish I've wanted to see Descent 3 be more portable via using embedded systems instead of x11 or wayland and use gles. However I understand that other things should happen first--if I were dead set on it, I would create a fork and work to my own ends there.

You've had other contributions merged. Compromise is not a one-way street.

GravisZro:
> You've had other contributions merged. Compromise is not a one-way street.

I've been doing this stuff for over two decades and I had to fight @winterheart just to allow me to make sure that stdcall was in the correct places because he doesn't understand how x86 calls work and called a working solution "a big ugly hack" because it used preprocessor statements which were literally the only way of determining the architecture consistently. Instead, he was ready to abandon everything not x86_64 (stdcall breaks non-x86/x86_64 platforms and if it's not right then x86 goes out the window too) because he didn't know, didn't care to learn, and was barely just barely willing to accept a working patch.

Why do I have to fight with people over things I have more knowledge and experience in dealing with?

DanielGibson:
I like the idea of using compiler intrinsics, but IMHO your changes are way to intrusive.
I see no value in renaming all those macros (INTEL_INT, OUTRAGE_BIG_ENDIAN etc) and thus all the code that uses them.

Furthermore, you didn't write portable_endian.h yourself, even if the original author put it under public domain, IMO he should be credited.

And I wonder why you don't use [GCC's byteswapping intrinsics](https://gcc.gnu.org/onlinedocs/gcc-4.8.2/gcc/Other-Builtins.html#Other-Builtins) (__builtin_bswap16, __builtin_bswap32, __builtin_bswap64).

If I wanted to introduce compiler intrinsics for byte swapping, I'd just modify the existing byteswap.h to use them (`#ifdef _MSC_VER .. implement SWAPINT, SWAPSHORT and SWAPFLOAT with visualc++-specific intrinsics .. #elif defined(_GNUC_) .. same with GCC/clang intrinsics (__builtin_bswap*) #else .. old generic SWAP* code .. #endif` - and keep on using OUTRAGE_BIG_ENDIAN, which is set by cmake so it should be correct and is cleaner than trying to detect endianess with lots of `#if defined(whatever)` checks

DanielGibson:
But eventually Descent should just use SDL2 (or SDL3) on all platforms, then it could use SDL_endian.h which already implements all the swapping for all the compilers and platforms (using intrinsics or inline ASM), and byteswap.h could be reduced to:

```
#include <SDL_endian.h>

#define SWAPSHORT(x) SDL_Swap16(x)
#define SWAPINT(x)   SDL_Swap32(x)
#define SWAPFLOAT(x) SDL_SwapFloat(x)

#define INTEL_INT(x)   SDL_SwapLE32(x)
#define INTEL_SHORT(x) SDL_SwapLE16(x)
#define INTEL_FLOAT(x) SDL_SwapFloatLE(x)

#define MOTOROLA_INT(x)   SDL_SwapBE32(x)
#define MOTOROLA_SHORT(x) SDL_SwapBE16(x)
#define MOTOROLA_FLOAT(x) SDL_SwapFloatBE(x)

// optionally also the next few lines, though this can as well be handled from CMake
#if SDL_BYTEORDER == SDL_BIG_ENDIAN
  #define OUTRAGE_BIG_ENDIAN
#endif
```

winterheart:
> @Arcnor I checked the assembly output and it didn't unroll the loop. Are you sure you using -O2 and not -O3? We don't use -O3 for compiling D3. Kinda seems like you are dismissing me without cause. I'm not happy about trying to improve the project an getting rebuffed with bogus claims.

Actually we do `-O3` on release CI. Just look for compilation log. macOS and Linux uses `-O3`.

Arcnor:
> @Arcnor I checked the assembly output and it didn't unroll the loop. Are you sure you using -O2 and not -O3? We don't use -O3 for compiling D3. Kinda seems like you are dismissing me without cause. I'm not happy about trying to improve the project an getting rebuffed with bogus claims.

I'm definitely not dismissing you without cause, I've stated my reasons on why your test is flawed, and even if it was not I've already mentioned how this is too intrusive for something that doesn't change much in the end. Also, I'm testing on Windows with MSVC and there is no `O3`.

As mentioned, we appreciate your contributions, like everyone else's, but sometimes changes don't pan out.

PR #171, opened 2024-04-24
Enable unittests on CI

winterheart:
Unittests requires optional dependency, and most users don't have to install it. Therefore, GTest is not included to vcpkg and brew files, but CI does.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,6 +17,7 @@ on:
 
 jobs:
   build:
+    name: ${{ matrix.os.runner }}, ${{ matrix.os.cxx }}, ${{ matrix.build_type }}
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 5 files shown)


Lgt2x:
Would it be possible to separate both workflows ? First build, then test in a separate job.

JeodC:
> Would it be possible to separate both workflows ? First build, then test in a separate job.

You could do this with `workflow_dispatch:` to trigger tests manually, but to link two workflows together I believe requires a github token which only the repository owner (Kevin) has access to at the moment.

winterheart:
> Would it be possible to separate both workflows ? First build, then test in a separate job.

Well, better then return build types into matrix, so I can write 2 jobs instead of 4:

```yaml
      - name: ${{ matrix.os.build_type }} Build
        run: cmake --build --preset ${{ matrix.os.preset }} --config ${{ matrix.os.build_type }} --verbose

      - name: ${{ matrix.os.build_type }} Test
        run: ctest --preset ${{ matrix.os.preset }} -C ${{ matrix.os.build_type }}
```

winterheart:
> IMO it's a better solution

Ok, will rewrite pipeline with build types in matrix.

Lgt2x (approved):

PR #176, opened 2024-04-24
Enable building 32-bit binaries on 64-bit Linux via toolchain file

th1000s:
Works with variables passed on the commandline via `-D CMAKE_..`, flags     from https://wiki.debian.org/Multiarch/Compiling

---

~~Avoids adding a toolchain file.~~ - It does not avoid this at all!

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -46,7 +46,17 @@ jobs:
             preset: linux
             cc: clang
             cxx: clang++
-            name: Linux-x64-clang
+            name: no_upload
+          - runner: ubuntu-latest
+            preset: linux-32
+            cc: gcc
+            cxx: g++
+            name: Linux-i386
+          - runner: ubuntu-latest
+            preset: linux-32
+            cc: clang
+            cxx: clang++
+            name: no_upload
         build_type:
           - Debug
           - Release
@@ -69,6 +79,14 @@ jobs:
           sudo apt install -y --no-install-recommends \
             ninja-build cmake g++ libgtest-dev libsdl1.2-dev libsdl-image1.2-dev libncurses-dev zlib1g-dev
 
+      - name: Install Linux dependencies for 32-bit build
+        if: ${{ matrix.os.preset == 'linux-32' }}
+        run: |
+          sudo dpkg --add-architecture i386
+          sudo apt update
+          sudo apt install -y --no-install-recommends \
+            ninja-build cmake g++ g++-multilib libgtest-dev:i386 libsdl1.2-dev:i386 libsdl-image1.2-dev:i386 libncurses-dev:i386 zlib1g-dev:i386
+
       - name: Configure CMake
         env:
           CC: ${{ matrix.os.cc }}
@@ -83,6 +101,7 @@ jobs:
         run: ctest --preset ${{ matrix.os.preset }} -C ${{ matrix.build_type }}
 
       - name: Upload Artifacts
+        if: ${{ matrix.os.name != 'no_upload' }}
         uses: actions/upload-artifact@v4
         with:
           name: Descent3_${{ matrix.build_type }}_${{ matrix.os.name }}
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -37,6 +37,13 @@
 				"lhs": "${hostSystemName}",
 				"rhs": "Linux"
 			}
+		},
+		{
+			"name": "linux-32",
+			"inherits": "linux",
+                        "cacheVariables": {
+                                "CMAKE_TOOLCHAIN_FILE": "tools/toolchain-32-bit.cmake"
+                        }
 		}
 	],
 	"buildPresets": [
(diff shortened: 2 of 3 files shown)


Arcnor:
Can we do this in a different way? Not sure we need to start adding hacks, there has to be a proper way of doing it...

CMakeLists.txt
```diff
@@ -53,6 +53,16 @@ if(UNIX)
...
+  if(CMAKE_SYSTEM_NAME STREQUAL "Linux" AND BITS MATCHES "-m32")
+    # hacks required to enable 32 bit binaries with a simple -D BITS=-m32 on various 64 bit deb or rpm distros:
+    if(EXISTS "/etc/debian_version")
+      set(CMAKE_LIBRARY_ARCHITECTURE i386-linux-gnu)
+    else()
+      set(CMAKE_SIZEOF_VOID_P 4)
+    endif()
+  endif()
+
+
```
winterheart, line 65: Don't add this to CMakeLists.txt. CMAKE_SIZEOF_VOID_P sets by CMake and should'nt be changed by user. Instead of this block just add `"CMAKE_LIBRARY_ARCHITECTURE": "i386-linux-gnu"` into `cacheVariables` of linux-32 preset. Other distros has own approaches to define crosscompile environment.

th1000s:
>  there has to be a proper way of doing it...

That was my thought as well. However the Debian one is adapted from modules shipped with cmake, which set lib specific paths in a similar way. The rpm/Fedora one though...

When setting these values from the outside (with `-D` directly or via `CMakePresets.json`) they will be overwritten and so the binaries are linked against the wrong 64 bit shared objects. Open to suggestions, any idea on that @bryanperris @DanielGibson @Lgt2x ?

I would argue that this is a small hack which will be removed later. For now it helps with tracking down 32/64 bit incompatibilities.

.github/workflows/build.yml
```diff
@@ -76,6 +88,7 @@ jobs:
         uses: actions/upload-artifact@v4
         with:
           name: build_${{ matrix.os.runner }}_${{ matrix.os.cxx }}
+          name: build_${{ matrix.os.runner }}${{ matrix.os.preset == 'linux-32' && '-32' || '' }}_${{ matrix.os.cxx }}
```
winterheart, line 91: You can change it with `name: build_${{ matrix.os.preset }}_${{ matrix.os.cxx }}`.

CMakeLists.txt
JeodC: Modifying `CMAKE_SIZEOF_VOID_P` directly in the CMakeLists.txt file isn't a recommended practice because it's a system-defined variable set by CMake based on the architecture it detects. Changing it could lead to unexpected behavior and compatibility issues with other parts of the build system or other projects.

Instead, setting `CMAKE_LIBRARY_ARCHITECTURE` within cache variables of a preset for Linux 32-bit is a cleaner approach. It provides a clearer separation of concerns and ensures that the necessary configurations for cross-compilation are applied in a more standard and maintainable manner.

th1000s:
Thanks @winterheart for the Gentoo cmake toolchain file hint.

I will update the README in a separate PR.

winterheart:
I think linux 32bit clang is redundant. We added clang to matrix for some sanity checks and asan features in future. Most users can use gcc build.

th1000s:
Yes, can I exclude clang from the artifacs with an `if: ${{ ..  }}` clause?

But for building, clang still has different warnings and might catch something in 32bit if we ever get to activating `-Wall -Wextra -Werror`. Oh and gcc has asan etc. as well.

Lgt2x:
@th1000s some conflicts in the workflow file need to be resolved now that unittests have been merged, but it seems good otherwise!

Lgt2x:
Now that Icculus' 64-bit changes have been merged (#233), do we still need a 32 bit build ?

GravisZro:
> Now that Icculus' 64-bit changes have been merged (#233), do we still need a 32 bit build ?

Of course. "64-bit safe" merely means the behavior is the same regardless of what it's built for.

Lgt2x:
> > Now that Icculus' 64-bit changes have been merged (#233), do we still need a 32 bit build ?
> 
> Of course. "64-bit safe" merely means the behavior is the same regardless of what it's built for.

No it's not obvious. The idea behind 64bit Linux builds was to ensure 64 bit was the same as 32 bit when we were porting the game to 64 bit. Now that Ryan's PR bringing 64 bit compatibility fixes is in, there is a lower chance that we will find bugs because of that. At this point, maintaining compatibility with 32 bit Linux makes less sense because it is "one more platform to support" and not "the reference implementation". This project aims to support current-day platforms, and 32 bit Linux is not one of them.

GravisZro:
> The idea behind 32bit Linux builds was to ensure 64 bit was the same as 32 bit when we were porting the game to 64 bit. Now that Ryan's PR bringing 64 bit compatibility fixes is in, there is a lower chance that we will find bugs because of that.

In that case, you should know that the fixes in this PR are not the only uses of architecture specific sizes.

JeodC:
Can we revisit this one and come to a decision?

Lgt2x:
Now that Windows is also on its way to being totally 64bit, we'd rather not maintain a full 32bit Linux build. I bet some things are already broken, I don't want to pretend that we support it.

PR #178, opened 2024-04-25
Update script CHECKSUM for 64 bit platforms

jcoby:
Osiris_CreateGameChecksum computes a different checksum on 64 bit architectures due to the game structures being larger, mostly from pointers being 8 bytes instead of 4.

Fixes #160

PR #179, opened 2024-04-25
Checksum unittests and Osiris types unification

winterheart:
Following #178, implementing Osiris checksum tests to ensure it matches after code updates. Also converted all integer types to stdint to unify them across all platforms.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,7 +21,7 @@ endif()
 
 # 64 bit machines have a different game checksum than 32 bit machines
 if (CMAKE_SIZEOF_VOID_P EQUAL 8)
-  add_definitions(-DCHECKSUM=2273889867UL)
+  add_definitions(-DCHECKSUM=2273889835UL)
 else()
   add_definitions(-DCHECKSUM=2273873307UL)
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,6 +18,14 @@ if(CMAKE_CXX_BYTE_ORDER STREQUAL "BIG_ENDIAN")
   add_definitions("-DOUTRAGE_BIG_ENDIAN")
 endif()
 
+
+# 64 bit machines have a different game checksum than 32 bit machines
+if (CMAKE_SIZEOF_VOID_P EQUAL 8)
+  add_definitions(-DCHECKSUM=2273889867UL)
+else()
+  add_definitions(-DCHECKSUM=2273873307UL)
+endif()
+
 if(BUILD_TESTING)
   find_package(GTest REQUIRED)
   enable_testing()
--- a/Descent3/CMakeLists.txt
+++ b/Descent3/CMakeLists.txt
@@ -309,3 +309,7 @@ endif()
 if(WIN32)
   set_target_properties(Descent3 PROPERTIES WIN32_EXECUTABLE ON)
 endif()
+
+if(BUILD_TESTING)
+  add_subdirectory(tests)
+endif()
--- a/Descent3/tests/porting-tests.cpp
+++ b/Descent3/tests/porting-tests.cpp
@@ -1,0 +1,54 @@
+/*
+ * Descent 3
+ * Copyright (C) 2024 Descent Developers
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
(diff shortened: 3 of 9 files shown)


bryanperris:
I think this PR should only focus on adding the tests, not also changing all the types, that should be its own PR.

bryanperris:
> > I think this PR should only focus on adding the tests, not also changing all the types, that should be its own PR.
> 
> We test the feature that's been changed, no wrong in that (and the commits are separate)

Then perhaps just change the title of this PR: Update to using std types for the Osiris scripting system.

This this PR is just called "checksum unittests" but it doesn't make it obvious that alot of types were just updated in critical parts of the core.

bryanperris:
Did you build and run this unit test before changes were made in the core?

winterheart:
> Did you build and run this unit test before changes were made in the core?

Yes, before and after type changes, both for 32 and 64 bit environment.

bryanperris:
> > Did you build and run this unit test before changes were made in the core?
> 
> Yes, before and after type changes, both for 32 and 64 bit environment.

Sounds good then.

Descent3/CMakeLists.txt
```diff
@@ -309,3 +309,7 @@ endif()
 if(WIN32)
   set_target_properties(Descent3 PROPERTIES WIN32_EXECUTABLE ON)
 endif()
+
+if(BUILD_TESTING)
+  add_subdirectory(tests)
```
Lgt2x, line 314: If possible I'd like to have all tests in a single place (root tests/ folder). I guess you need to include the Descent directory to find your imports there

Descent3/tests/porting-tests.cpp
```diff
@@ -0,0 +1,60 @@
...
+  value += sizeof(vector) * 23;
+  value += sizeof(matrix) * 29;
+  value += sizeof(tGameCinematic) * 31;
+  value += sizeof(tCannedCinematicInfo) * 37;
+
+  value = value << 2;
+
+/*
+  int index = 0;
+  while (Osiris_module_init.fp[index]) {
```
Lgt2x, line 47: do yo uneed to keep  that ?

winterheart: Replaced with comment why to need add 123 to checksum.

Descent3/CMakeLists.txt
winterheart: I'm slowly working on isolating modules from each other in order to able test them as static libraries with available APIs. For that I'd like keep this structure with own tests for each component.

Lgt2x: no problem

Descent3/tests/porting-tests.cpp
```diff
@@ -0,0 +1,54 @@
...
+ */
+
+#include <gtest/gtest.h>
+
+#include "osiris_predefs.h"
+#include "osiris_share.h"
+#include "player_external_struct.h"
+
+// This is copy of actual Osiris_CreateGameChecksum(void) from OsirisLoadandBind.cpp
+uint Osiris_CreateGameChecksumTest() {
```
jcoby, line 26: Is there a reason Osiris_CreateGameChecksumTest can't actually call Osiris_CreateGameChecksum? you'd have to call Osiris_CreateModuleInitStruct first. seems like it would be better to test the actual code instead of something almost but not exactly the running code.

CMakeLists.txt
```diff
@@ -18,6 +18,14 @@ if(CMAKE_CXX_BYTE_ORDER STREQUAL "BIG_ENDIAN")
   add_definitions("-DOUTRAGE_BIG_ENDIAN")
 endif()
 
+
+# 64 bit machines have a different game checksum than 32 bit machines
```
jcoby, line 22: IMO the CHECKSUM value was previously only in the scripts module because they're the ones that get the value blindly where the runtime computes it. not totally for or against moving either way.

Descent3/tests/porting-tests.cpp
bryanperris: Yeah very good point @jcoby, a unit test should be performing the check against the actual code used in game, not a copy. Then if someone breaks the real code, the test is still going to pass no matter what.

winterheart: To import Osiris_CreateGameChecksum() I need somehow import nearly half of project's code in order to just link it. Here goes cfile, mem, debug, sound components etc... For now I just wrote functions that reproduce Osiris_CreateGameChecksum() logic.
I hope in future code would be isolated enough to import Osiris_CreateGameChecksum() itself.

CMakeLists.txt
winterheart: I moved it to level up, so both script and Descent will have access for define. Logically Osiris_CreateGameChecksum() belongs to Descent, while script only consumes CHECKSUM.

This is not ideal solution, and I planning sometime just move some defines to generated by cmake header file.

Descent3/tests/porting-tests.cpp
Lgt2x: Yeah something is odd with the osiris includes, we'll need to clean that

Lgt2x:
Unit tests pass locally, they also should on CI. merging

PR #181, opened 2024-04-25
Slight improvement to misc/logfile.cpp and lib/logfile.h

Nakhr11n:
Printf wrapper from [https://github.com/fmtlib/fmt/issues/2391#issuecomment-869043202](https://github.com/fmtlib/fmt/issues/2391#issuecomment-869043202).

### DIFF
--- a/lib/logfile.h
+++ b/lib/logfile.h
@@ -16,26 +16,27 @@
 * along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-
 #ifndef LOGFILE_H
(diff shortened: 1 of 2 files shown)


winterheart:
Personally, I'd like to use something more feature-rich logger like spdlog or something based on fmt::fmt.
But, your changes will make current logger better.

Lgt2x:
@bryanperris is this change still relevant with #212 ?

bryanperris:
> @bryanperris is this change still relevant with #212 ?

No.

PR #186, opened 2024-04-26
Remove MacOS (Classic) code and all references

tophyr:
The MACINTOSH define refers to MacOS Classic (not OS X) which we do not plan to support. Rather than carry the cruft forever, let's delete it.

Test Plan:
On Mac, build both Debug and Release
```
cmake --build --preset mac --config Debug
cmake --build --preset mac --config Release
```


tophyr:
Abandoning in favor of https://github.com/DescentDevelopers/Descent3/pull/188, which contains this commit.

PR #188, opened 2024-04-26
Kill unused srcs

tophyr:
Remove a bunch of unused-code cruft.

Test Plan:
On all three of `[win, mac, linux]`:
```
cmake --preset <platform>
cmake --build --preset <platform> --config Debug
cmake --build --preset <platform> --config Release
```


Lgt2x:
Please do not remove the legacy folder, it is there for a reason. File endings do not need to change either (or at least in a separate PR)

JeodC:
How has this been verified as unused? The project uses C++ 17, is this based on incompatibility with that? We want to maintain some form of legacy compatibility if possible.

winterheart:
Didn't see whole changeset, but there mostly old Mac OS 9 code (`#ifdef MACINTOSH`) and files that not included to CMakeLists.txt files.

tophyr:
The first commit uses `unifdef` to remove any `MACINTOSH`-dependent preprocessor blocks, and the second commit does some string parsing on on `CMakeLists.txt` files to find all files that are _are_ referenced, and then deletes any `*.c(pp)?` that isn't part of that list.

Builds were performed (but resulting binaries not tested beyond that) on all of Linux, Mac and Windows for each commit.

Finally, `legacy/` was manually restored in each commit per @Lgt2x's advice.

tophyr:
am updating commits with process for posterity, don't merge yet

InsanityBringer:
One request I want to make, the files in the renderer dir "Direct3D.cpp", "renderer.cpp", and "opengl.cpp" should be moved to a new renderer dir in the legacy dir, since these are potentially important for reference, since the different renderers have different features.

tophyr:
@InsanityBringer will do!

Descent3/GameLoop.cpp
```diff
@@ -1252,9 +1252,6 @@ void ProcessNormalKey(int key) {
   switch (key) {
 
     // ingnore the modifier keys
-#ifdef MACINTOSH
```
MaddTheSane, line 1255: This could be repurposed for Mac OS X.

Descent3/config.cpp
```diff
@@ -1049,9 +946,6 @@ struct sound_menu {
   };
 };
 
-#ifdef MACINTOSH
-#pragma mark toggles --
```
MaddTheSane, line 1053: `#pragma mark` macros are supported by compilers. You could probably remove the `MACINTOSH` check and just have it out bare.

tophyr:
Ok, each commit now has the process used to generate it ("Delete unused srcs" is a doozy) and the commits are better broken apart to show the steps. The renderer files @InsanityBringer requested are moved to `legacy/` for preservation.

JeodC:
tophyr ran his script in jmarshall's fork (the one improving the dev editor) and confirmed there are no editor dependencies removed in the script. This should be safe.

tophyr:
We gon do this one? The longer it stays, the more likely we'll run into merge conflicts.

JeodC:
If they aren't used, there shouldn't be merge conflicts. I believe this is safe to merge though.

tophyr:
> If they aren't used, there shouldn't be merge conflicts.

lol, good point - my concern was more around the `MACINTOSH` ifdef's.

PR #194, opened 2024-04-26
Fixup .gitignore

tophyr:
* Remove VSCode-internal settings files from exclusion from gitignore
  * (Why were these files excluded? They seem quite developer-specific.)
* Add gamedata/ dir for consolidating gamedata

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -380,10 +380,6 @@ FodyWeavers.xsd
 
 # VS Code files for those working on multiple tools
 .vscode/*
-!.vscode/settings.json
-!.vscode/tasks.json
-!.vscode/launch.json
-!.vscode/extensions.json
 *.code-workspace
 
 # Local History for Visual Studio Code
@@ -406,3 +402,6 @@ cmake-build-*
 # Linux stuff
 build
 builds
+
+# Gamedata collections
+gamedata/


bryanperris:
Where do you get `gamedata`?  Looking at my layout, the game data stuff is at the root of the directory like for example, hog files.

tophyr:
@bryanperris i copied my game data just into a `gamedata/` dir i created.

bryanperris:
There isn't a good ennough reason to commit that `gamedata` directory under .gitignore, as we aren't mixing the game assets into the cloned repository.  Once I have compiled the executable `Descent3, I copy it into my Descent3 directory that has been duplicated from Steam.

tophyr:
> Once I have compiled the executable Descent3, I copy it into my Descent3 directory that has been duplicated from Steam.

This gives a pretty poor debugging experience - there are multiple steps involved, and you typically need to copy the symbols and any shared libs as well.

While an alternative option is to simply create an out-of-tree directory for the gamedata or point the executable toward the retail installation with `-setdir`, both can be troublesome. An out-of-tree copy of gamedata is annoyingly non-self-contained, and using the retail installation directly can be troublesome - for example OS X doesn't treat `.app` packages as normal directories, and Steam may additionally compete for file locks or permissions.

@bryanperris what would you think of creating just a general `.ignored/` directory that individual developers could fill with whatever they see fit?

Arcnor:
If you want to ignore things specific to your computer or environment, you should use `.git/info/exclude`.

PR #195, opened 2024-04-27
Reorganizing fix library

winterheart:
Fixed point math library was part of the Vector math library until it was splitted to its own. Before that there was bundled version of vecmat code in scripts. This PR makes the library an isolated submodule that can be linked to any other.

Additional enhancements includes fixing minor warnings and removing unused Mac OS and Windows asm code.


Lgt2x (approved):
Looks legit. Tested on linux. Maybe a second pair of eyes could be useful @jcoby   @Arcnor ?

fix/fix.cpp
```diff
@@ -100,9 +97,6 @@ void InitMathTables() {
 
   asin_table[256] = asin_table[255];
   acos_table[256] = acos_table[255];
-
-  //	Initialize a random seed.
-  ps_srand(time(NULL));
```
Arcnor, line 105: So this is not needed to initialize the random number generator?

Lgt2x: they relocated seed initialization to init.cpp

winterheart: Random seed was injected into InitMathTables when it was part of vecmat submodule. Logically seed initialization not belongs to vecmat or fixmat components so I decided to move it out to common initialization function.

Issue #112, opened 2024-04-20
Move away from C style arrays


bryanperris:
There are many classic C arrays allocated with fixes sizes, and some conditions that simply lack sane bounds checking, this leads to many potential invalid memory access exceptions, and is the culprit load file crash for Linux.

Instead of C arrays, use C++ std::array or std::vector classes instead.

kevinbentley:
I think this is a good idea in theory, but I think it will take a massive amount of work. There are a lot of pointers to arrays used. Something I think we can look at in the future, but I think we have bigger fish to fry for a while.

bryanperris:
Isn't this part of moving towards modern C++?

Lgt2x:
We can enforce using modern C++ for newly added code, but re-writing existing code for the sake of it is likely to introduce more problems that it would solve

bryanperris:
I thought the whole codebase was basically going to be re-written into modern C++, not just mixing it.  I know for now, lots of legacy stuff just needs to be fixed to get things at stable point.  It is I thought of unit testing should be started early on, so later when classic code is moved over to modern C++, the tests can ensure things still correctly work.

Lgt2x:
Descent3 is around 1 million lines of C++, it'll be a while before we rewrite all of that =)

I do agree that unit tests are a priority, and we're trying to figure out the best way to do them.

bryanperris:
I understand there is a lot of code, bu how much of that is in core system?  What about moving the core engine into a separate library, like making it a cmake submodule.

Since we are using cmake here, you can leverage Gtest with cmake: https://google.github.io/googletest/quickstart-cmake.html

PR #196, opened 2024-04-27
Use vector instead of array, and proper referencing

tophyr:
Instead of using a hardcode-size array for app instances, use a vector instead. This cleans up some bad referencing as well.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,6 +13,8 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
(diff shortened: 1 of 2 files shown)


bryanperris:
See my closed issue: https://github.com/DescentDevelopers/Descent3/issues/112

Lgt2x:
CMakeLists.txt conflict needs resolved. Does this PR actually solve any compilation or runtime problem ?

winterheart:
CMakeLists.txt changes are not related and failing Windows build. Please remove them from PR.

Lgt2x:
In addition, please refrain from refactoring complete files when not absolutely necessary. We don't have a full testing pipeline right now, we're just trying really hard to get the base game working, and full file refactors can easily introduce problems. I'd prefer a simpler fix for getting Asan to work if possible. Thanks!

PR #197, opened 2024-04-27
Refactor and remove unnecessary in misc/pstring.cpp and dependents.

Nakhr11n:
## Pull Request Type

- [ ] GitHub Workflow changes
- [x] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [x] Other changes

### Description

Remove unnecessary code.

### Checklist

### DIFF
--- a/misc/pstring.cpp
+++ b/misc/pstring.cpp
@@ -39,37 +39,8 @@
  */
 
 #include <string.h>
-#include <stdarg.h>
-#include <stdio.h>
-#include <stdlib.h>
+
 #include "pstring.h"
-
-#ifdef __LINUX__
-#include "lnxfix.h"
-#endif
-
-//	Pvsprintf
-//		Similar to vsprintf/_vsnprintf, however handles the case of a buffer overflow.  Arguments are the
-//	same as _vsnprintf.  In the case of count, pass in the size of the buffer, in the case where there is
-//	a buffer overflow, a \0 will be tacked on as the last byte in the buffer, ensuring a valid string.
-int Pvsprintf(char *buffer, int count, const char *format, va_list argptr) {
-  int ret = _vsnprintf(buffer, count - 1, format, argptr);
-  if (ret == -1)
-    buffer[count - 1] = '\0';
-  return ret;
-}
-
-//	Psprintf
-//		Similar to sprintf/_snprintf, however handles the case of a buffer overflow.  Arguments are the
-//	same as _snprintf.  In the case of count, pass in the size of the buffer, in the case where there is
-//	a buffer overflow, a \0 will be tacked on as the last byte in the buffer, ensuring a valid string.
-int Psprintf(char *buffer, int count, const char *format, ...) {
-  va_list ap;
-  va_start(ap, format);
-  int ret = Pvsprintf(buffer, count, format, ap);
-  va_end(ap);
-  return ret;
-}
 
 // CleanupStr
 //		this function strips all leading and trailing spaces, keeping internal spaces.  this goes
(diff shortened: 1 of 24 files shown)


misc/pstring.cpp
```diff
@@ -38,140 +38,22 @@
...
-    } else if (ch > ' ' && err >= 1) {
-      end = i;
-    }
+//    This function strips all leading and trailing spaces, keeping internal spaces. This goes for tabs too.
+int CleanupStr(char *dest, const char *src, std::size_t destlen) {
+  if ((src != nullptr) && (dest != nullptr)) {
+    std::string s{src};
+    s.erase(s.begin(), std::find_if(s.begin(), s.end(), [](unsigned char ch) { return !std::isspace(ch); }));
+    s.erase(std::find_if(s.rbegin(), s.rend(), [](unsigned char ch) { return !std::isspace(ch); }).base(), s.end());
+    std::strncpy(dest, s.c_str(), destlen);
```
th1000s, line 55: Since you are benchmarking, a `std::string_view` and `remove_prefix/suffix` (since C++17) should be faster :)

winterheart:
```cpp
int CleanupStr(char *dest, const char *src, std::size_t destlen) {
  if ((src != nullptr) && (dest != nullptr)) {
    std::string_view str{src};
    auto it1 = std::find_if_not(str.begin(), str.end(), [](char ch) { return std::isspace(ch); });
    str = str.substr(it1 - str.begin());

    auto it2 = std::find_if_not(str.rbegin(), str.rend(), [](char ch) { return std::isspace(ch); });
    str = str.substr(0, str.rend() - it2);

    int size = str.size() < destlen ? str.size() : destlen;
    std::strncpy(dest, str.data(), size);
    dest[size] = '\0';
  }
```

Would be faster, but no so fast as C:

```c
int CleanupStr(char *dest, const char *src, size_t destlen) {
  if (destlen == 0)
    return 0;

  const char *end;
  size_t out_size;

  // Trim leading space
  while (isspace((unsigned char)*src))
    src++;

  // All white spaces
  if (*src == 0) {
    *dest = 0;
    return 0;
  }

  // Trim trailing space
  end = src + strlen(src) - 1;
  while (end > src && isspace((unsigned char)*end))
    end--;
  end++;

  out_size = (end - src) < destlen - 1 ? (end - src) : destlen - 1;

  memcpy(dest, src, out_size);
  dest[out_size] = 0;

  return 0;
}
```
(Not tested properly, just need careful check)

winterheart:
Need to split first commit into removing and changing code ones.

misc/pstring.cpp
```diff
@@ -38,140 +38,23 @@
...
-    } else if (ch > ' ' && err >= 1) {
-      end = i;
-    }
+//    This function strips all leading and trailing spaces, keeping internal spaces. This goes for tabs too.
+int CleanupStr(char *dest, const char *src, std::size_t destlen) {
+  if ((src != nullptr) && (dest != nullptr)) {
+    std::string_view s{src};
+    s.remove_prefix(std::count_if(s.begin(), s.end(), [](unsigned char ch) { return !std::isspace(ch); }));
+    s.remove_suffix(std::count_if(s.rbegin(), s.rend(), [](unsigned char ch) { return !std::isspace(ch); }));
+    std::strncpy(dest, s.data(), destlen);
```
winterheart, line 56: This is dangerous move. `string_view.data()` never returns null-terminated string, so at dest's end may be garbage. You had to manually detect end of string and add \0 to end.

misc/pstring.cpp
```diff
@@ -38,46 +38,36 @@
...
+      s.copy(dest, s.length());
+    } else if (s.back() == '\0' && s.length() > destlen) {
+      s.copy(dest, destlen - 1);
+      dest[destlen - 1] = '\0';
+    } else if (s.back() != '\0' && s.length() + 1 <= destlen) {
+      s.copy(dest, s.length());
+      dest[s.length()] = '\0';
+    } else if (s.back() != '\0' && s.length() + 1 > destlen) {
+      s.copy(dest, destlen - 1);
+      dest[destlen - 1] = '\0';
```
winterheart: This looks bulky. Have you tried my implementation?

```cpp
    int size = str.size() < destlen ? str.size() : destlen;
    std::strncpy(dest, str.data(), size);
    dest[size] = '\0';
```

winterheart:
Tested on Linux, thank you!

PR #198, opened 2024-04-27
Fix sound stuttering in MVE on Linux 64bit platforms

winterheart:
In MVE decoder on sound decoding occurs unnecessary long to int and int to long conversions with byteswap, which causes sound cuts.


winterheart:
OK, here may some license issues, and we should swap current MVE implementation to something other (#2).

PR #199, opened 2024-04-27
Unify Windows/Unix command line

th1000s:
Any` --foo` `-foo` and `+foo` is now accepted on any OS. Also enables single char arguments like `-f` everywhere. Previously `-foo` had to be passed as `--foo` on Linux, and macOS behaved like Windows.

### DIFF
--- a/Descent3/args.cpp
+++ b/Descent3/args.cpp
@@ -19,9 +19,8 @@
 #include "args.h"
 #include "mono.h"
 #include <string.h>
-#ifdef __LINUX__
-#include "lnxfix.h"
-#else
+
+#if defined(_WIN32)
 #define strcasecmp stricmp
 #endif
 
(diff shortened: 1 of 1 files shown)


Lgt2x:
Nice! tested on Linux. @JeodC could you please update the [wiki page](https://github.com/DescentDevelopers/Descent3/wiki/Running-Descent-3#command-line-options) ?

JeodC:
> Nice! tested on Linux. @JeodC could you please update the [wiki page](https://github.com/DescentDevelopers/Descent3/wiki/Running-Descent-3#command-line-options) ?

The wiki page already uses single dashes for all commands. Did you want them to show all possible combinations? That seems excessive.

Lgt2x:
> > Nice! tested on Linux. @JeodC could you please update the [wiki page](https://github.com/DescentDevelopers/Descent3/wiki/Running-Descent-3#command-line-options) ?
> 
> The wiki page already uses single dashes for all commands. Did you want them to show all possible combinations? That seems excessive.

just to mention that double dash is equivalent

PR #203, opened 2024-04-27
Fix window drag

pzychotic:
This change allows the mouse from being captured exclusively only in fullscreen mode. So in window mode, the window can actually be dragged around again.

### DIFF
--- a/Descent3/init.cpp
+++ b/Descent3/init.cpp
@@ -1671,6 +1671,14 @@ void InitIOSystems(bool editor) {
 
   Descent->set_defer_handler(D3DeferHandler);
 
+  if (!editor && !FindArg("-windowed")) {
+    if (Dedicated_server) {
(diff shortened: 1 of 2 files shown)


JeodC:
Tested and verified working (in debug mode). Thanks!

JeodC:
This introduces a new issue in debug windowed mode: the hardware cursor will minimize the window if it clicks to bring something else into focus. This means firing weapons is more difficult.

pzychotic:
> This introduces a new issue in debug windowed mode: the hardware cursor will minimize the window if it clicks to bring something else into focus. This means firing weapons is more difficult.

I see. I can restrict the mouse cursor to the window area, but than we're back to square one on the drag issue. Need to at least prevent the window from being minimized when inactive, so we can tab to another window and drag the game window as long as it is still inactive.

PR #215, opened 2024-04-28
Externalizing libacm as third party library

winterheart:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description
Reorganize libacm as external git submodule, creating directory layout for future expansion.

Updated ACM decoder to reflect recent changes.

### Checklist


JeodC (commented):
I think this can be organized better. In github, submodules look much cleaner in the root--they appear as blue folders to distinguish them from the project files. Maybe we can eliminate the `third_party` folder entirely and put the CMakeLists.txt modifications in the root version? Here is an example to illustrate:

https://github.com/HarbourMasters/Shipwright

winterheart:
It's better to keep external dependencies in separate directory and write for them separate CMakeLists.txt to avoid polluting project's main CMakeLists.txt. In that way code will be more organized.

JeodC:
> It's better to keep external dependencies in separate directory and write for them separate CMakeLists.txt to avoid polluting project's main CMakeLists.txt. In that way code will be more organized.

We can easily clean root CMakeLists.txt as well: https://github.com/HarbourMasters/Shipwright/blob/develop/CMakeLists.txt#L86

DanielGibson:
I don't see a point in introducing submodules, which make checking out the repo and building the code for users more intricate and thus error prone.
Especially not for a minimal lib that consists of two files and that should rarely require updates, now that it appears to work with the ACM files from Descent3.

JeodC:
That's why I tagged it for reviews. I think it's cleaner but you make a good point with photo files.

DanielGibson:
photo files?

Arcnor:
Agreed with the above, let's not add submodules unless necessary, and I don't think there is a need on this project.

JeodC:
> photo files?

Two. My phone loves to dyac.

bryanperris:
I thought about this too, but I see repos that do it have many active dependencies.   Here we have one third-party that is used for deciding an old game format.  It seems like an overkill for now, since the libacm won't be receiving majors updates in the next billion years.

PR #218, opened 2024-04-28
Heavy patching for compiler warnings.

icculus:
The vast majority of this is fixing up `char *` that should be `const char *` but a handful of other fixes, like potential buffer overflows that GCC noticed, etc, were applied as well.

This removes `-Wno-write-strings` from CMakeLists.txt, as it is no longer necessary, as there is no longer a flood of compiler warning spam when building.

This does not fix all compiler warnings; there are still a handful, and they are legitimate, but they can be dealt with in a future commit.

This work was originally done for the Steam release, since I was building with a compiler that treated a `char *` where it expected a `const char *` as a fatal error instead of a warning. Not sure if newer C++ standards or compilers loosened up to make this a simple warning, but it's a legit issue to be fixed in any case: if you modify a string literal, it'll crash the program on most platforms, where those are stored in read-only memory pages.

Note that patch does not make the codebase perfectly const-correct. This fixes every time the compiler threw a warning because a string literal was passed to something expecting a `char *`, and then correcting all the constness dominoes that fell as other functions then needed better const correctness. This cleaned up a surprising amount of code, but it would be a lot of work to _fully_ clean out the codebase...it would be a noble goal for someone to actually do it, but not today.

So this was originally a ~20000 line patch in my repo, and since clang-format made applying it impossible (thanks a lot), I did this work from scratch. I would strongly recommend someone evaluate this and merge it before it starts collecting conflicts, as this was many hours of tedious work and I don't want to do it a third time.  :)

This is a DRAFT pull request because I have not tested this yet! It compiles (and ...should...? work) but this computer doesn't have the game data installed and it's late, so I'll let this marinate until tomorrow. If you try this before I do, please report back!


icculus:
(Also I promise I won't write a novel like this for each Pull Request.)

icculus:
(Also I can see GitHub Actions has caught errors on Windows, etc, so I'll fix that up here too.)

JeodC:
Thank you for this submission.

#188 removed all references to `MACINTOSH` ifdefs (unused) which resulted in these conflicts. The `ddio_mac` and `mac` folders are also removed. The `renderer` directory was moved to `legacy/`, and those files in `scripts` were unused and removed.

If this causes problems, let us know and we'll revert the PR--hopefully it will make things easier in the future.

icculus:
This has now been rebased after #188 was merged, but I still need to fix up some things to make GitHub Actions happy. I'll try to sort that out today.

bryanperris:
I understand this is a big one, but could warning fixes be split based on their category?
One commit for just const char fixes, then other warning fixes in their own commits.
If the request is too insane, that is fine.

icculus:
Almost all of it is constness stuff, but there are a handful of sprintf things, etc.

I can try to split it out if it's important, but then you're going to have a few 100 line diffs and one 19,000 line diff.   :)

bryanperris:
> Almost all of it is constness stuff, but there are a handful of sprintf things, etc.
> 
> I can try to split it out if it's important, but then you're going to have a few 100 line diffs and one 19,000 line diff. :)

Yeah, don't want to make it anymore complicated than it already is, I think now, just list the important file:line in a comment.

Lgt2x:
Hi @icculus , I went through the change list, it all looks good to me. I made a few changes for the CI to compile (Windows Release still has problems), feel free to cherry-pick [b35f662](https://github.com/Lgt2x/Descent3/commit/b35f6628c690b1c97f239cfd208ee7115cb13342) if the changes seem correct, I'm not really sure about them. The game runs with these changes. Once CI is fixed, we can merge and avoid painful conflicts :)

icculus:
(Sorry, 17 thousand force-pushes going on here as I get this building on everything, disregard this for a few minutes.)

icculus:
> Hi @icculus , I went through the change list, it all looks good to me. I made a few changes for the CI to compile (Windows Release still has problems), feel free to cherry-pick [b35f662](https://github.com/Lgt2x/Descent3/commit/b35f6628c690b1c97f239cfd208ee7115cb13342) if the changes seem correct, I'm not really sure about them. The game runs with these changes. Once CI is fixed, we can merge and avoid painful conflicts :)

I ended up with most of these same changes. The `ssize_t` -> `int` change was incorrect, though: that was a 64-bit fix, as pointers get pushed through that interface, so you need the datatype to be an integer that can contain the system's address space...I didn't realize Visual Studio doesn't offer `ssize_t` though, so I changed it to `ptrdiff_t`, which seems to be widely supported.

icculus:
Okay, we are building everywhere! This is ready to merge when y'all are comfortable with it.

I can see from GitHub Actions that there are still more compiler warnings to clean out, especially in release builds, but the rest can wait for a future PR.

bryanperris:
Can you add a few of us as reviewers?

PR #227, opened 2024-04-28
Windows/CMake: add /MP for multi process MSBuild

th1000s:
### Description

"The `/MP` option can reduce the total time to compile the source files on the command line. The `/MP` option causes the compiler to create one or more copies of itself, each in a separate process."

<https://learn.microsoft.com/en-us/cpp/build/reference/mp-build-with-multiple-processes?view=msvc-170>

---

Yes, a _compile_ option which modifies the build process...

## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Runtime changes

### Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -125,6 +125,8 @@ elseif(CMAKE_SYSTEM_NAME STREQUAL "Windows")
     add_compile_options("/we4715")  # 'function' : not all control paths return a value
     add_compile_options("/we4834")  # discarding return value of function with [[nodiscard]] attribute
   endif()
+
+  add_compile_options("/MP")  # so msbuild builds with multiple processes
 endif()
 
 add_compile_definitions($<$<CONFIG:Release>:RELEASE>)


bryanperris (approved):
CI build time for windows took 58 seconds rather than 1 and a half minutes.

th1000s:
Indeed, so:

- [x] I have tested my changes ~~locally~~ **_in the cloud_** and verified that they work as intended.

Is an option to disable this parallelism needed? Not sure if `-D CMAKE_CXX_FLAGS="/MP 1"` works (no Windows here, this is copied from _$WORK_). I often use `ninja -j1` so the output is cleaner, but that doesn't require a full cmake reconfigure.

bryanperris:
> Indeed, so:
> 
> * [x]  I have tested my changes ~locally~ **_in the cloud_** and verified that they work as intended.
> 
> Is an option to disable this parallelism needed? Not sure if `-D CMAKE_CXX_FLAGS="/MP 1"` works (no Windows here, this is copied from _$WORK_). I often use `ninja -j1` so the output is cleaner, but that doesn't require a full cmake reconfigure.

Lets wait for a Windows based developer to check over this.

bryanperris:
SaladBadger has approved the change

PR #233, opened 2024-04-29
64 Bit-Safe Updates

jcoby:
Imported Icculus' 64-bit changes.

Changes `[un]signed long` data types to `[un]signed int`.


lib/pstypes.h
```diff
@@ -26,7 +26,7 @@ typedef unsigned char ubyte;
 typedef signed char sbyte;
 typedef unsigned short ushort;
 typedef unsigned int uint;
-typedef unsigned long ulong;
+typedef unsigned int ulong;
```
winterheart, line 29: This cause compilation failures. Need to revert back to long.

JeodC: I'm not sure that's the problem. Converting `unsigned long` to `unsigned int` means LPDWORDs need static casts.

JeodC (approved):
Runs fine on windows

PR #235, opened 2024-04-29
CMake: expose BUILD_TESTNG option, rename LOGGER to ENABLE_LOGGER

Lgt2x:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

Expose BUILD_TESTING as a CMake option, to show up in CMake-gui and ccmake. Also rename LOGGER to ENABLE_LOGGER (options need verbs)


bryanperris (approved):

th1000s:
`if(LOGGER)`,  maybe the compile definition name, and the README need to be adapted as well.

Lgt2x:
> `if(LOGGER)`, maybe the compile definition name, and the README need to be adapted as well.

yes yes I pushed too promptly :)

winterheart:
We may just change CMakeLists.txt alone and left #ifdefs in code as LOGGER as names of defines is hidden from users.

Lgt2x:
> We may just change CMakeLists.txt alone and left #ifdefs in code as LOGGER as names of defines is hidden from users.

as you wish, I reverted the define change

winterheart:
It's ok for merge after resolving conflicts.

PR #241, opened 2024-04-30
Initial SDL2 port

icculus:
This just gets the basic SDL2 work in place. The resolution changes, etc, are coming in a separate PR.

These are split across several commits because that's how it was in my fork, but I could squash this down to a single unit of work and force-push this PR if that's preferable.

Note that SDL2 has no CD-ROM subsystem, so this removes physical disc support from ddio_lnx/lnxcdrom.cpp. If that's a serious problem in 2024, we can talk about what to do about it; likely we can get away with something quick and dirty if this is even necessary anymore.

This has gone as far as "it compiles and works on my machine," so please test it! I'm marking this as a draft PR until GitHub Actions gives the thumbs up.


GravisZro:
I would prefer this not be merged as my particular target platform only has SDL 1.2 and not SDL2. A preferred solution would be to make it optional.

icculus:
> particular target platform

What platform is this? Dreamcast?

GravisZro:
> What platform is this? Dreamcast?

Yes... but there are also other old platforms in the same boat.

winterheart:
Our major platforms are Linux, macOS and Windows (actual platforms for most of users). We cannot support both SDL 1.2 and SDL 2 at same time. SDL 1.2 nearly reached EOL support and lacks of features like hardware acceleration support and advanced rendering support, that we count on in future development.

JeodC:
It's currently in the pipeline as we work on getting windows up to parity with Ryan's 2020 Linux and mac ports. If you want to keep it, you should create a fork of this project and branch it.

jcoby:
This causes keyboard controls to act weird. pitch and yaw start out fast but quickly ramp down to a crawl. roll is limited to about 45°. might be related to removal of `SDL_EnableKeyRepeat`?

renderer/HardwareOpenGL.cpp
```diff
@@ -595,15 +597,18 @@ int opengl_Setup(oeApplication *app, int *width, int *height) {
     rend_SetErrorMessage(buffer);
     return (0);
   } // if
-  SDL_SetEventFilter(d3SDLEventFilter);
+  SDL_SetEventFilter(d3SDLEventFilter, NULL);
 
   bool fullscreen = false;
+
+#if 0  // this was a workaround for 3DFx Voodoo cards back in the day. Presumably no longer needed?    --ryan 04/27/2024
```
Arcnor, line 604: I'd say let's kill it

icculus:
> This causes keyboard controls to act weird.

Whoops, there's a later commit that fixed this I haven't moved over yet. I'll verify today and get that pushed to this PR.

renderer/HardwareOpenGL.cpp
```diff
@@ -690,45 +690,46 @@ int opengl_Setup(oeApplication *app, int *width, int *height) {
...
-    linux_permit_gamma = (gammarc == 0);
+    Uint16 ramp[256];
+    SDL_CalculateGammaRamp(Render_preferred_state.gamma, ramp);
+    linux_permit_gamma = (SDL_SetWindowGammaRamp(GSDLWindow, ramp, ramp, ramp) == 0);
   } // else
 
   if (ParentApplication) {
     reinterpret_cast<oeLnxApplication *>(ParentApplication)->set_sizepos(0, 0, *width, *height);
-    SDL_FillRect(SDL_GetVideoSurface(), NULL, SDL_MapRGB(SDL_GetVideoSurface()->format, 255, 0, 0));
+    //SDL_FillRect(SDL_GetVideoSurface(), NULL, SDL_MapRGB(SDL_GetVideoSurface()->format, 255, 0, 0));
```
Jayman2000, line 732: Why did you choose to comment this line out instead of deleting it?

Descent3/lnxmain.cpp
```diff
@@ -322,6 +322,11 @@ int d3SDLEventFilter(const SDL_Event *event) {
     return (sdlMouseButtonUpFilter(event));
   case SDL_MOUSEBUTTONDOWN:
     return (sdlMouseButtonDownFilter(event));
+  case SDL_QUIT:
+    SDL_Quit();
+    _exit(0);
+    break;
+  default: break;
```
Jayman2000, line 329: I would not include a default case instead of including a default case that doesn’t do anything.

icculus:
> This causes keyboard controls to act weird. pitch and yaw start out fast but quickly ramp down to a crawl. roll is limited to about 45°. might be related to removal of SDL_EnableKeyRepeat?

This is fixed in 851f47537b8a40477543eeaa274c1692895e78f9 (and yes, it was a key-repeat thing, good call!).

icculus:
> When I run the version of the game that’s in this PR, cutscene audio still doesn’t work, but it’s much worse.

This is fixed in 65b1a7dc3bb921a40205a02b944208ef5179babe.  (This was because I had abandoned the MVE code in my fork, so this never got fixed before.)

Descent3/lnxmain.cpp
icculus: It _does_ do something: it prevent compiler warnings since we have a `switch` on an enum when we (intentionally) don't handle every value in that enum.

renderer/HardwareOpenGL.cpp
icculus: My thinking was "in case this turned out to be important, we can revisit it later," but honestly it's probably not important, so I've removed the line in b362bc9c543f9688f6f269167f83c6606ae679aa.

renderer/HardwareOpenGL.cpp
icculus: It's gone, as of 6f07fd5a94190f67813d637ffbce77d8507753a7 !

icculus:
> Dependencies in README.md need to be updated. They still say to install SDL 1.2.

Resolved in bd3a596975aeadfb3288102530823b0287796333, thanks!

JeodC:
> > I can’t bind controls to the scroll wheel.
> 
> 
> 
> Fixed in 80335136e3fa46aa6d2860915dc6a48dd0ffadcc ...this also adds support for mouse buttons beyond left/middle/right, but this is untested, as I don't have a mouse that supports it.

I do, I'll give this a try today and let you know.

icculus:
> I do, I'll give this a try today and let you know....assuming I can try on windows.

Eventually you can, but right now the SDL2 work only builds on Linux and macOS; Windows still talks directly to the Win32 API and DirectX instead.

icculus:
Okay, I think I've resolved all the concerns for this PR; there will be more SDL2 work in the future, including moving it to Windows, and other patches on top of this, like all the video resolution improvements, but this seems like a reasonable baseline, if people are happy with it.

Jayman2000 (approved):

winterheart (approved):
Only thing that I've noticed - app runs in windowed mode, but that ok.
Tested on Linux 64bit.

icculus:
Oh yeah, I changed it back to default to fullscreen right before shipping. It's a one-line change, I'll switch it shortly.

icculus:
Fullscreen mode defaults to on as of 6d837e20830dfd2e297e5ddebfcf6403930a926d.

renderer/lnxscreenmode.cpp
```diff
@@ -115,6 +115,7 @@ CLnxVideoModes::~CLnxVideoModes() {
 // enumerate all the possible video modes
 bool CLnxVideoModes::Init(void) // Display *dpy,int screen)
 {
+#if 0  // with SDL2, we'll just use FULLSCREEN_DESKTOP and never change the physical vidmode
```
Lgt2x, line 118: to be deleted sometime

Lgt2x (approved):
LGTM, tested on Linux x64. We can merge as it is and solve further problems from here

PR #243, opened 2024-04-30
Remove czip/unzip references

GravisZro:
Removes references to the czip/unzip libraries which are to be replaced with zlib.

## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [x] Network changes
  - [ ] Other changes

### Checklist


Descent3/mission_download.cpp
```diff
@@ -694,7 +693,7 @@ void _get_zipfilename(char *output, char *directory, char *zipfilename) {
 // return 0 on failure
 // return 1 on success
 int msn_ExtractZipFile(char *zipfilename, char *mn3name) {
-
+/*
```
Lgt2x, line 696: so you just disable the feature ?

GravisZro: Most of "mission download" was already commented out due to lack of a location to download from. The only thing I did was comment out the unzipping after it was downloaded. This seems to be the only place that the deflate decompression algorithm is used.

JeodC: Mission files hosted by a server contain a url pointing clients to the download location. If zlib is meant to replace, it should be integrated here.

GravisZro: Presently, downloading is entirely disabled. Replacing the compression code right now is a moot point and cannot be tested.

Lgt2x: This change will make our lives harder when/if we want to enable mission download again, I don't think we should merge it as is

PR #246, opened 2024-04-30
Remove libGLU references.

icculus:
The game doesn't actually use GLU (just OpenGL), so this can go.

### DIFF
--- a/Descent3/CMakeLists.txt
+++ b/Descent3/CMakeLists.txt
@@ -275,14 +275,14 @@ set(CPPS
   weather.cpp)
 
 if(WIN32)
-  set(PLATFORM_LIBS dd_sndlib dd_grwin32 dd_vidwin32 ddio_win win32 wsock32.lib winmm.lib Glu32.lib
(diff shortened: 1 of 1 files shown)


Lgt2x:
`D3Launch.cpp` still mentions `Glu32.dll` in some functions and `Soft/HardwareOpenGL.cpp` still import `Glu.h`, should we remove those as well ?

icculus:
I didn't touch `legacy` because I wasn't sure what the rules are with that, but I'm happy to make changes!

GLU.h in HardwareOpenGL.cpp (and all of SoftwareOpenGL.cpp) went away when #244 was merged.

Lgt2x:
> I didn't touch `legacy` because I wasn't sure what the rules are with that, but I'm happy to make changes!
> 
> GLU.h in HardwareOpenGL.cpp (and all of SoftwareOpenGL.cpp) went away when #244 was merged.

Indeed, the branch I was looking at was not rebase. all good !

PR #254, opened 2024-05-01
Correctly load mercenary assets (extra.hog)

### LINKED ISSUE #255: [Runtime Issue]: Mercenary game assets confusion
### Build Version

current

- [X] Microsoft Windows (32-bit)
- [X] Microsoft Windows (64-bit)
- [X] Mac OS X

### Description

The game will check for the existence of Mercenary expansion via registry string. If it doesn't exist, the Black Pyro ship is supposedly unable to be selected at Pilot->Configure (after passing Merc Level 2). #254 attempts to fix that, but Merc assets still failed to load from the steam version of the game where they are located in `extra.hog`.

Only after installing the CD-ROM, installing the Merc CD, then patching to 1.4, can the game function as expected by copying `merc.hog` and `EXTRA.HOG` (renamed to `extra.hog`) to the other install directory.

In essence #254 is only a partial fix. The underlying cause of this bug is unknown.

File breakdown:

`EXTRA.HOG` contains the following:
image

`merc.hog` contains a lot more:
image

In comparing the two, it appears the steam release `extra.hog` is a copy of `merc.hog`, and `EXTRA.HOG` from the CD is missing entirely.

JeodC:
## Pull Request Type

- [ ] GitHub Workflow changes
- [x] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

Some mercenary assets are not loaded properly since merc.hog is named extra.hog in digital releases. Adjust the check here. Resolves #255 and enables mercenary stuff for other platforms.

### DIFF
--- a/Descent3/init.cpp
+++ b/Descent3/init.cpp
@@ -1693,21 +1693,10 @@ void InitIOSystems(bool editor) {
   ddio_MakePath(fullname, LocalD3Dir, "extra.hog", NULL);
   extra_hid = cf_OpenLibrary(fullname);
 
-  // Opens the mercenary hog if it exists and the registry entery is present
-  // Win32 only.  Mac and Linux users are SOL for now
-#ifdef WIN32
-  HKEY key;
-  DWORD lType;
-  LONG error;
-
-#define BUFLEN 2000
-  char dir[BUFLEN];
-  DWORD dir_len = BUFLEN;
-
   // Always look for merc.hog first
   merc_hid = cf_OpenLibrary("merc.hog");
   // Check if merc_hid is valid, if not, open extra.hog
-  if (merc_hid == INVALID_HANDLE_VALUE) {
+  if (merc_hid == nullptr) {
     merc_hid = cf_OpenLibrary("extra.hog");
   }
   // Open this for extra 1.3 code (Black Pyro, etc)
--- a/Descent3/init.cpp
+++ b/Descent3/init.cpp
@@ -1696,7 +1696,7 @@ void InitIOSystems(bool editor) {
   // Always look for merc.hog first
   merc_hid = cf_OpenLibrary("merc.hog");
   // Check if merc_hid is valid, if not, open extra.hog
-  if (merc_hid == nullptr) {
+  if (merc_hid == -1) {
     merc_hid = cf_OpenLibrary("extra.hog");
   }
   // Open this for extra 1.3 code (Black Pyro, etc)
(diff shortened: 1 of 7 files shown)


Descent3/init.cpp
```diff
@@ -1690,38 +1690,20 @@ void InitIOSystems(bool editor) {
...
+  }
 
-    if (error == ERROR_SUCCESS) {
-      merc_hid = cf_OpenLibrary("merc.hog");
-    }
+  // Always look for extra.hog first
+  merc_hid = cf_OpenLibrary("extra.hog");
+  // Check if merc_hid is valid, if not, open merc.hog
+  if (merc_hid == -1) {
+    merc_hid = cf_OpenLibrary("merc.hog");
```
Jayman2000, line 1705: I want to make sure that I’m understanding this code correctly. Would this make it so that `merc.hog` only gets loaded if `extra.hog` fails to load?

JeodC: Correct. If you have `extra1.hog`, then the mercenary assets are in `extra.hog`. If you have `extra.hog`, mercenary assets are in `merc.hog`.

Jayman2000: I installed the Windows version Descent 3+Mercenary from CDs, and I have both `extra.hog` and `merc.hog`. In that scenario, I think that this code would make it so that `merc.hog` doesn’t get loaded. This code would load `extra.hog` successfully and then ignore `merc.hog`.

JeodC:
Verified working with both combinations on windows.

JeodC:
Tested and verified working in windows. Red guidebot is back!

JeodC:
This will make an access violation crash if the player attempts to use the black pyro in multiplayer without mercenary hog assets present (missing cockpit graphic for the ship).

Descent3/init.cpp
```diff
@@ -1548,40 +1548,6 @@ tTempFileInfo temp_file_wildcards[] = {{"d3s*.tmp"}, {"d3m*.tmp"}, {"d3o*.tmp"},
                                        {"d3c*.tmp"}, {"d3t*.tmp"}, {"d3i*.tmp"}};
 int num_temp_file_wildcards = sizeof(temp_file_wildcards) / sizeof(tTempFileInfo);
 
-// Returns true if Mercenary is installed
-bool MercInstalled() {
```
jcoby, line 1552: instead of blowing this away, how about looking for `merc_hid > 0`? that'll support CD installs without the merc expansion.

JeodC: That's way more elegant. I think I'll do this all locally when I have time to keep from so many commits in the history.

PR #272, opened 2024-05-02
x86-64 supports the `int $3` opcode to trigger a breakpoint.

icculus:
This architecture obviously didn't exist in 1999.  :)


JeodC:
This may not be needed for libmve since it's changing, but merging anyway for reference.

PR #304, opened 2024-05-07
C++ Algorithm pokes

MaddTheSane:
## Pull Request Type

- [ ] GitHub Workflow changes
- [x] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [x] Other changes

### Description
This changes some calls to `std::max` and `std::min` to use the `initializer_list` version, which allows multiple elements to be passed in at once.
This also changes some calls to `std::max` and `std::min` to `std::clamp`.

### Checklist

### Additional Comments
Looks like `std::clamp` was added in C++17.

### DIFF
--- a/renderer/HardwareOpenGL.cpp
+++ b/renderer/HardwareOpenGL.cpp
@@ -2854,7 +2854,7 @@ void rend_DrawSpecialLine(g3Point *p0, g3Point *p1) {
     }
 
     // Finally, specify a vertex
-    float z = std::max<double>(0, std::min(1.0, 1.0 - (1.0 / (pnt->p3_z + Z_bias))));
+    float z = std::clamp<double>(1.0 - (1.0 / (pnt->p3_z + Z_bias)), 0, 1);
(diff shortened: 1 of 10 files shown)


renderer/HardwareOpenGL.cpp
```diff
@@ -2854,7 +2854,7 @@ void rend_DrawSpecialLine(g3Point *p0, g3Point *p1) {
     }
 
     // Finally, specify a vertex
-    float z = std::max<double>(0, std::min(1.0, 1.0 - (1.0 / (pnt->p3_z + Z_bias))));
+    float z = std::clamp<double>(1.0 - (1.0 / (pnt->p3_z + Z_bias)), 0, 1);
```
Lgt2x, line 2857: remove `<double>` specialization and clamp between `0.0` and `1.0` instead (same elsewhere)

MaddTheSane: I just migrated the code to use `std::clamp`.

Still, will do.

Lgt2x: Thanks

Issue #295, opened 2024-05-06
CI artifacts lack licenses for libraries


Jayman2000:
Descent 3 links to several open-source libraries. At the moment, Descent 3’s CI artifacts don’t comply with those libraries’s licenses. Specifically, they require that you include a copy of the license when distributing the libraries. At the moment, CI artifacts are missing licenses for the following:

|Name | Inclusion point | License |
|-----|-----------------|---------|
| libacm | [`AudioEncode/CMakeLists.txt:6`](https://github.com/DescentDevelopers/Descent3/blob/0e69bf465fc6f7bdbf70f46d3a4d4cf1593bc2f6/AudioEncode/CMakeLists.txt#L6) | [Link](https://github.com/DescentDevelopers/Descent3/blob/0e69bf465fc6f7bdbf70f46d3a4d4cf1593bc2f6/libacm/libacm.h#L4) |
| spdlog | [`misc/CMakeLists.txt:12`](https://github.com/DescentDevelopers/Descent3/blob/0e69bf465fc6f7bdbf70f46d3a4d4cf1593bc2f6/misc/CMakeLists.txt#L12) | [Link](https://github.com/gabime/spdlog/blob/27cb4c76708608465c413f6d0e6b8d99a4d84302/LICENSE)
| RSA Data Security, Inc. MD5 Message-Digest Algorithm | [`Descent3/CMakeLists.txt:304`](https://github.com/DescentDevelopers/Descent3/blob/0e69bf465fc6f7bdbf70f46d3a4d4cf1593bc2f6/Descent3/CMakeLists.txt#L304) | [Link](https://github.com/DescentDevelopers/Descent3/blob/0e69bf465fc6f7bdbf70f46d3a4d4cf1593bc2f6/md5/md5.h#L29)

JeodC:
This should be considered "in progress".

- libacm can be included, we have no plans to switch the audio library at the moment
- spdlog may be partially reverted (see #271) which may change its license terms
- RSA MD5 should be changed entirely, see https://github.com/twitter/twemproxy/issues/120

JeodC:
> This should be considered "in progress".
> 
> * libacm can be included, we have no plans to switch the audio library at the moment
> * spdlog may be partially reverted (see [Revert "Introduce spdlog logging system" #271](https://github.com/DescentDevelopers/Descent3/pull/271)) which may change its license terms
[...]

PR #313, opened 2024-05-08
Use OpenSSL MD5 instead of RFC 1321 implementation.

Nakhr11n:
## Pull Request Type

- [x] GitHub Workflow changes
- [x] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [x] Other changes

### Description

Use OpenSSL for MD5 digest instead of the reference RFC 1321 code.
Code adapted from: [https://github.com/openssl/openssl/blob/master/demos/digest/EVP_MD_demo.c](https://github.com/openssl/openssl/blob/master/demos/digest/EVP_MD_demo.c)
Works with OpenSSL 1.\*.\* and 3.\*.\*
It should work with any other implementation of OpenSSL as well, such as LibreSSL, WolfSSL, etc.

### Related Issues

#295

### Checklist

### Additional Comments

OpenSSL is [licensed under Apache 2.0](https://www.openssl.org/source/license.html).
I have no experience with open-source licensing, but since we are not redistributing any OpenSSL source code, it seems like we do not have to include the license.

This is a single-file header because when I tried splitting them, for some reason the rest of the project stopped seeing the MD5Update symbols, and it couldn't link it all together.
Maybe someone knows a solution? But I think it's fine as it is.


Arcnor:
This adds more external dependencies (which complicates porting to platforms that don't use OpenSSL for example) for something that already works, I don't think we should do this.

Lgt2x:
Sorry but adding openssl as a dependency just for an alternative md5 implementation is not what we want.

JeodC:
We should change MD5 anyway to [avoid the current advertising clause]

> This adds more external dependencies (which complicates porting to platforms that don't use OpenSSL for example) for something that already works, I don't think we should do this.

It works but we are supposed to advertise RSA MD5 whatever. It's something to get away from, which motivated this PR, just adding another dependency for it is the wrong direction.

PR #315, opened 2024-05-08
Renderer refactor

Arcnor:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description
This is the first of a number of PRs that refactor the renderer to make it easier to port to different APIs. This first chunk of work simply moves all that's not OpenGL specific to a base file.

If we want to allow multiple renderers & dynamic change between them this will need a bigger change to actually use classes (or any other mechanism) for easier overriding of behavior. That said, even without that a proper way of overriding will be needed for renderers that actually need to change the behavior of those base functions I just moved.

Each commit on this PR should compile on its own and should present no changes when ran.

### Checklist


renderer/HardwareBaseGPU.cpp
```diff
@@ -0,0 +1,436 @@
+/*
+* Descent 3
+* Copyright (C) 2024 Parallax Software
```
winterheart, line 3: As this is new genuine code, there need add our own copyright header:

```
* Copyright (C) 2024 Descent Developers
```

Arcnor: I will change it, although it's not really new, but a mix...

renderer/HardwareOpenGL.cpp
```diff
@@ -166,10 +149,8 @@ ushort *opengl_packed_Upload_data = NULL;
 ushort *opengl_packed_Translate_table = NULL;
 ushort *opengl_packed_4444_translate_table = NULL;
 
-rendering_state OpenGL_state;
-static float Alpha_multiplier = 1.0f;
-
-renderer_preferred_state OpenGL_preferred_state = {0, 1, 1.5};
+extern rendering_state gpu_state;
```
Lgt2x, line 152: sharing state using `extern` variables everywhere makes it harder to understand. I hope we can use real dependency injection at some point.

Arcnor: Agreed, I'm just following the existing pattern everywhere, and just want it to work before we do big architectural changes. As mentioned above, this will need to be better organized for other renderers to work at the same time, or even at all, but let's start slow :).

renderer/HardwareInternal.h
```diff
@@ -58,4 +58,7 @@ void rend_TransformSetViewport(int lx, int ty, int width, int height);
 void rend_TransformSetProjection(float trans[4][4]);
 void rend_TransformSetModelView(float trans[4][4]);
 
+int rend_ReInit();
+float rend_GetAlphaMultiplier();
```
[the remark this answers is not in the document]
Arcnor: But those are not implemented on HardwareBaseGPU, but every renderer, that's why I added it to Internal for now.

Lgt2x (approved):
There is much more work to do for the renderer to be better organized but that's a good first step! Don't hesitate to create a tracking issue to share your full vision and progress ;)

PR #318, opened 2024-05-08
Introducing installation steps

winterheart:
## Pull Request Type

- [x] GitHub Workflow changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

Added installation steps for all built targets. Added FORCE_PORTABLE_INSTALL cmake option that controls portable installation (only supported for now).

### Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,6 +11,7 @@ project(Descent3
 )
 
 option(FORCE_COLORED_OUTPUT "Always produce ANSI-colored compiler warnings/errors (GCC/Clang only; esp. useful with ninja)." OFF)
+option(FORCE_PORTABLE_INSTALL "Install all files into local directory defined by CMAKE_INSTALL_PREFIX" ON)
 option(ENABLE_LOGGER "Enable logging to the terminal" OFF)
 option(BUILD_TESTING "Enable testing. Requires GTest." OFF)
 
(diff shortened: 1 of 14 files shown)


JeodC (commented):
Missing dependencies like fmt.dll and should also pack dmfc if it exists (separate PR open to build that).

CMakeLists.txt
```diff
@@ -11,6 +11,7 @@ project(Descent3
 )
 
 option(FORCE_COLORED_OUTPUT "Always produce ANSI-colored compiler warnings/errors (GCC/Clang only; esp. useful with ninja)." OFF)
+option(FORCE_PORTABLE_INSTALL "Install all files into local directory defined by CMAKE_INSTALL_PREFIX" OFF)
```
Lgt2x, line 14: This install is only meant to be portable, should we set CMAKE_INSTALL_PREFIX to a path relative to the build directory by default when not set by the user ([like there](https://stackoverflow.com/a/16076855))? The default CMAKE_INSTALL_PREFIX is a system path and don't want game files to end up there

CMakeLists.txt
```diff
@@ -33,6 +34,14 @@ if(FORCE_COLORED_OUTPUT)
   endif()
 endif()
 
+include(GNUInstallDirs)
```
Lgt2x, line 37: How does Windows behave here?

CMakeLists.txt
```diff
@@ -33,6 +34,14 @@ if(FORCE_COLORED_OUTPUT)
...
+if(FORCE_PORTABLE_INSTALL)
+  set(CMAKE_INSTALL_BINDIR ".")
+  set(CMAKE_INSTALL_LIBDIR ".")
```
Lgt2x, line 40: If I understand correctly we use these variables to prepare for a future possible system installation of the game? This looks like an anti-pattern to me to include them just to override them after.

JeodC: We should never "install" the game in the traditional sense imho.

Lgt2x: if we ever want to package for Linux distributions it can, but I don't see this being implemented in the near future.

CMakeLists.txt
winterheart: Noop on Windows. This include for future expansion and re-using already defined CMAKE_INSTALL_* variables.

CMakeLists.txt
winterheart: @Lgt2x is correct about my intentions about future installation game on Linux systems. These variables redefined only for "portable variant", which finds many applications such correct way to build and install all required targets to assemble portable game installation instead using custom commands in CMakeLists.txt.

PR #327, opened 2024-05-09
spdlog: move to third-party directory

Lgt2x:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

spdlog has a header-only version that is small enough to be in-tree. This way, we remove the requirement of installing a third party package for logging.
This PR also makes logging enabled all the time, thus removing `ENABLE_LOGGING` CMake option. 


JeodC:
Does this include the reversion in #271?

Lgt2x:
> Does this include the reversion from full spdlog in #271?

Well... No, this PR keeps spdlog

JeodC:
> > Does this include the reversion from full spdlog in #271?
> 
> Well... No, this PR keeps spdlog

Yes, but it's the header, not the complete library. What becomes of #271 then?

Lgt2x:
> > > Does this include the reversion from full spdlog in #271?
> > 
> > 
> > Well... No, this PR keeps spdlog
> 
> Yes, but it's the header, not the complete library. What becomes of #271 then?

The header here contains the whole library, that's the point of header only libraries. #271 should be closed if this change is accepted

Jayman2000:
Could you add an option to use your system’s version of spdlog instead of the one in this repo? I’d rather use the version of spdlog that my distro ships.

Lgt2x:
> Could you add an option to use your system’s version of spdlog instead of the one in this repo? I’d rather use the version of spdlog that my distro ships.

I'd prefer not to. The main reason of this change is to make sure every build gets the same exact version of the library, without depending on any package manager. We don't offer any guarantee that your distro's version will always be compatible with the one in the repo in case of breaking change or regression. Is there a specific why you would want that?

Arcnor:
That's all great, but we're making a game here. When we ship with a library, we expect users to use what we dictate, same as with all other games you can find anywhere.

If you want to change the library you can definitely do so, by modifying the source, it's there for this after all, but we won't go out of our way to support that use case.

Jayman2000:
> That's all great, but we're making a game here. When we ship with a library, we expect users to use what we dictate, same as with all other games you can find anywhere.

I don’t really know about all other games you can find anywhere, but if that is true, then that’s quite a shame. It would be better if all games allowed you to do this.

> If you want to change the library you can definitely do so, by modifying the source, it's there for this after all, but we won't go out of our way to support that use case.

Point of clarification: if we were to “go out of our way to support that use case”, what additional work would have to be done?

winterheart:
I think, if there so much back-and-forth discussions about logging subsystem, maybe we just step back and make another attempt to choose alternative logger? Ryan already proposed SDL2's `SDL_Log()` logger facility. Since we going to switch SDL2 as main rendering back-end, this proposal looks better than using any external dependencies.

JeodC (approved):

Lgt2x:
Replaced by #341

PR #333, opened 2024-05-10
Minor improvements to `cf_OpenLibrary()`’s HOG2 id detection code

Jayman2000:
## Pull Request Type

- [ ] GitHub Workflow changes
- [x] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [x] Other changes

### Description

This PR contains two minor improvements to the code that detects whether or not `.hog` files actually begin with the ASCII characters HOG2. See the commit messages for details.

### Checklist

### DIFF
--- a/cfile/hogfile.h
+++ b/cfile/hogfile.h
@@ -65,6 +65,7 @@
 
 #define HOG_HDR_SIZE (64)
 #define HOG_TAG_STR "HOG2"
+#define HOG_TAG_LEN (4)
 #define HOG_FILENAME_LEN (36)
 
 typedef struct tHogHeader {
--- a/cfile/cfile.cpp
+++ b/cfile/cfile.cpp
@@ -147,8 +147,7 @@ int cf_OpenLibrary(const char *libname) {
     mem_free(lib);
     return 0; // CF_NO_FILE;
   }
-  fread(id, HOG_TAG_LEN, 1, fp);
-  if (strncmp(id, HOG_TAG_STR, HOG_TAG_LEN)) {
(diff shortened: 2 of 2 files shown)


GravisZro:
It would be better to simply replace `#define HOG_TAG_STR "HOG2"` with a std::string_view. However, my understanding is that  @winterheart plans on replacing `cfile.*` and `hogfile.*` entirely which would negate any work on them.

cfile/hogfile.h
```diff
@@ -62,9 +62,11 @@
 
 #include <cstdint>
 #include <cstdio>
+#include <string>
 
 #define HOG_HDR_SIZE (64)
 #define HOG_TAG_STR "HOG2"
+#define HOG_TAG_LEN std::string::traits_type::length(HOG_TAG_STR)
```
winterheart, line 69: No need to call std::string::length, we already knows that there always be 4 bytes. Just define it as is:

```cpp
#define HOG_TAG_LEN (4)
```

winterheart (commented):
I'd prefer move header check into `ReadHogHeader()` function of hogfile.cpp (with some refactoring and additional sane checks). There a lot of low-level checks in `cf_OpenLibrary()` already.

cfile/hogfile.h
Jayman2000: OK, done.

cfile/hogfile.h
```diff
@@ -62,9 +62,11 @@
 
 #include <cstdint>
 #include <cstdio>
+#include <string>
```
winterheart, line 65: This include unneeded.

Jayman2000: Good catch. I just force pushed a new version that doesn’t have that line.

winterheart:
OK, ready to merge, thank you!

PR #334, opened 2024-05-11
Cleanup init.cpp

JeodC:
## Pull Request Type

- [ ] GitHub Workflow changes
- [x] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

Cleans `init.cpp`. Removes a lot of cruft and adds some extra messages for debugging.

### DIFF
--- a/Descent3/init.cpp
+++ b/Descent3/init.cpp
@@ -1870,46 +1870,37 @@ void InitD3Systems1(bool editor) {
   InitGraphics(editor);
 
   //	initialize data structures
-  INIT_MESSAGE(("Initializing data structures."));
   InitObjectInfo();
   InitVClips();
   InitRooms();
 
   //	initialize lighting systems
-  INIT_MESSAGE(("Initializing lighting systems."));
   InitLightmapInfo();
   InitSpecialFaces();
   InitDynamicLighting();
 
   // Initialize missions
-  INIT_MESSAGE(("Initializing missions."));
   InitMission();
 
   // Initializes the ship structure
-  INIT_MESSAGE(("Initializing ships."));
   InitShips();
 
   // Initializes the fvi structures for quicker operation
-  INIT_MESSAGE(("Initializing FVI."));
   InitFVI();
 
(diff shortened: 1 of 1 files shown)


Descent3/init.cpp
```diff
@@ -1343,26 +1241,26 @@ void LoadGameSettings() {
     strcpy(Game_gauge_usefile, GameArgs[tt_arg + 1]);
   }
 
-  Detail_settings.Specular_lighting = false;
+  Detail_settings.Specular_lighting = true;
```
Lgt2x, line 1244: This is not cleanup. This is changing defaults. Same for bitdepth and render distance. It needs a separate PR motivating the changes.

JeodC: For bitdepth, 16bit is no longer supported by anything isn't it? I can make it separate but bumping the defaults to new minimums seems reasonable.

Lgt2x: I have not studied the implications of the bit depth change in this file. It will require more testing and understanding of the code, and is outside of the scope of a code cleanup MR anyway.

JeodC: Fair enough.

Descent3/init.cpp
```diff
@@ -1993,82 +1866,51 @@ void InitD3Systems1(bool editor) {
...
-#endif
-      exit(0);
-    }
-  }
-#endif
   INIT_MESSAGE(("Initializing GFX"));
   InitGraphics(editor);
 
   //	initialize data structures
+  INIT_MESSAGE(("Initializing data structures."));
```
Lgt2x, line 1873: Every init message in this function after `InitGraphics` crashes the game on Linux, probably because the graphics are half-initialized at this point

JeodC: Does the existing message "Initializing I/O systems" also crash it?

Lgt2x (approved):

PR #338, opened 2024-05-12
Bump MAX_CELLS_TO_RENDER

### LINKED ISSUE #280: [Runtime Issue]: Render depth resets to lowest on map edge
### Build Version

0e55dee0dec22b2895f14beb0c789bbbd489d5c3

- [ ] Mac OS X

### CPU Environment

- [ ] x86 (32-bit Intel/AMD)
- [ ] ARM (32-bit)
- [ ] Other (RISC V, PPC...)

### Game Modes Affected

- [X] Single player
- [ ] Anarchy
- [ ] Hyper-Anarchy
- [ ] Robo-Anarchy
- [ ] Team Anarchy
- [ ] Capture the Flag
- [ ] Bounty
- [ ] Entropy
- [ ] Hoard
- [ ] Monsterball
- [ ] Cooperative

### Game Environment

Descent 3: Retribution, Map 2.
When you exit out to the surface and get to the large open area this becomes most noticeable. 

### Description

The Render Depth for me is being reset from the highest (100) to the lowest (40) setting when touching the map edge between the ground and space. It doesn't seem to be consistent when happening, but flying around and wedging myself on the map edge was the best way to make this happen. 

### Regression Status

I don't know.

### Steps to Reproduce

1. Set your Render Depth to 100
2. Load map 2 for the base campaign (Descent 3: Retribution)
3. Play the map till you go out into the wide open crater.
4. Wedge yourself into the edge of the map (between ground and space)
5. If you are looking at a spot far enough away it will disappear when it happens.
6. Looking at your Render Depth will show that it has been set to 40

JeodC:
## Pull Request Type

- [ ] GitHub Workflow changes
- [x] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

Bump max cells to 32678 from 8000, allowing a much larger render distance for outdoor terrain. This commit also fixes MAX_HORIZON_PIECES, which may have been broken during a clang-format.

Testing on Level 7 yielded no performance loss, and in fact revealed an existing bug with z-ranges where blue electricity effects from disabled turrets are visible overtop of terrain.

image

### Related Issues

Fixes #280

### DIFF
--- a/Descent3/terrain.h
+++ b/Descent3/terrain.h
@@ -41,7 +41,7 @@
 
 #define MAX_LOD_SIZE (1 << (MAX_TERRAIN_LOD - 1))
 
-#define MAX_CELLS_TO_RENDER 8000 // The maximum number of cells we can render in a frame
+#define MAX_CELLS_TO_RENDER 32768 // The maximum number of cells we can render in a frame
 
 #define MAX_TERRAIN_HEIGHT 350.0f                              // Max height of our terrain
 #define TERRAIN_HEIGHT_INCREMENT (MAX_TERRAIN_HEIGHT / 255.0f) // Incremental jumps
@@ -50,9 +50,7 @@
 // Sky defines
 #define MAX_STARS 600    // how many stars in our sky
 #define MAX_SATELLITES 5 // max satellites in our sky
-#define MAX_HORIZON_PIECES                                                                                             \
-  16 // how many segments of the horizon
-     // there are around our sphere
+#define MAX_HORIZON_PIECES 16 // how many segments of the horizon there are around our sphere
 
 // Sky flags
 #define TF_STARS 1      // whether or not our terrain is starred


Lgt2x:
It's unclear to me without diving in the rendering code why decreasing the maximum number of horizon pieces increases the render distance. What happens there?

JeodC:
> It's unclear to me without diving in the rendering code why decreasing the maximum number of horizon pieces increases the render distance. What happens there?

The horizon distance change has nothing to do with the render distance. In the current `terrain.h` file though it looks like this:
```cpp
// Sky defines
#define MAX_STARS 600    // how many stars in our sky
#define MAX_SATELLITES 5 // max satellites in our sky
#define MAX_HORIZON_PIECES                                                                                             \
  16 // how many segments of the horizon
     // there are around our sphere
     ```
The change just brings it inline like the rest. I'm not sure if it was broken or not, but it was inconsistent.

**Comment by JeodC** [comment]
> I was talking about the `MAX_CELLS_TO_RENDER` change, sorry. The other change is fine.

It's an increase, not a decrease. Original value was 8000, and in `TerrainSearch.cpp`:
```cpp
  if (*ccount >= MAX_CELLS_TO_RENDER) {
    mprintf((0, "Trying to render too many cells!  Cell limit=%d\n", MAX_CELLS_TO_RENDER));
#ifndef NEWEDITOR
    Detail_settings.Terrain_render_distance = 80.0 * TERRAIN_SIZE;
#endif
    return;
  }
```
That's within the `CheckCellOccupancy` function. If it evaluates to true, it stops rendering the terrain cells. If it's false, terrain gets rendered. The problem is that if it does evaluate to true, the game's render depth setting gets reset to 40--even if the player had it at 100. The solution to the issue is to increase the `MAX_CELLS_TO_RENDER` variable--it's unclear why it was set to such a low value in the first place. The alternative is to comment out this safety check entirely, which would allow excessive rendering and potential stability issues.

JeodC:
Here's a screenshot of the same level from the retail edition:

image

Lgt2x:
Thanks for the clear explanation!

Lgt2x (approved):

PR #340, opened 2024-05-13
Hide module symbols by default

MaddTheSane:
## Pull Request Type

- [ ] GitHub Workflow changes
- [X] Build and Dependency changes
- [X] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [X] Other changes

### Description
This changes the visibility of the symbols of the plug-ins, the net games, and the TCP/IP net module. This can be great for LTO and dead code stripping, as only exported symbols (and symbols that the exported symbols reference) are included, reducing the size of linked objects.

### Checklist

### Additional Comments
Due to how DMFC is statically-linked and built for Linux and macOS, a lot of the DMFC symbols will still be exported. The next part would be changing visibility of the DMFC library for macOS and Linux to not export its symbols if they're not building for Windows.


MaddTheSane:
An alternative to marking specific symbols as exportable in the source code is to have an exported symbols list that is passed to the linker. The problem is, macOS and Windows start each symbol with an underscore, while Linux/other UNIXes do not. This would resolve the DMFC symbols being exported, though.

JeodC:
I did some research on this subject and came up with some possible downsides:

- Compatibility-wise we can't test this yet because netgames aren't fully working (last I tried). You ticked the box for that though--did you test with retail netgames, then?
- Reduced visibility will make debugging more of a challenge in an area we haven't touched extensively. The filesizes of these things aren't overtly huge, so is hidden symbols really warranted at this time?

MaddTheSane:
* It doesn't just target net games, but also the level scripts.
* If you feel like this is too soon, you can wait until the project is a bit more mature.

JeodC:
> * It doesn't just target net games, but also the level scripts.
> * If you feel like this is too soon, you can wait until the project is a bit more mature.

@jcoby @winterheart @Arcnor What do you think?

winterheart:
I have little experience on dll using, but changes looks fine for me.

netgames/entropy/CMakeLists.txt
```diff
@@ -4,6 +4,7 @@ set(CPPS EntropyBase.cpp EntropyPackets.cpp EntropyRoom.cpp)
 set(NETGAME_MODULE entropy)
 
 add_library(${NETGAME_MODULE} MODULE ${CPPS} ${HEADERS})
+set_target_properties(${NETGAME_MODULE} PROPERTIES CXX_VISIBILITY_PRESET "hidden")
```
Lgt2x, line 7: I'm thinking about writing a custom CMake command for netgames that are all built the same way, maybe for a future contribution

Lgt2x (approved):

PR #343, opened 2024-05-13
Reduce coupling between headers associated to lanclient

Lgt2x:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

- mdllinit.h is now a standalone header instead of inline include
- MULTI_USE_ALL_OPTIONS define is no more, decoupling con_dll.h from lanclient.cpp
- con_dll.h has include guards and required includes to make it standalone

NEEDS THOROUGH MULTIPLAYER TESTING

### DIFF
--- a/netcon/includes/con_dll.h
+++ b/netcon/includes/con_dll.h
@@ -264,8 +264,15 @@
  */
 ///////////////////////////////////////////////
 
+#pragma once
+
 #include "ship.h"
 #include "pstypes.h"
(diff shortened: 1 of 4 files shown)


Lgt2x:
outdated

PR #352, opened 2024-05-15
Cleanup comments in save & load related files

JeodC:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

- Remove commented code
- Move versioning comments to header
- Add comments in gamesave.cpp for readability


Descent3/gamesave.cpp
```diff
@@ -319,23 +319,26 @@ extern int Times_game_restored;
 // available for all.
 int Quicksave_game_slot = -1;
 
+// Quicksave function
```
Lgt2x, line 322: Most added comments in this file are pointless and only add noise

Descent3/loadstate.cpp
```diff
@@ -1603,423 +1528,4 @@ int LGSSnapshot(CFILE *fp) {
   }
 
   return bm_handle;
-}
-
-//@@	// movement info.
-//@@		switch (op->movement_type)
-//@@		{
-//@@		case MT_SHOCKWAVE:
```
Lgt2x, line 1611: do you have any idea what this whole code block could have been for?

JeodC: If you search `@` you'll find a lot of files with commented blocks like this one. In this block, it looks like old loading code that was replaced.

Descent3/gamesave.cpp
Lgt2x: Well you should learn to read code. `Game_mode & GM_MULTI` is as clear as `// Check if game is in multiplayer mode`, etc. If code is not understandable easily, it's likely because of bad naming. Read Clean Code chap 2 and 4.

Issue #312, opened 2024-05-08
[Runtime Issue]: Newly built dmfc.dll and gamemodes crash


JeodC:
### Build Version

current

### Operating System Environment

- [ ] Microsoft Windows (32-bit)
- [X] Microsoft Windows (64-bit)
- [ ] Mac OS X
- [ ] Linux (specify distribution and version below)

### CPU Environment

- [ ] x86 (32-bit Intel/AMD)
- [X] x86\_64 (64-bit Intel/AMD)
- [ ] ARM (32-bit)
- [ ] ARM64 (64-bit; sometimes called AArch64)
- [ ] Other (RISC V, PPC...)

### Game Modes Affected

- [ ] Single player
- [X] Anarchy
- [X] Hyper-Anarchy
- [X] Robo-Anarchy
- [X] Team Anarchy
- [X] Capture the Flag
- [X] Bounty
- [X] Entropy
- [X] Hoard
- [X] Monsterball
- [X] Cooperative

### Description

If one builds a new `dmfc.dll` (done by making `netgames/dmfc` a SHARED library as it needs to be), and then attempts to use it (to resolve #213), the dedicated server will immediately crash. This occurs with all game modes.

Tested with:

- x86 Windows CD v1.4 patched d3ms
- Steam release d3ms
- Newly built d3ms

GDB traces are likely different for various cases. If you don't have a `dedicated.cfg` file, create one and paste the following into it:
```
[server config file]
PPS=8
TimeLimit=30
KillGoal=0
RespawnTime=30
GameName=Dedicated Server!
MissionName=fury.mn3
Scriptname=ANARCHY
ConnectionName=Direct TCP~IP
SetLevel=1
SetDifficulty=0
```

### Steps to Reproduce

- Build `dmfc.dll` as a `SHARED` library and add it to your installdir
- Run `Descent3.exe -dedicated dedicated.cfg` (may take a few tries for the server to pass I/O Init, a separate bug)
- Observe crash

JeodC:
Piccu Engine ships with a `dmfc.dll` that functions. We can bisect with: https://github.com/DescentDevelopers/Descent3/commits/main/netgames

JeodC:
This is caused by https://github.com/DescentDevelopers/Descent3/commit/c9d59b90739a190aa71b4adde8bf3f4c101e05e9

pzychotic:
[...]

PR #297, opened 2024-05-06
Revamp usage instructions

Jayman2000:
## Pull Request Type

- [ ] GitHub Workflow changes
- [x] Documentation or Wiki changes
- [x] Build and Dependency changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

This commit revamps the usage instructions in order to accomplish two goals:

1. Make the instructions easier to follow by turning them into a numbered list.
2. Make sure that macOS and Linux users install all of the built libraries for their platforms.

In order to accomplish goal number 2, this PR makes it so that the `online/` and `netgames/` directories are automatically generated for macOS and Linux users.

### Related Issues

Resolves #120. Fixes #369.

### Checklist

- [x] I have tested my changes locally and verified that they work as intended.
- [x] I have documented any new or modified functionality.
- [x] I have reviewed the changes to ensure they do not introduce any unnecessary complexity or duplicate code.
- [x] I understand that by submitting this pull request, I am agreeing to license my contributions under the project's license.

### Additional Comments

Here are some things that I wasn’t sure about:
- One of the commits creates a new CMake target named `NetgamesDir`. I wasn’t sure if I should call it `NetgamesDir` or `NetGamesDir`. I went with `NetgamesDir` because the folder itself doesn’t do anything to separate the two words.
- One of the commits adds text that tells users to copy “All `.hog` files, except for `d3-linux.hog`”. This is done to make sure that an old version of `d3-linux.hog` doesn’t get used. Is there a macOS equivalent of that file? [The Steam version of Descent 3 for macOS uses `d3-linux.hog`.](https://steamdb.info/depot/273593/) Does the GraphSim version of Descent 3 have something similar to `d3-linux.hog`?

PR #354, opened 2024-05-15
Windows x64 build

pzychotic:
## Pull Request Type

- [ ] GitHub Workflow changes
- [x] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [x] Other changes

### Description

Getting this out as a starting point for a possible Windows x64 build.

So far my local tests in Singleplayer work fine. I haven't tried any Multiplayer.

**Quick summary:**
- DirectX SDK updated with x64 support
- Prevent 32bit truncations of WinApi types
- Compilation of scripts and netgames projects

**Cut corners:**
- Disabled callstack and exception info logging: https://github.com/pzychotic/Descent3/commit/caf19810158723708ac93dbf1880148c132f5a1d
  - Maybe this can be replaced with writing a MiniDump file nowadays?
- Disabled some code for 'mono' debugger: https://github.com/pzychotic/Descent3/commit/bac882c7d6a47e3ec990be25dd339101115cef09
  - Do we even still need this?

**Open questions:**
- I had to load the new d3-win.hog last so the x64 DLLs from it will be found before the x86 DLLs from the original game data: https://github.com/pzychotic/Descent3/commit/435b1f48bb101295d868b426a4de3ec8edf53378
  - Shouldn't this also be done for Linux/Mac? 
- Array access with negative index in dsound3d.cpp only crashes in x64: https://github.com/pzychotic/Descent3/commit/11af2e7ec0540001e023f8380aa8491773d09460
  - This problem is already present in x86 builds, but doesn't crash...

**Known Issue:**
- Savegames from 32bit and 64bit versions aren't compatible: https://github.com/DescentDevelopers/Descent3/discussions/366

### Checklist


Lgt2x:
Hey, that looks promising! Just so you know, we will be working on replacing DirectX with SDL2 on Windows just like Linux and MacOS do, so don't spend too much effort on the DirectX code

pzychotic:
> Hey, that looks promising! Just so you know, we will be working on replacing DirectX with SDL2 on Windows just like Linux and MacOS do, so don't spend too much effort on the DirectX code

Thanks for the heads-up on DirectX stuff. I wasn't planning spending time on it, just replaced the SDK for the x64 support to get the current stuff to compile and run.

GravisZro:
Don't forget `dd_sndlib/Ds3dlib.cpp`

MaddTheSane:
> 
> * I had to load the new d3-win.hog last so the x64 DLLs from it will be found before the x86 DLLs from the original game data:[pzychotic@435b1f4](https://github.com/pzychotic/Descent3/commit/435b1f48bb101295d868b426a4de3ec8edf53378)
>   
>   * Shouldn't this also be done for Linux/Mac?

Maybe for Linux, but macOS can have multiple architectures in one binary.

Descent3/init.cpp
JeodC, line 1: Should this also be ifdef WIN64, or just generic?

CMakeLists.txt
```diff
@@ -199,9 +200,5 @@ add_subdirectory(Descent3)
 add_subdirectory(tools)
 add_subdirectory(netcon)
 
-# For now we don't need to build the scripts under windows, so we'll only include
-# the directory when building for linux/osx. In the future we may want to to fix bugs, etc.
-if(UNIX)
-  add_subdirectory(netgames)
-  add_subdirectory(scripts)
-endif()
+add_subdirectory(netgames)
```
JeodC, line 203: Since netgames are now being built, can we add them to the netgames directory in Descent3\Release or Descent3\Debug to package them with artifacts?

Descent3/init.cpp
JeodC: > If we want to use our new code, especially for the fixed crashes concerning the dmfc lib, we need it to be generic.

By that I meant just `WIN`, not WIN32 or WIN64. I suppose WIN32 is interchangeable here?

CMakeLists.txt
pzychotic: Sure.
I need to look at the CMake stuff, how best to it. Or someone points me to a good solution for it.

Descent3/init.cpp
pzychotic: Oh I see. _WIN32 is always defined for Windows, regardless of 32 or 64 bit. And that define is already used everywhere else.

CMakeLists.txt
JeodC: It's already done for the TCP~IP.d3c file. You could use that as a template.

pzychotic: Thanks for the pointer.
I was about to add something like `set_target_properties(${NETGAME_MODULE} PROPERTIES LIBRARY_OUTPUT_DIRECTORY "$<TARGET_FILE_DIR:Descent3>/netgames")` to every CMake file in the netgames subdirectories, before I noticed that @Jayman2000 already has a custom target defined in #297 that copies the files. Maybe we can land his changes before I try to solve this differently.

JeodC: Indeed, will try to get that merged today.

JeodC:
I can't review further, other than verifying it works in windows, but I'd like to note that as pzychotic [nicely explained](https://github.com/DescentDevelopers/Descent3/issues/312#issuecomment-2119263229), this pull will resolve the outstanding multiplayer issues.

JeodC (approved):

CMakeLists.txt
```diff
@@ -95,7 +95,7 @@ elseif(CMAKE_SYSTEM_NAME STREQUAL "Windows")
           #[[NOGDI]] # We need GDI for now, enable when GDI is actually not needed (or when all windows.h mentions are gone)
   )
 
-  list(APPEND CMAKE_LIBRARY_PATH "lib/win" "lib/win/directx")
+  list(APPEND CMAKE_LIBRARY_PATH "lib/win" "lib/win/directx/${CMAKE_CXX_COMPILER_ARCHITECTURE_ID}")
```
winterheart, line 98: Since there changes happens... Can you please change it to `lib/win/DirectX/...`?

Descent3/init.cpp
```diff
@@ -1541,6 +1541,13 @@ void InitIOSystems(bool editor) {
   ddio_MakePath(fullname, LocalD3Dir, "extra13.hog", NULL);
   extra13_hid = cf_OpenLibrary(fullname);
 
+  // last library opened is the first to be searched for DLLs, so put this one
+  // at the end to find our newly build script DLLs first
+#ifdef _WIN32
+  ddio_MakePath(fullname, LocalD3Dir, "d3-win.hog", NULL);
+  sys_hid = cf_OpenLibrary(fullname);
+#endif
+
```
winterheart, line 1550: Better to merge this with other d3-<PLATFORM>.hog loading. After that we can move `sys_hid = cf_OpenLibrary(fullname);` out of #ifdefs.

winterheart (commented):
We need additionally mention somewhere that we use static DirectX libraries (and here issue about licensing them with GPLv3 code).

CMakeLists.txt
pzychotic: Good catch, done.

Descent3/init.cpp
pzychotic: Done.

pzychotic:
> We need additionally mention somewhere that we use static DirectX libraries (and here issue about licensing them with GPLv3 code).

These are not static libraries, these are just the link libs needed by the linker to resolve the symbols to the DLLs.

winterheart (approved):
OK from me. Good work!

PR #356, opened 2024-05-16
Replace ioctl with fcntl, fix ioctlsocket calls (64-bit fix)

GravisZro:
These take two different types of arguments but worked in the past due to long being 32-bit.

## Pull Request Type

- [ ] GitHub Workflow changes
- [x] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [x] Network changes
  - [ ] Other changes

### Description
The `ioctl` argument for UNIX systems is `int*` while the `ioctlsocket` argument for Windows is `unsigned long*`. On 32-bit systems theses coincided so no problems but not on 64-bit. However, the behavior of `ioctl` calls is non-standard which is why [code should use `fnctl`.](https://stackoverflow.com/questions/1150635/unix-nonblocking-i-o-o-nonblock-vs-fionbio#answers)

See also:
- [ioctlsocket documentation](https://learn.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-ioctlsocket)
- [fnctl documentation](https://pubs.opengroup.org/onlinepubs/9699919799/functions/fcntl.html)
- [ioctl documentation](https://pubs.opengroup.org/onlinepubs/9699919799/functions/ioctl.html)

### Checklist

### DIFF
--- a/lib/networking.h
+++ b/lib/networking.h
@@ -237,7 +237,10 @@ static inline void INADDR_GET_SUN_SUNB(struct in_addr *st, uint8_t *s_b1, uint8_
 #define WSAEINVAL EINVAL
 #define WSAENOPROTOOPT ENOPROTOOPT
 
-static inline int WSAGetLastError() { return errno; }
+#ifndef WSAGetLastError
+#define WSAGetLastError() errno
+#endif
+
 extern bool Use_DirectPlay;
 
 // helper macros for working with SOCKADDR_IN to make it look nicer between windows and Linux
@@ -343,6 +346,10 @@ typedef struct {
 extern BOOL DP_active;
 extern BOOL TCP_active;
 extern BOOL IPX_active;
+
+// create a new non-blocking socket
+int make_nonblocking(SOCKET sock);
+
 // Get the info from RAS
 uint32_t psnet_ras_status();
 
--- a/Descent3/dedicated_server.cpp
+++ b/Descent3/dedicated_server.cpp
@@ -140,6 +140,7 @@ typedef int socklen_t;
 #include "init.h"
 #include "ship.h"
 #include "hud.h"
+#include "networking.h"
 
 
 bool Dedicated_server = false;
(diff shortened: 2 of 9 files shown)


winterheart:
I would rewrite socket initialization as
```cpp
#ifdef WIN32

dedicated_listen_socket = socket(AF_INET, SOCK_STREAM, 0);

// ...

unsigned long arg = 1;
ioctlsocket(dedicated_listen_socket, FIONBIO, &arg);

#else // WIN32

#ifndef SOCK_NONBLOCK
#include <fcntl.h>
#define SOCK_NONBLOCK O_NONBLOCK
#endif

dedicated_listen_socket = socket(AF_INET, SOCK_STREAM | SOCK_NONBLOCK, 0);

#endif
```

And you don't have to use additional `fcntl(..., F_SETFL (fcntl(..., F_GETFL) ))` calls on Linux / macOS platforms. There a plenty code duplication around socket initialization, so would be some helper functions would be useful.

GravisZro:
@winterheart Not all of these are socket initializations. As such, I consolidated the code into a single function in `networking.h`. Including `networking.h` in other headers lead to conflicting duplicate code which is why there are now code deletions.

Lgt2x (approved):

lib/networking.h
```diff
@@ -343,6 +346,10 @@ typedef struct {
 extern BOOL DP_active;
 extern BOOL TCP_active;
 extern BOOL IPX_active;
+
+// create a new non-blocking socket
+int make_nonblocking(SOCKET sock);
```
jcoby, line 351: This should probably be called something like `nw_SocketSetNonblock()` to match the rest of the functions defined here.

GravisZro: It was considered but it's used on other sockets as well where the `nw_*` scheme wouldn't be appropriate.

winterheart (approved):

GravisZro:
Would someone please merge this? It's been reviewed by two people already.

winterheart:
PR has regression on Linux, see #398.

PR #361, opened 2024-05-17
Converted line endings to Unix format to match .gitattributes file

pzychotic:
### Description

There are some files with Windows or mixed line endings that should be Unix style according to the .gitattributes file.
This messes up my local clone on Windows, since git constantly shows these file as modified even if they aren't.


JeodC (approved):
The string tables need a pass to correct accented characters for languages, but this can be done later. Correcting the line endings is a priority to resolve commit conflicts for contributors.

PR #372, opened 2024-05-19
Get executable path for -useexedir

### LINKED ISSUE #322: [Runtime Issue]: `-useexedir` causes game to crash instantly
### Build Version

v1.5.0 7d38dde2

- [ ] Mac OS X

### CPU Environment

- [ ] x86 (32-bit Intel/AMD)
- [ ] ARM (32-bit)
- [ ] Other (RISC V, PPC...)

### Game Modes Affected

- [X] Single player
- [X] Anarchy
- [X] Hyper-Anarchy
- [X] Robo-Anarchy
- [X] Team Anarchy
- [X] Capture the Flag
- [X] Bounty
- [X] Entropy
- [X] Hoard
- [X] Monsterball
- [X] Cooperative

### Game Environment

I’m running NixOS 23.11 and using game data from the Windows Dual-Jewel version of Descent 3 v1.4.

### Description

Whenever I try to run the game with `-useexedir`, it crashes with these errors:

```
Error: Unable to set temporary directory to: "./Descent3/custom/cache"

Error: Unable to set temporary directory to: "./Descent3/custom/cache"(Descent 3 Error)

System Error

Press OK to exit, CANCEL to ignore this error and continue.
SIGNAL 5 caught, aborting
DDIO: ddio_InternalClose() called.
DDIO: ddio_InternalClose() returning.
```

### Regression Status

I don’t know.

### Steps to Reproduce

1. Run `Descent3 -useexedir`.

JeodC:
## Pull Request Type

- [ ] GitHub Workflow changes
- [x] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [x] Other changes

### Description

Actually retrieve the executable directory from a relative path so `-useexedir` functions. This was tested in windows where the realpath was `S:\Descent 3\Descent3.exe` and from a terminal in S:\ ran `start ./Descent 3/Descent3.exe" --useexedir`. Verified working.

### Related Issues

Resolves #322 if Linux/MacOS are also functional. Needs testing.


JeodC:
@Jayman2000 Can you try this and let me know if it works?

Descent3/init.cpp
```diff
@@ -1413,33 +1424,63 @@ static const tTempFileInfo temp_file_wildcards[] = {{"d3s*.tmp"}, {"d3m*.tmp"},
...
+    if (dirarg) {
+        strncpy(Base_directory, GameArgs[dirarg + 1], sizeof(Base_directory) - 1);
+        Base_directory[sizeof(Base_directory) - 1] = '\0';
+    } else if (exedirarg) {
+        memset(exec_path, 0, sizeof(exec_path));
+        // Populate exec_path with the executable path
+        if (getExecutablePath(exec_path, sizeof(exec_path)) == NULL) {
+            Error("Failed to get executable path\n");
+        } else {
+            // Extract directory from the executable path
```
winterheart, line 1467: Probably we can use here `std::filesystem` functions from C++17, code would much cleaner and easier.
Like

```cpp
Base_directory = std::filesystem::path::root_path(Base_directory).c_str();
```

Descent3/init.cpp
```diff
@@ -1413,33 +1424,63 @@ static const tTempFileInfo temp_file_wildcards[] = {{"d3s*.tmp"}, {"d3m*.tmp"},
                                                     {"d3c*.tmp"}, {"d3t*.tmp"}, {"d3i*.tmp"}};
 static const int num_temp_file_wildcards = sizeof(temp_file_wildcards) / sizeof(tTempFileInfo);
 
+
+// Get the binary path
+char exec_path[_MAX_PATH];
+char* getExecutablePath(char *exec_path, size_t len) {
```
winterheart, line 1430: Better to move it to ddio module.

Descent3/init.cpp
JeodC: It's not really any less lines, but it works.

Descent3/init.cpp
JeodC: Where would you move it to? I think this gets called before ddio is initialized.

winterheart: ddio_lnx/lnxfile.cpp / ddio_win/winfile.cpp has some generic file / directory operations.

Jayman2000 (changes requested):
I tested this, and it does indeed fix the problem. On an unrelated note, it looks like you accidentally used four spaces for indentation. The rest of the code uses two spaces for indentation. I would change this PR to use two spaces for indentation.

Descent3/init.cpp
```diff
@@ -986,6 +988,17 @@
...
+/*
+#ifdef _WIN32
+#include <windows.h>
+#include <shlwapi.h>
+#pragma comment(lib, "shlwapi.lib")
+#else
+#include <limits.h>
+#include <unistd.h>
+#endif
+*/
```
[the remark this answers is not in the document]
JeodC: Old code from before implementing winterhearts suggestion. Commented in case it needed to be reverted, but forgot to remove after successful build and test.

ddio_lnx/lnxio.cpp
```diff
@@ -99,3 +99,11 @@ void ddio_DebugMessage(unsigned err, char *fmt, ...) {
   mprintf((0, "%s\n", buf));
 }
 
+char* getExecutablePath(char *exec_path, size_t len) {
+  if (realpath("/proc/self/exe", exec_path) == NULL) {
```
jcoby, line 103: this won't work on macos. it needs to be a call to `_NSGetExecutablePath`

lib/ddio.h
```diff
@@ -320,6 +320,9 @@ void ddio_MouseSetVCoords(int width, int height);
 //	File Operations
 //	---------------------------------------------------------------------------
 
+// Gets the full path of the executable file
+char* getExecutablePath(char *exec_path, size_t len);
```
jcoby, line 324: ddio functions should probably retain the `ddio_` prefix?

Descent3/init.cpp
```diff
@@ -1418,26 +1419,29 @@ static const int num_temp_file_wildcards = sizeof(temp_file_wildcards) / sizeof(
...
+    } else {
+        std::filesystem::path executablePath(exec_path);
+         std::string baseDirectoryString = executablePath.parent_path().string();
+         strncpy(Base_directory, baseDirectoryString.c_str(), sizeof(Base_directory) - 1);
+         Base_directory[sizeof(Base_directory) - 1] = '\0';
+         mprintf((0, "Using working directory of %s\n", Base_directory));
+      }
+    } else {
+        ddio_GetWorkingDir(Base_directory, sizeof(Base_directory));
       }
```
Jayman2000, line 1444: It looks like the indentation still isn’t quite right. I would change it to this:

```suggestion
    } else {
      std::filesystem::path executablePath(exec_path);
      std::string baseDirectoryString = executablePath.parent_path().string();
      strncpy(Base_directory, baseDirectoryString.c_str(), sizeof(Base_directory) - 1);
      Base_directory[sizeof(Base_directory) - 1] = '\0';
      mprintf((0, "Using working directory of %s\n", Base_directory));
    }
  } else {
    ddio_GetWorkingDir(Base_directory, sizeof(Base_directory));
  }
```

Jayman2000 (changes requested):
I tested the code again. It still works, and it still fixes the problem.

Descent3/init.cpp
```diff
@@ -1413,31 +1426,55 @@ static const tTempFileInfo temp_file_wildcards[] = {{"d3s*.tmp"}, {"d3m*.tmp"},
                                                     {"d3c*.tmp"}, {"d3t*.tmp"}, {"d3i*.tmp"}};
 static const int num_temp_file_wildcards = sizeof(temp_file_wildcards) / sizeof(tTempFileInfo);
 
+
+// Get the binary path
+char exec_path[_MAX_PATH];
+char* getExecutablePath(char *exec_path, size_t len) {
```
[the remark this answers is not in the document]
Jayman2000: I noticed that you marked this as resolved but chose to continue to return a `char *`. Why did you choose to return a `char *`?

ddio_lnx/lnxio.cpp
JeodC: Hopefully I did that right. I'm not familiar with mac filesystem stuff.

jcoby: looks and works good

Lgt2x:
Merged without approval

Jayman2000:
Unfortunately, it looks like the final version of this PR introduced a new indentation mistake:

```c++
    if (!ddio_GetBinaryPath(exec_path, sizeof(exec_path))) {
    Error("Failed to get executable path\n");
    } else {
```

JeodC:
> Unfortunately, it looks like the final version of this PR introduced a new indentation mistake:
> 
> 
> 
> ```c++
> 
>     if (!ddio_GetBinaryPath(exec_path, sizeof(exec_path))) {
> 
>     Error("Failed to get executable path\n");
> 
>     } else {
> 
> ```

Dang it, missed one! I'll get it in a second pass the next time I touch init.cpp. Good catch.

PR #289, opened 2024-05-05
Implementing own MVE library

winterheart:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Documentation or Wiki changes
- [x] Build and Dependency changes
- [x] Runtime changes
  - [x] Render changes
  - [x] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

Implementing own MVE library under GPLv3+ license based on D2X work. Currently works video with audio and framed playback (for main menu) on Linux and probably macOS. Windows build is expected to fail as need to adapt DirectSound PCM playback, and there I can use some help with experience of DirectX 6 programming.

Old MVE implementation still here in codebase, we need to decide what to do with it.

### Related Issues

Fixes #253.

### Checklist

- [x] I have tested my changes locally and verified that they work as intended.
- [x] I have documented any new or modified functionality.
- [x] I have reviewed the changes to ensure they do not introduce any unnecessary complexity or duplicate code.
- [x] I understand that by submitting this pull request, I am agreeing to license my contributions under the project's license.

PR #380, opened 2024-05-21
Fix cross-compiling "int $3" build error

### LINKED ISSUE #311: [Build Failure]: fix "bad instruction 'int $3'" error on the Raspberry Pi 4 32 bit
### Build Version

v1.5

- [ ] Mac OS X

### CPU Environment

- [ ] x86 (32-bit Intel/AMD)
- [X] ARM (32-bit)
- [ ] Other (RISC V, PPC...)

### Description

For the game to compile in raspberry pi os 32 bit I had to change mveasm.cpp. In line 2060 I changed "#define int3 asm volatile("int $3");" to "#define int3". I did this to fix a "bad instruction 'int $3'" error on the Raspberry Pi 4

### Regression Status

_No response_

### Steps to Reproduce

_No response_

GravisZro:
`libmve/mveasm.cpp` was using a macro to invoke `int $3` instead of using the portable `Int3()` macro in `pserror.h`.

## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description
Replace non-portable macro `int3` with the portable macro `Int3()` from `pserror.h`.
Fixes #311

### Checklist

### Additional Comments

I've cross-compiled a the libmve module but not the rest of the code because I don't have a port of SDL2 for that.

### DIFF
--- a/ddebug/windebug.cpp
+++ b/ddebug/windebug.cpp
@@ -189,7 +189,7 @@
  */
 
 #include "d3_version.h"
-#include "Debug.h"
+#include "debug.h"
 #include "mono.h"
 
 #include <windows.h>
--- a/lib/pserror.h
+++ b/lib/pserror.h
@@ -232,7 +232,7 @@ static inline void SetDebugBreakHandlers(void (*stop)(), void (*resume)()) {
     if (_heapchk() != _HEAPOK)                                                                                         \
       Int3();                                                                                                          \
   } while (0)
-#elif defined(LINUX)
+#elif defined(__LINUX__)
 #include "SDL.h"
 // For some reason Linux doesn't like the \ continuation character, so I have to uglify this
 #define DEBUG_BREAK()                                                                                                  \
--- a/libmve/mveasm.cpp
+++ b/libmve/mveasm.cpp
@@ -16,6 +16,7 @@
 **
 */
 #include "pstypes.h"
+#include "pserror.h"
 #include "mvelibl.h"
 #include "mvelibi.h"
 #include <string.h>
@@ -2053,20 +2054,6 @@ void gfxVres(unsigned char misc, unsigned char *crtc);
 void MVE_gfxWaitRetrace(int state);
 void MVE_gfxSetSplit(unsigned line);
 
-#if defined(__LINUX__)
-
-#if !defined(MACOSXPPC) && !defined(__aarch64__)
-
-#define int3 __asm__ __volatile__("int $3");
-#else
-#define int3
-#endif
-#else
-#define int3                                                                                                           \
-  __asm { int 3}                                                                                                        \
-  ;
-#endif
-
 void DECOMP_BODY(bool HI_COLOR_FLAG, const unsigned char *&comp, unsigned int _x, unsigned int _y, unsigned int _w,
                  unsigned int _h) {
   unsigned int HI_COLOR_SCALE = (HI_COLOR_FLAG) ? 2 : 1;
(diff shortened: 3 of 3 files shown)


JeodC:
Potentially dumb question, why not use `debugbreak()` as suggested in the issue? I also recommend tagging winterheart for review once this is prepared since the new libmve may not use asm--and therefore this change not needed in the long run.

GravisZro:
> Potentially dumb question, why not use `debugbreak()` as suggested in the issue?

The `Int3()` macro is a bit more sophisticated for debugging but it does actually end up calling the `debug_break` function.

> I also recommend tagging winterheart for review 

As far as I can tell, I don't have the ability to do that.

JeodC:
> > Potentially dumb question, why not use `debugbreak()` as suggested in the issue?
> 
> The `Int3()` macro is a bit more sophisticated for debugging but it does actually end up calling the `debug_break` function.
> 
> > I also recommend tagging winterheart for review
> 
> As far as I can tell, I don't have the ability to do that.

Done. Over in #289, `libmve.asm` is completely gone. I'm not sure how close we are to merging that though.

pzychotic:
Uhm, I also changed that file in #354 for Windows x64 to compile, but opted for the `debug_break()` route. I wasn't aware, we have an issue for that.
If the `Int3()` solution is preferable, I should revert my change in favor of this.

JeodC:
> Uhm, I also changed that file in #354 for Windows x64 to compile, but opted for the `debug_break()` route. I wasn't aware, we have an issue for that. If the `Int3()` solution is preferable, I should revert my change in favor of this.

`Int3()` is a simple macro that's strategically placed in the codebase to quickly identify issues that should only occur in rare circumstances. Unlike `debug_break()` it doesn't trigger breakpoints directly. The latter is more platform-dependent from what I gather. From my understanding, `debug_break()` should be preferred when using debug builds, but `Int3()` should be used in release builds for easier user-friendly reporting.

JeodC:
Actually, at the moment `Int3()` is rather unhelpful. It's defined at `pserror.h` where it calls `debug_break()` if it's available. If not, then it just does the `int3` messagebox. The `mprintf` statement, however, is the user-friendly part I was referring to.
https://github.com/DescentDevelopers/Descent3/blob/main/lib/pserror.h#L222-L229

There are tons of `mprintf` statements scattered throughout the codebase. It used to output to a monochrome debug console:
https://github.com/DescentDevelopers/Descent3/blob/main/ddebug/mono.h#L72
https://github.com/DescentDevelopers/Descent3/blob/main/ddebug/winmono.cpp#L370-L410

We have a form of spdlog in the code now (header only iirc, unless it did get removed entirely), but I don't know to what extent it takes advantage of existing debug messages.

pzychotic:
It works in Windows too, after we fixed it in #256 (https://github.com/DescentDevelopers/Descent3/commit/b59115460eb0d608a3fdd4bfca206dc56c23fc38).
It outputs to the VS output window https://github.com/DescentDevelopers/Descent3/blob/cffeb20a3e09d8c869134a465555374b7a9b40ae/ddebug/winmono.cpp#L407 (or any other debugger supporting it) and to a log file https://github.com/DescentDevelopers/Descent3/blob/cffeb20a3e09d8c869134a465555374b7a9b40ae/ddebug/winmono.cpp#L404 when activated with `-logfile`.

JeodC:
> -DLOGGER works in Windows too, after we fixed it in #256 (https://github.com/DescentDevelopers/Descent3/commit/b59115460eb0d608a3fdd4bfca206dc56c23fc38).
> 
> It outputs to the VS output window https://github.com/DescentDevelopers/Descent3/blob/cffeb20a3e09d8c869134a465555374b7a9b40ae/ddebug/winmono.cpp#L407 (or any other debugger supporting it) and to a log file https://github.com/DescentDevelopers/Descent3/blob/cffeb20a3e09d8c869134a465555374b7a9b40ae/ddebug/winmono.cpp#L404 when activated with `-logfile`.

That explains it. I expected it to output to terminal when I ran the exe with it. We were going to have a rough writeup of how to use logging, but it hasn't come about yet.

Lgt2x:
`mveasm.cpp` will be removed by #289, there is not point in fixing the current version of libmve

GravisZro:
> `mveasm.cpp` will be removed by #289, there is not point in fixing the current version of libmve

I don't understand why you close this PR. I'm trying to resolve #311. If nothing else, it will fix the build problem for ARM.

PR #387, opened 2024-05-23
Fix RAND_MAX overflow warnings

GravisZro:
In `psrand.h` it attempts to undefine RAND_MAX and then define it once again. This may work on some compilers but not G++/Clang. To resolve this error RAND_MAX was renamed to D3_RAND_MAX but **only** in files `.cpp` that included `psrand.h`. The code behavior should restored to that of the official release.

**NOTE:** Not _every_ overflow is fixed because as far as I can tell, that's how it is in the official release as `psrand.h` is not included everywhere RAND_MAX is used.

## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description
When compiling D3, there are several overflow warning regarding `RAND_MAX + 1` because the value of `RAND_MAX` contains the same as INT32_MAX (`0x7FFFFFFF`). However, `psrand.h` does `#undef RAND_MAX` and then defines it as 0x7FFF. This does not work with G++ and Clang and the value `0x7FFFFFFF`is retained. Since compilers assume a given value is a signed integer (unless specified otherwise) adding one will cause an overflow and thus causing a overflow warning. This is a sign that the value of RAND_MAX was not redefined to `0x7FFF`. The solution is to simply rename references to `RAND_MAX` that _would_ be redefined (on a more forgiving compiler) from `RAND_MAX` to `D3_RAND_MAX`.

However, not all parts of the code that use `RAND_MAX` also include `psrand.h` which means the ISO `RAND_MAX` value was likely used in the official release. This is why some overflow warnings will still persist after this patch.

### Checklist

### DIFF
--- a/lib/psrand.h
+++ b/lib/psrand.h
@@ -33,12 +33,15 @@
  *
  * $NoKeywords: $
  */
+#ifndef PS_RAND_H
+#define PS_RAND_H
 
 #include <cstdint>
-#undef RAND_MAX
 
-#define RAND_MAX 0x7fff
+#define D3_RAND_MAX 0x7fff
 
-void ps_srand(unsigned int seed);
+void ps_srand(uint32_t seed);
 
-int ps_rand(void);
+int32_t ps_rand(void);
+
+#endif
--- a/Descent3/AIGoal.cpp
+++ b/Descent3/AIGoal.cpp
@@ -517,7 +517,7 @@ void GoalInitWanderAround(object *obj, goal *goal_ptr) {
   goal_ptr->g_info.pos = pos;
   goal_ptr->g_info.roomnum = roomnum;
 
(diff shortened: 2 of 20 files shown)


lib/psrand.h
```diff
@@ -34,9 +34,7 @@
  * $NoKeywords: $
  */
 
-#undef RAND_MAX
```
Lgt2x, line 37: This file seems to be missing include guards

bryanperris: I approve these changes, the original file didn't have the guard either.

bryanperris (approved):

PR #389, opened 2024-05-24
Add option to disable git hash

GravisZro:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description
Sometimes I need to make builds in deterministic fashion which has made the mandatory git hash being added quite annoying.  I was forced to restructure how the d3_version.h file was made. It's much simpler now: default values unless overridden by compiler defines.  The whole thing was overly complex to start with.

The new option is called `ENABLE_GITHASH` and it's a boolean value that defaults to `ON`.

### Checklist


bryanperris:
Is this something that bothers you for when its in the binary filename or when its displayed in the game?

winterheart:
> Ok, this is another case of nobody reading the actual PR.
> 
> ### winterheart
> > Sorry, but knowing on what commit user gets troubles is more useful. You also removed project based version generation, and I don't see any reason why we shouldn't use this feature.
> 
> @winterheart I know that which is why it's `ENABLED BY DEFAULT` and can only be disabled by a cmake option. Read what I wrote, dammit because nobody is removing the feature. You closed this PR waaaaay too fast because you have jumped to conclusions. `READ THE POST.`

I do not agree with your approach "It's annoying me, I cannot understand logic behind it and I want it be removed". With your rational on disabling git_hash (which nobody else complained) you entirely removing version generation based on project property without any cause for it. Instead of that you added `add_compile_definitions` and now on EVERY commit WHOLE project will be recompiled just because global `D3_GIT_HASH` define was changed. This should be avoided in order speed up building.

So please, be polite, be nice and get ready that some of your contributions can be rejected.

JeodC:
I don't think the motivation here is removal of an "annoying" feature. However, winterheart is correct that in effort to allow deterministic builds, this pull request converted `GIT_HASH` to a _global compile definition_, which means that every time the hash changes, every file in the source gets recompiled. The goal is better accomplished, I think, by selectively including `GIT_HASH` only in files where it's needed, which is the main menu code--could even be done with a header file.

That could have been pointed out in a review or comment rather than a dismissive closure, though.

GravisZro:
@JeodC I agree which part of why #393 exists.

PR #390, opened 2024-05-24
Add new C++ Random functions to replace old C ones.

Nakhr11n:
## Pull Request Type

- [ ] GitHub Workflow changes
- [x] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [x] Other changes

### Description

There are three uses of randomness in the codebase:
 * to get a random int within a range.
 * to get a random float within a range.
 * in if statements to do things x% of the time.

In all cases some variant of the C `rand()` function is used.
In many cases this results in code that is very hard to read, which makes it difficult to understand in what range the randomness will lie.

I propose to add these three functions which add modern C++ equivalents to `rand()` for the three use cases.
The new functions are faster, safer and clearer.

My thinking is to then go over the code and replace all uses of randomness with these functions, and then remove `misc/psrand.cpp` and `lib/psrand.h`.

### Checklist

### DIFF
--- a/Descent3/AIGoal.cpp
+++ b/Descent3/AIGoal.cpp
@@ -505,8 +505,8 @@ void GoalInitWanderAround(object *obj, goal *goal_ptr) {
       pos.y += (ai_info->biased_flight_max + ai_info->biased_flight_min) / 2.0f;
     }
   } else {
-    int tx = (ps_rand() % (TERRAIN_WIDTH - 56)) + (56 / 2);
-    int tz = (ps_rand() % (TERRAIN_DEPTH - 56)) + (56 / 2);
+    int tx = D3::RandomInt(56 / 2, TERRAIN_WIDTH - 56 / 2 - 1);
+    int tz = D3::RandomInt(56 / 2, TERRAIN_DEPTH - 56 / 2 - 1);
 
     int cell = tx + TERRAIN_WIDTH * tz;
     roomnum = 0x80000000 | cell;
@@ -517,7 +517,7 @@ void GoalInitWanderAround(object *obj, goal *goal_ptr) {
   goal_ptr->g_info.pos = pos;
   goal_ptr->g_info.roomnum = roomnum;
 
-  goal_ptr->next_path_time = Gametime + MIN_NEXT_PATH_INTERVAL + ps_rand() / (float)D3_RAND_MAX;
+  goal_ptr->next_path_time = Gametime + MIN_NEXT_PATH_INTERVAL + D3::RandomFloat();
   goal_ptr->flags |= GF_HAS_PATH;
 }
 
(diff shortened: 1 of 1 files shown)


bryanperris:
Is this still work in progress, if so, can you make a draft instead?

random/random.hpp
```diff
@@ -0,0 +1,30 @@
+#pragma once
```
winterheart, line 1: Please add standard copyright header to beginning of each new file:

```
/*
 * Descent 3
 * Copyright (C) 2024 Descent Developers
 *
 * This program is free software: you can redistribute it and/or modify
 * it under the terms of the GNU General Public License as published by
 * the Free Software Foundation, either version 3 of the License, or
 * (at your option) any later version.
 *
 * This program is distributed in the hope that it will be useful,
 * but WITHOUT ANY WARRANTY; without even the implied warranty of
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 * GNU General Public License for more details.
 *
 * You should have received a copy of the GNU General Public License
 * along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
```

random/random.hpp
```diff
@@ -0,0 +1,30 @@
...
+
+#include <cstdint>
+#include <cstdlib>
+
+#include <random>
+#include <type_traits>
+
+namespace Random {
```
winterheart, line 9: Please use `D3` namespace (or `D3::Random` at least).

winterheart:
I'd recommend place it to misc directory and then reuse it as part of misc module. In that way we can gracefully replace old psrand() generator.

Please note, that psrand implementation uses some magic numbers on generating "random" numbers:
```cpp
int32_t ps_rand(void) { return (((ps_holdrand = ps_holdrand * 214013L + 2531011L) >> 16) & 0x7fff); }
```
I cannot say if this really used in order to receive predictable random sequences, but it would be great if we somehow preserve this approach.

bryanperris:
We can keep this to a single PR

GravisZro:
Personally, I prefer the simplicity of `srand` and `rand`.
Also, I found these when working on `RAND_MAX` overflow issue:

- [qRandomChance](https://github.com/DescentDevelopers/Descent3/blob/8485cb0c4d0e5d80e5885850c5d6f69cade03202/scripts/DallasFuncs.cpp#L6125)
- [qRandomValue](https://github.com/DescentDevelopers/Descent3/blob/8485cb0c4d0e5d80e5885850c5d6f69cade03202/scripts/DallasFuncs.cpp#L6145)

If you can make the existing D3 code more readable with some function shortcuts then I'm in favor of it though I'm not sure the PRNG needs work. From what I can tell, the existing code (`ps_rand`) is simplified explicitly to be faster than the lib C `rand`.

Lgt2x:
Have you investigated the use cases of random functions in the game? I'm afraid that changing random functions used (and thus potentially distribution of values) could have side-effects affecting game play that are difficult to predict.

PR #396, opened 2024-05-25
More clean up

GravisZro:
## Pull Request Type

### Description
Clean up:

- Remove self-referential and unused typedefs
- Replace `PSPATHNAME_LEN` with `_MAX_PATH`
- Replace `MAX_PATH` with `_MAX_PATH`
- Remove `MAX_PATH` definition
- Fix `__stdcall` definition

### Checklist

### DIFF
--- a/Descent3/LoadLevel.cpp
+++ b/Descent3/LoadLevel.cpp
@@ -3915,7 +3915,7 @@ int LoadLevel(char *filename, void (*cb_fn)(const char *, int, int)) {
       else if (ISCHUNK(CHUNK_SCRIPT)) {
 #if (defined(EDITOR) || defined(NEWEDITOR))
         if (FindArg("-savescript")) {
-          char path[MAX_PATH], name[MAX_PATH];
+          char path[_MAX_PATH], name[_MAX_PATH];
           ddio_SplitPath(filename, path, name, NULL);
           strcat(path, name);
           strcat(path, ".scr");
--- a/Descent3/Mission.cpp
+++ b/Descent3/Mission.cpp
@@ -679,7 +679,7 @@
 level_info Level_info;
 tMission Current_mission;
 tLevelNode *Current_level = NULL;
-char D3MissionsDir[PSPATHNAME_LEN];
+char D3MissionsDir[_MAX_PATH];
 extern int Times_game_restored;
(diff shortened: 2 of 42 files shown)


winterheart:
`MAX_PATH` is Windows compatibility macros. There no point renaming one C macro to other, this change is not giving any enhancement to the code.

GravisZro:
The end result is fewer defined macros in the code. That's enough of a benefit for me.

Lgt2x (approved):
The change makes sense for me, we don't need 3 definitions for the exact same value. Are you ok with merging this @winterheart ?

winterheart:
I would retain `MAX_PATH` instead of `MAX_PATH_` macros as `MAX_PATH` is part of Windows SDK:

```cpp
#ifndef MAX_PATH
#define MAX_PATH 260
#endif
```

GravisZro:
> I would retain `MAX_PATH` instead of `MAX_PATH_` macros as `MAX_PATH` is part of Windows SDK:

I actually had questions about `MAX_PATH` versus `_MAX_PATH` and it turns out that `_MAX_PATH` is specific to their implementation of libc while `MAX_PATH` part of WinSDK. https://learn.microsoft.com/en-us/cpp/c-runtime-library/path-field-limits

This is an important distinction because theoretically `MAX_PATH` could diverge from `_MAX_PATH` which means it could be an invalid value for the libc implementation. As such, the wise move is to use the version that is associated with the functions we use, which in this case are libc functions. That is why `_MAX_PATH` is the correct choice in this particular matter.

winterheart:
There no MAX_PATH definition other than in Windows, so defining MAX_PATH on systems that is lacks of it (all other platforms except Windows) is most natural way.

GravisZro:
@winterheart 

Why define `MAX_PATH` at all? Why do we need a second definition for `_MAX_PATH`?

winterheart:
Because is how it defined in Win SDK. On other platforms we need compatibility define.

GravisZro:
> On other platforms we need compatibility define.

No, you do not because `MAX_PATH` is no longer used in the D3 code base.

Lgt2x:
Merging with @winterheart 's approval

PR #400, opened 2024-05-27
Remove unnecessary `find_package(OpenGL)`

Jayman2000:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

Before this change, `CMakeLists.txt` would run `find_package(OpenGL)`, but it would never use [any of the targets or variables that calling `find_package(OpenGL)` produces][1]. As a result, the call to `find_package(OpenGL)` didn’t really do anything.

[1]: https://cmake.org/cmake/help/latest/module/FindOpenGL.html

### Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -89,7 +89,6 @@ if(UNIX)
 
   find_package(SDL2 REQUIRED)
   find_package(Curses REQUIRED)
-  find_package(OpenGL REQUIRED)
   # Some versions of the SDL2 find_package set SDL2_INCLUDE_DIR and some set a plural SDL2_INCLUDE_DIRS. Check both.
   message("SDL2 Include Dir is ${SDL2_INCLUDE_DIR} ${SDL2_INCLUDE_DIRS}")
 endif()


winterheart:
Actually OpenGL functions get loaded with dlopen() in `dyna_gl.h`. On Linux OpenGL functions get loaded via `SDL_GL_GetProcAddress`. By checking OpenGL we ensuring that system has OpenGL and SDL2 can use functions from it.

Lgt2x:
> Actually OpenGL functions get loaded with dlopen() in `dyna_gl.h`. On Linux OpenGL functions get loaded via `SDL_GL_GetProcAddress`. By checking OpenGL we ensuring that system has OpenGL and SDL2 can use functions from it.

However this logic only applies on the machine building the game, not the one running it. Using `find_package` is only useful at compile time to link our executable against the targets found, which is not the case here. A runner compiling the game in theory should not need OpenGL at all.

Lgt2x:
> > Actually OpenGL functions get loaded with dlopen() in `dyna_gl.h`. On Linux OpenGL functions get loaded via `SDL_GL_GetProcAddress`. By checking OpenGL we ensuring that system has OpenGL and SDL2 can use functions from it.
> 
> However this logic only applies on the machine building the game, not the one running it. Using `find_package` is only useful at compile time to link our executable against the targets found, which is not the case here. A computer compiling the game in theory should not need OpenGL at all.

Lgt2x:
> Actually OpenGL functions get loaded with dlopen() in `dyna_gl.h`. On Linux OpenGL functions get loaded via `SDL_GL_GetProcAddress`. By checking OpenGL we ensuring that system has OpenGL and SDL2 can use functions from it.

However this logic only applies on the machine building the game, not the one running it. Using `find_package` is only useful at compile time to link our executable against the targets found, which is not the case here. A computer compiling the game in theory should not need OpenGL at all.

Jayman2000:
@winterheart

> Actually OpenGL functions get loaded with dlopen() in `dyna_gl.h`.

If we were linking the library at compile time, then I would agree that it would be good to give an error at compile time. However, if we’re using `dlopen()` to open the library at runtime, then I think think that we should only give an error at runtime. [We already give an error at runtime.](https://github.com/DescentDevelopers/Descent3/blob/599c1f759ee2d691f3fe948a8d6a81df18fedad8/renderer/dyna_gl.h#L226-L249)

---

@Lgt2x

> A computer compiling the game in theory should not need OpenGL at all.

Interestingly enough, there actually is a situation where computer compiling the game would need OpenGL. [`renderer/dyna_gl.h` says:](https://github.com/DescentDevelopers/Descent3/blob/599c1f759ee2d691f3fe948a8d6a81df18fedad8/renderer/dyna_gl.h#L21-L25)

```cpp
#if defined(WIN32)
#include <GL/gl.h>
#else
#include "SDL_opengl.h"
#endif
```

Windows is the only platform that has a direct dependency on OpenGL, but it’s also the only platform where we don’t run `find_package(OpenGL)`. (The line of code that I deleted was in an `if(UNIX)`). It’s OK to not call `find_package(OpenGL)` on Windows because we already depend on the Windows API, and [the Windows API has included `GL/gl.h` since Windows 2000](https://learn.microsoft.com/en-us/windows/win32/opengl/glclear).

Lgt2x (approved):
Thanks for the great explanation. Your change looks good to me.

Lgt2x (approved):

winterheart (approved):
Fine by me.

PR #403, opened 2024-05-28
Move global static variables out of headers

GravisZro:
## Pull Request Type

### Description
A few headers had some global static variables declared in their headers which is very bad. I also moved a function containing a static because it's only referenced in the implementation file.

### Checklist

### DIFF
--- a/ddebug/debug.h
+++ b/ddebug/debug.h
@@ -134,11 +134,11 @@
 #endif
 // #define DEBUG_LEVEL 0		//DAJ
 #if defined(WIN32) || defined(__LINUX__)
-static const int OSMBOX_OK = 1;
-static const int OSMBOX_YESNO = 2;
-static const int OSMBOX_YESNOCANCEL = 3;
-static const int OSMBOX_ABORTRETRYIGNORE = 4;
-static const int OSMBOX_OKCANCEL = 5;
+constexpr const int OSMBOX_OK = 1;
+constexpr const int OSMBOX_YESNO = 2;
+constexpr const int OSMBOX_YESNOCANCEL = 3;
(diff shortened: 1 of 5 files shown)


ddebug/debug.h
```diff
@@ -134,11 +134,11 @@
...
-static const int OSMBOX_YESNOCANCEL = 3;
-static const int OSMBOX_ABORTRETRYIGNORE = 4;
-static const int OSMBOX_OKCANCEL = 5;
+constexpr const int OSMBOX_OK = 1;
+constexpr const int OSMBOX_YESNO = 2;
+constexpr const int OSMBOX_YESNOCANCEL = 3;
+constexpr const int OSMBOX_ABORTRETRYIGNORE = 4;
+constexpr const int OSMBOX_OKCANCEL = 5;
 #else
 #define OSMBOX_OK 1
```
Lgt2x, line 143: should these definitions go away?

Lgt2x (approved):

ddebug/debug.h
GravisZro: That is not in the scope of this PR.

Lgt2x: I know, it was more of a general remark and an idea for a future cleanup

GravisZro: Oh, I see, the definition below. Yeah, they should go and the `constexpr` ones should be universal. Also there seems to be a duplicate sets that merely leave of "OS" from the beginning. I'll clean it up. Need to merge this first, though.

bryanperris (approved):

GravisZro:
Let the merging begin?

PR #410, opened 2024-05-31
Cleanup mono debugging code

winterheart:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

Remove unused mono debug code.

### Checklist


ddebug/debug.cpp
```diff
@@ -0,0 +1,55 @@
...
+#include <cstdlib>
+#include <cstring>
+
+#include "debug.h"
+
+void Debug_LogClose();
+
+FILE *Debug_logfile = nullptr;
+
+bool Debug_Logfile(const char *filename) {
```
Lgt2x, line 30: Should we merge this with our logging system set up at `misc/log.cpp` ?

Lgt2x: or did I misunderstand the purpose of this module

winterheart: Yes, it is, but our logging facility still not written, so I decided left it as is for a while.

Lgt2x: fair enough. How can I test this change?

winterheart: Log file is enabled by `-logfile` option. After starting logfile will be created in current directory under "d3.log" filename. Content will be identical to console output.

Lgt2x: file is created but empty on Linux, is that expected?

winterheart: -DLOGGER need to be enabled too as mprintf() is enabled only under  LOGGER define. This is works only in Debug build.

Lgt2x: yes my build is Debug and ENABLE_LOGGER is ON, investigating

Lgt2x: `Debug_ConsolePrintf` seems to call `Debug_LogWrite` only on Windows (winmono) and not on Linux (lnxmono)

Lgt2x: Confirming that it works when adding `  Debug_LogWrite(Mono_buffer);` in `Debug_ConsolePrintf` in `lnxmono.cpp`

winterheart: Seems this code was broken at first place. Added `Debug_LogWrite(Mono_buffer)` call in `Debug_ConsolePrintf` for Linux.

Lgt2x (approved):
Now working on Linux

PR #411, opened 2024-05-31
Moved multiple header files out of `lib`

GravisZro:
## Pull Request Type

### Description
I moved several include files out of the `lib` directory and back to their own modules. The `CMakeList.txt` files have been updated accordingly.

Some filenames were normalize to match their implementation or header counterparts.

### Related Issues
#406


GravisZro:
In my professional opinion we should be including `PROJECT_SOURCE_DIR` and using `<subproject/headerfile.h>` (use `<`/`>` to avoid relative file name collisions) for includes. This would eliminate the need for the obscene number of `target_include_directories` directives **AND** it would instantly inform you which subproject the header is part of.

GravisZro:
In my professional opinion we should be including `PROJECT_SOURCE_DIR` and using `"subproject/headerfile.h"` for includes.  Using a common point of origin (`PROJECT_SOURCE_DIR`) would eliminate the need for adding `target_include_directories` directives. As a bonus it makes it clear which subproject the header is being included from.

This PR uses the current method which is to include the directory of the header you need (via `target_include_directories`) and if it needs a header in a directory that isn't in `target_include_directories` then you recursively repeat the process until you have finally included all referenced directories.  This is a waste of time/effort that could easily be remedied if making this change was allowed.

Lgt2x:
> In my professional opinion we should be including `PROJECT_SOURCE_DIR` and using `"subproject/headerfile.h"` for includes. Using a common point of origin (`PROJECT_SOURCE_DIR`) would eliminate the need for adding `target_include_directories` directives. As a bonus it makes it clear which subproject the header is being included from.

My experience says otherwise :)
It's very tempting to include everything from the project root, but thinking at a "module" level, our approach makes more sense: a CMake-defined library declares public header files. If you want to include them in another module, you need to link against the module that declares them and make it a dependency. This makes module coupling more transparent, which is useful for instance for testing or moving modules around.

With your described approach, how do you know which subproject depends on which? You have to parse every file from the module and extract include prefixes.

2dlib/CMakeLists.txt
```diff
@@ -13,3 +13,12 @@ add_library(2dlib STATIC ${HEADERS} ${CPPS})
...
 )
+
+target_include_directories(2dlib PUBLIC
+  $<BUILD_INTERFACE:
+    ${PROJECT_SOURCE_DIR}/bitmap
+  >
+  $<BUILD_INTERFACE:
+    ${PROJECT_SOURCE_DIR}/renderer
+  >
+)
```
winterheart, line 24: This should be not target_include_directories, but target_link_libraries (with prevoius one):
```
target_link_libraries(2dlib PRIVATE
  bitmap
  cfile
  renderer
)
```

After that you should declare build interface of module itself

```
target_include_directories(2dlib PUBLIC
  $<BUILD_INTERFACE:
    ${PROJECT_SOURCE_DIR}/2dlib
  >
)
```

Same goes to all other target_include_directories changes. See cfile submodule layout as common example how it should looks like.

AudioEncode/CMakeLists.txt
```diff
@@ -1,6 +1,13 @@
-set(HEADERS)
+set(HEADERS
+  audio_encode.h)
```
winterheart, line 2: Don't add headers into CMake project, cmake should handle them as target dependency automatically.

winterheart (dismissed):
Changes too broad for project. Better try to update each submodule individually.

2dlib/CMakeLists.txt
GravisZro: > After that you should declare build interface of module itself

This is incorrect because it's translated to `-Idirectory` which is a search path.

AudioEncode/CMakeLists.txt
GravisZro: > Don't add headers into CMake project, cmake should handle them as target dependency automatically.

This is an incorrect use of `target_include_directories`. Each directory added to `target_include_directories` is a search path for `#include` directives which means nothing is actually added and therefore it will not show up in any IDE.

2dlib/CMakeLists.txt
GravisZro: > This should be not target_include_directories, but target_link_libraries (with prevoius one):

Doing so would result in overlapping/recursive linking dependencies.

GravisZro: > This should be not target_include_directories, but target_link_libraries (with prevoius one):

This creates an impossible situation because you are recursively linking. This infinite recursions ends on a seemingly arbitrary subproject. In this case, it tried to compile `AudioEncode` when I requested `linux` to be built.

``` sh
FAILED: AudioEncode/CMakeFiles/AudioEncode.dir/Debug/audio_encode.cpp.o 
/usr/bin/c++ -DCHECKSUM=2273889835UL -DLOGGER -DPRIMARY_HOG=\"d3-linux.hog\" -D_DEBUG -D_USE_OGL_ACTIVE_TEXTURES -D__LINUX__ -DCMAKE_INTDIR=\"Debug\" -I/home/gravis/project/Descent3/cfile -I/home/gravis/project/Descent3/ddebug -I/home/gravis/project/Descent3/fix -I/home/gravis/project/Descent3/lib -I/home/gravis/project/Descent3/linux -I/home/gravis/project/Descent3/Descent3 -I/home/gravis/project/Descent3/lib/linux -I/usr/include/SDL2 -g -fPIC -Wno-multichar -std=c++17 -MD -MT AudioEncode/CMakeFiles/AudioEncode.dir/Debug/audio_encode.cpp.o -MF AudioEncode/CMakeFiles/AudioEncode.dir/Debug/audio_encode.cpp.o.d -o AudioEncode/CMakeFiles/AudioEncode.dir/Debug/audio_encode.cpp.o -c /home/gravis/project/Descent3/AudioEncode/audio_encode.cpp
/home/gravis/project/Descent3/AudioEncode/audio_encode.cpp:25:10: fatal error: aencode.h: No such file or directory
   25 | #include "aencode.h"                                                                                                                                          
      |          ^~~~~~~~~~~                                                                                                                                                                  
compilation terminated.                   
```
`AudioEncode` needs `aencode.h` and here's the `CMakeLists.txt`

``` sh
set(HEADERS
  audio_encode.h)
set(CPPS
  audio_encode.cpp)

add_library(AudioEncode STATIC ${HEADERS} ${CPPS})
target_link_libraries(AudioEncode PUBLIC
  libacm
)
```
Notice that there is no `-I` for libacm in the first `g++` command despite being a linking requirement.  Just in case you needed to know, here' the directory:

``` sh
~/project/Descent3$ ls libacm/
adecode.cpp  adecode.h  aencode.cpp  aencode.h  CMakeLists.txt  decode.c  libacm.h
```

**I cannot do what you ask because it will not compile.**

GravisZro:
@Lgt2x 
Please step in here because @winterheart doesn't seem to understand what these cmake directives actually do.

2dlib/CMakeLists.txt
winterheart: You didn't `target_include_directory` on submodule. Again **see cfile for the reference**.

```
set(CPPS
  encoder.cpp
)

add_library(AudioEncode STATIC ${CPPS})
target_link_libraries(AudioEncode libacm)
target_include_directories(AudioEncode PUBLIC
  $<BUILD_INTERFACE:
  ${PROJECT_SOURCE_DIR}/AudioEncode
  >
)
```

```

  Checking Build System
  Building Custom Rule C:/Users/azama/CLionProjects/Descent3/libacm/CMakeLists.txt
  aencode.cpp
  adecode.cpp
  decode.c
  libacm.vcxproj -> C:\Users\azama\CLionProjects\Descent3\builds\win32\libacm\Debug\libacm.lib
  Building Custom Rule C:/Users/azama/CLionProjects/Descent3/AudioEncode/CMakeLists.txt
  encoder.cpp
  AudioEncode.vcxproj -> C:\Users\azama\CLionProjects\Descent3\builds\win32\AudioEncode\Debug\AudioEncode.lib

Build finished
```

GravisZro: I can't believe it worked.

AudioEncode/CMakeLists.txt
GravisZro: @winterheart Fixing the situation with IDEs (especially my own) is the entire purpose of doing this. I don't know why you are a so dead set against listing the header files but this is something I require. I have done everything else you have asked, please remove your block on this merge.

JeodC: Try here is a mix of cmakelists with header files set this way already. I fail to see the harm in consistency.

SiriusTR: Depends on the IDE and how you use it. e.g. I believe this is true if you use CMake to spit out a .vcproj, but if you open the folder directly in VS (Ninja etc) it finds the headers on its own.

GravisZro: > @JeodC: There is a mix of cmakelists with header files set this way already.

Good point.

> @SiriusTR: Depends on the IDE and how you use it.

An IDE **should not** display include directories specifically because it could be a system directory which could easily have hundreds of header files. They may go to great lengths to try and only include what it believes is not a system directory but it's impossible to know and based on assumptions about your system.  The proper way is to display files explicitly indicated as being part of the project. I see no reason to oppose proper header file inclusion.

bryanperris: Trying to follow this conversation here, so bare with me. First off, we shouldn't care about what IDE is used, when this project can be edited in something simple as VI and built just from command line, so we only care about the cmake structure here.  I am used to cmake projects with cmake submodules, and in those submodules, their have their own set of header files, and perhaps even their own `include` directory too.  I think when one submodule wants to include another, just needs to add that cmake defined submodule into their dep list.

GravisZro: > we shouldn't care about what IDE is used, when this project can be edited in something simple as VI and built just from command line,

I'm not modifying it for a specific IDE just _an_ IDE, any IDE. While not technically required for the build process, it also does not harm the build process. We should be encouraging people to work on the project rather than discouraging it. IDEs are tools to help developers so going out of your way to make it more difficult for developers to work on your project is antithetical to being open source in the first place.

If you look at the PR, the only headers I've listed are the ones in the same directory that are used by their implementation file counterparts. I have not listed headers from other submodules, only the current one.

winterheart: @GravisZro we requested changes to your contribution. If you disagree with them without a good enough reason, then PR is probably not going to be merged. Enumerating headers in cmake project is undesirable practice that prevents to detect unused headers on some IDEs (for example).

Earlier I suggested update your particular QtCreator to recent version, where your issue does not exists.

GravisZro: @winterheart 
> IDEs are tools to help developers so going out of your way to make it more difficult for developers to work on your project is antithetical to being open source in the first place.

How is this not a good enough reason to keep the header files listed? Why are you being so uncompromising?

bryanperris: > > we shouldn't care about what IDE is used, when this project can be edited in something simple as VI and built just from command line,
> 
> I'm not modifying it for a specific IDE just _an_ IDE, any IDE. While not technically required for the build process, it also does not harm the build process. We should be encouraging people to work on the project rather than discouraging it. IDEs are tools to help developers so going out of your way to make it more difficult for developers to work on your project is antithetical to being open source in the first place.
> 
> If you look at the PR, the only headers I've listed are the ones in the same directory that are used by their implementation file counterparts. I have not listed headers from other submodules, only the current one.

As I said, we will just follow the best cmake standards, where it is expected some IDE with cmake support will also follow the same suit.  Most of the cmake based C++ projects on GitHub just follow that basic standard, so the source can be built with just CLI as a minimal.  People encouraged working on Descent 3 will be the ones that love both the game and C++.

GravisZro:
@Lgt2x I've done everything I can to appease @winterheart but he still insists I remove the listed header files, knowing that this will harm my ability to work on the project. Can you please do something? What happened to compromise?

AudioEncode/CMakeLists.txt
GravisZro: > where it is expected some IDE with cmake support will also follow the same suit.

Except cmake is a build system, not a project management system. If there were a project file then I would use that but there is only cmake files.

> Most of the cmake based C++ projects on GitHub just follow that basic standard, so the source can be built with just CLI as a minimal. 

Literally nothing I'm proposing will interfere with cmake building D3.

The question is, why is this small compromise unacceptable?

bryanperris: I didn't say the changes were unacceptable, still need re-review them again.  I wanted to kick off a cmake discussion, to get a feel with everyone on best direction to go.  When everyone is on the same page, I hope that PRs aren't delayed by any further arguments.

Cmake is a build system, but it has some project generators, then there are IDEs with the ability to establish the project structure based on cmake info.

Lgt2x: I did not step in before because I thought things would sort themselves out.

Most projects I've seen list headers in the source files, especially for IDE support, it's not a big pain to do it. You add them in the target sources like you do for cpps. CMake can do without, but it's not wrong at all to do it. Everything I found online about this issue corroborates this [1](https://stackoverflow.com/questions/13703647/how-to-properly-add-include-directories-with-cmake) [2](https://stackoverflow.com/questions/61057405/cmake-add-header-files-to-target). Craig Scott (author of the main CMake book) suggests [file sets](https://discourse.cmake.org/t/target-include-directories-vs-headers-in-target-sources/2625/2) as an alternative, that's something we can look into later.

tl;dr I think this change is totally acceptable.

GravisZro:
@winterheart @Lgt2x 
Hey, what's the deal?

Lgt2x:
> @winterheart @Lgt2x Hey, what's the deal?

he's not available this week, I'll take a closer look at the PR whenever I get the chance

libacm/CMakeLists.txt
```diff
@@ -9,3 +12,9 @@ set(LIB_SRC
   libacm.h)
 
 add_library(libacm STATIC ${HEADERS} ${CPPS} ${LIB_SRC})
+
+target_include_directories(libacm PUBLIC
+  $<BUILD_INTERFACE:
+    ${PROJECT_SOURCE_DIR}/libacm
```
Lgt2x, line 18: ideally this module should only expose `adecode.h/aencode.h` and not `libacm.h` from upstream libacm, but that'll require moving it to a third party (idea for a future improvement)

linux/CMakeLists.txt
```diff
@@ -1,3 +1,10 @@
+set(HEADERS
+  linux_fix.h
+  lnxapp.h
+  lnxcontroller.h
+  registry.h
```
Lgt2x, line 5: Missing headers from this module

netcon/inetfile/CMakeLists.txt
```diff
@@ -1,8 +1,17 @@
...
   Chttpget.cpp
   inetgetfile.cpp)
 
 add_library(inetfile STATIC ${HEADERS} ${CPPS})
+
+target_link_libraries(inetfile PRIVATE
+  networking
+)
+
+target_include_directories(inetfile PUBLIC .)
```
Lgt2x, line 17: `${PROJECT_SOURCE_DIR}/...`

netcon/lanclient/CMakeLists.txt
```diff
@@ -1,17 +1,30 @@
...
+  module
+  manage
+  renderer
+  2dlib
   $<$<PLATFORM_ID:Windows>:
     ws2_32
   >
 )
 
+target_include_directories(Direct_TCP_IP PUBLIC .)
```
Lgt2x, line 26: `${PROJECT_SOURCE_DIR}/...`

renderer/CMakeLists.txt
```diff
@@ -15,6 +18,7 @@ set(CPPS
 )
 
 if(UNIX)
+  set(HEADERS ${HEADERS} lnxscreenmode.h)
```
Lgt2x, line 21: use `list(APPEND`

renderer/CMakeLists.txt
```diff
@@ -1,6 +1,9 @@
 set(HEADERS
   dyna_gl.h
   HardwareInternal.h
+  rend_d3d.h
+  renderer.h
+  rend_opengl.h
```
Lgt2x, line 6: `3d.h` missing

ui/CMakeLists.txt
```diff
@@ -19,4 +25,11 @@ set(CPPS
...
 target_link_libraries(ui PRIVATE
   ddio
+  bitmap
+  grtext
+  renderer
+  dd_video
+  2dlib
 )
+
+target_include_directories(ui PUBLIC .)
```
Lgt2x, line 35: `${PROJECT_SOURCE_DIR}/...` (next ones too)

Lgt2x (changes requested):
First pass, some small adjustments to make, but good job otherwise. I'd still like a second pair of eyes to review this thoroughly, changes to the build system are always scary

dd_grwin32/CMakeLists.txt
```diff
@@ -9,3 +9,17 @@ set(CPPS
...
+  $<$<PLATFORM_ID:Linux,Darwin>:
+    linux
+  >
+  $<$<PLATFORM_ID:Windows>:
+    win32
+  >
+)
+
+
+target_include_directories(dd_grwin32 PUBLIC .)
```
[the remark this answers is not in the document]
GravisZro: What about using `${CMAKE_CURRENT_SOURCE_DIR}` instead?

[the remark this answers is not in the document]
Lgt2x: that works yeah

GravisZro:
> First pass, some small adjustments to make, but good job otherwise. I'd still like a second pair of eyes to review this thoroughly, changes to the build system are always scary

The requested changes have been made and the branch is now up to date.

Lgt2x (commented):
The code looks good to me now, it just needs a trivial rebase on master to include #437. However I have a crash on startup on Linux, on the line `CRegistry *oeLnxAppDatabase::GetSystemRegistry() { return database; }` / program.cpp:72

GravisZro:
@Lgt2x After investigating the issue, I'm not seeing how the linux version managed to work before. The static `Database` pointer value is initialized as `NULL` and `dbase` uses that as a pointer to a parent. It seem like there was some dependency on static initialization order and as a result of changing the order, it would fail. Anyway, it now longer uses `Database` to init `dbase` which works just fine on Linux.  I would make another PR but it's not a big change and it would just cause more conflicts.

Lgt2x (approved):

Lgt2x:
That's one reason why static variables suck, and D3 is full of them :(
Ok on Linux, I'll test on Windows and merge if nobody opposes

Lgt2x:
@GravisZro could you run it on Win32 properly? I get an error "Unable to find version key for Descent 3" error on startup, maybe I messed something up

GravisZro:
@Lgt2x I got that error too when I tried it (wine). However, I think I've fully corrected the situation (no error box on start) but I need to you to do a full test on actual Windows.

Lgt2x:
The error seems to persist

Lgt2x:
Do you need help debugging Windows build?

Lgt2x:
Do we absolutely need this last commit to fix the windows build? This PR is already big enough, it'd be better to keep modifications minimal

GravisZro:
> Do we absolutely need this last commit to fix the windows build? This PR is already big enough, it'd be better to keep modifications minimal

I'll be splitting this into another PR that will replace some static variable with functions that self-initialize. This will fix the linking order problem and simplify this PR.

Lgt2x:
outdated, mostly covered by other Pull Requests. CMakeLists now declare headers properly

Issue #279, opened 2024-05-03
[Runtime Issue]: Demo file not supported


donotsharepersonalinformation:
### Build Version

0e55dee0dec22b2895f14beb0c789bbbd489d5c3

### Operating System Environment

- [ ] Microsoft Windows (32-bit)
- [ ] Microsoft Windows (64-bit)
- [ ] Mac OS X
- [X] Linux (specify distribution and version below)

### CPU Environment

- [ ] x86 (32-bit Intel/AMD)
- [X] x86\_64 (64-bit Intel/AMD)
- [ ] ARM (32-bit)
- [ ] ARM64 (64-bit; sometimes called AArch64)
- [ ] Other (RISC V, PPC...)

### Game Modes Affected

- [ ] Single player
- [ ] Anarchy
- [ ] Hyper-Anarchy
- [ ] Robo-Anarchy
- [ ] Team Anarchy
- [ ] Capture the Flag
- [ ] Bounty
- [ ] Entropy
- [ ] Hoard
- [ ] Monsterball
- [ ] Cooperative

### Description

With a fresh CD install I was clicking different menus. I noticed in the View Demo menu I have a file that seems to ship with the game called Secret2.dem. When trying to run that file it will say that "boa not valid" and that the demo file is invalid. While not a show stopper I am not sure to the level of backward compatibility for such files are wanted. 

On a side note the line that shows the full path will have the last character cut off.

### Regression Status

I don't know.

### Steps to Reproduce

1. Install game from CD. (Not sure if file is included on digital releases).
2. Go to View Demo menu.
3. Load Secret2.dem (I also don't have access to other demo files so I'm not sure if others will fail).

JeodC:
Duplicate of #105

Jayman2000:
> When trying to run that file it will say that "boa not valid" and that the demo file is invalid.

What happens after that? Does the demo playback normally?

donotsharepersonalinformation:
> > When trying to run that file it will say that "boa not valid" and that the demo file is invalid.
> 
> What happens after that? Does the demo playback normally?

It will give an error "Bad demo file, unable to read" and then boot you back to the main menu.

JeodC:
I tested this with the CD version and it loaded fine. I am unable to reproduce the issue on Windows. @donotsharepersonalinformation do you mind providing more info?

- CPU
- Which CD you used (presumably Linux Loki x86)?

donotsharepersonalinformation:
[...]

PR #412, opened 2024-06-01
Modernize: Move some arrays to std::array

GravisZro:
## Pull Request Type
- [x] Runtime changes
  - [x] Other changes

### Description
The ultimate goal is to switch several variables to vectors in order to expand game limitations but the first step is to switch C arrays to `std::array` so that we ensure that memory bounds are respected. There is more in-depth work to adapting how the arrays are used but I didn't want to make the PR too large.

Other stuff
- The `SAVE_DATA_TABLE` was replaced with a template function version
- The `OBJNUM` macro was used in place of `obj - Objects`
- Three bugs found and fixed.
  - In `ObjInitGeneric` in for loop where `obj_info` wasn't being used.
  - Twice in WeaponFire.cpp where `fq.thisobjnum` was being assigned `Objects - obj` instead of `obj - Objects` (replaced with `OBJNUM` macro)

### Checklist


winterheart:
I don't see value of these changes. If you trying to solve particular issue, focus on particular part of code. General rule of thumb: "Don't fix that aren't broken".

GravisZro:
> I don't see value of these changes.

That's your mistake because there are currently significant issues open that are likely due to memory access issues.

winterheart:
> > I don't see value of these changes.
> 
> That's your mistake because there are currently significant issues open that are likely due to memory access issues.

Prove it, please. Do you have checked code with Valgrind, is there memory issues on changed code? Do all changed in this PR code suffer these issues?

GravisZro:
The purpose of the changes are to ensure that if a memory violation occurs that it will immediately be caught rather than continue, possibly crashing at a later point. I don't understand why you are being so oppositional to improving the code's reliability. Also, I did point out that I found and fixed some bugs while doing this.

GravisZro:
@winterheart 
Do you not want bugfixes? This contains bugfixes.

Descent3/AImain.cpp
```diff
@@ -1625,7 +1625,7 @@ const char *const Ai_movement_subtype_walking_strings[MAX_AI_INIT_MOVEMENT_SUBTY
 #define AI_MAX_MELEE_RANGE 5.0f
 
 int AI_NumRendered;
-int AI_RenderedList[MAX_OBJECTS];
+std::array<int, MAX_OBJECTS> AI_RenderedList;
```
tophyr, line 1628: As long as we're doing these conversions, let's default-initialize the elements in the arrays:
```suggestion
std::array<int, MAX_OBJECTS> AI_RenderedList{};
```
Both `T[N]` and `std::array<T, N>` default-initialize only the _array object_, not the array _elements_. IE, this gives an indeterminate value:
```
std::array<int, 4> arr;
int val = arr[0];
// val is indeterminate
```

Descent3/Game2DLL.cpp
```diff
@@ -66,15 +66,15 @@
...
+extern std::array<int, NUM_CAMERA_VIEWS> Camera_view_mode;
 
 //	Osiris_CreateModuleInitStruct
 //	Purpose:
 //		This function initializes a Module Init Struct with all the needed data to get sent
 //	to the module during initialization.
 extern void Osiris_CreateModuleInitStruct(tOSIRISModuleInit *mi);
 module GameDLLHandle = {NULL};
-extern ddgr_color Player_colors[];
+extern std::array<ddgr_color, 32> Player_colors;
```
tophyr, line 77: is there a constant that can be used here instead of `32`?

Descent3/Game2DLL.cpp
```diff
@@ -136,16 +136,16 @@ static void DUMMYrend_DrawScaledBitmap(int x1, int y1, int x2, int y2, int bm, f
 }
 
 void GetGameAPI(game_api *api) {
-  api->objs = (int *)Objects;
-  api->rooms = (int *)Rooms;
+  api->objs = (int *)std::data(Objects);
```
tophyr, line 139: don't like that this is still c-casting to `int*` - this is no safer than before. can the `game_api::...` fields be converted to just be correctly-typed?

tophyr: also, why `std::data` on some of the references but not others?

Descent3/ObjInit.cpp
```diff
@@ -734,8 +734,8 @@ int ObjInitGeneric(object *objp, bool reinit) {
   // Deal with deleted type
   if (obj_info->type == OBJ_NONE) {
     int i;
-    for (i = 0, obj_info = Object_info; i < MAX_OBJECT_IDS; i++, obj_info) { // find other object of same type
-      if (Object_info[i].type == objp->type)
+    for (i = 0, obj_info = std::data(Object_info); i < MAX_OBJECT_IDS; i++, obj_info++) { // find other object of same type
```
tophyr, line 737: could this just be an iterator loop?
```suggestion
    for (auto const& obj_info : Object_info) { // find other object of same type
```

Descent3/WeaponFire.cpp
```diff
@@ -1183,7 +1183,7 @@ void AquireElectricalTarget(object *obj) {
     fq.startroom = obj->roomnum;
     fq.p1 = &dest;
     fq.rad = .0001f;
-    fq.thisobjnum = Objects - obj;
+    fq.thisobjnum = OBJNUM(obj); // bugfix?
```
tophyr, line 1186: unclear what this comment means - this doesn't look markedly different from the other `OBJNUM` pattern usages. what's the bug?

Descent3/aistruct.h
```diff
@@ -713,17 +715,17 @@ struct ain_weapon_hit_info {
 
 #define AI_MEM_DEPTH 5
 
-struct ai_mem {
+struct [[gnu::packed]] ai_mem {
```
tophyr, line 718: why is struct packing needed? without context, this seems smelly

Descent3/aistruct.h
```diff
@@ -713,17 +715,17 @@ struct ain_weapon_hit_info {
...
   // Computed at end of memory frame
-  float shields;
-  int16_t num_enemies;
-  int16_t num_friends;
+  float shields = 0.0f;
```
tophyr, line 720: <3 defaults

Descent3/fireball.cpp
```diff
@@ -1091,7 +1092,7 @@ void CreateSplintersFromBody(object *obj, float explosion_mag, float lifetime) {
     dest = center + obj->pos;
 
     // Now create splinter with that faces center position
-    int s_objnum = ObjCreate(OBJ_SPLINTER, pm - Poly_models, obj->roomnum, &dest, NULL);
+    int s_objnum = ObjCreate(OBJ_SPLINTER, pm - std::data(Poly_models), obj->roomnum, &dest, NULL);
```
tophyr, line 1095: possible to use similar pattern as `OBJNUM`?

Descent3/gamesave.cpp
```diff
@@ -826,16 +826,17 @@ bool SaveGameState(const char *pathname, const char *description) {
...
-    highest_index = -1;                                                                                                \
-    for (i = 0; i < (_nitems); i++)                                                                                    \
-      if (_array[i].used)                                                                                              \
-        highest_index = i;                                                                                             \
-    gs_WriteShort(fp, highest_index + 1);                                                                              \
-    for (i = 0; i <= highest_index; i++)                                                                               \
-      cf_WriteString(fp, _array[i].used ? _array[i].name : "");                                                        \
-  } while (0)
+template <typename T, size_t sz>
+void save_data_table(CFILE *fp, const std::array<T, sz>& data) {
```
tophyr, line 830: image

Descent3/multi.h
```diff
@@ -784,9 +787,7 @@ extern int Num_network_games_known;
 extern int Game_is_master_tracker_game;
 
 #define TRACKER_ID_LEN 10 // Don't change this!
-extern char Tracker_id[TRACKER_ID_LEN];
-
-extern uint16_t Turrett_position_counter[MAX_OBJECTS];
```
tophyr, line 789: did `Turrett_position_counter` move or was it simply unused externally?

scripts/LEVEL15.cpp
```diff
@@ -842,28 +844,38 @@ struct tMyHandle {
   int handle;
 };
 
-tMyHandle MyObjects[] = {
+std::array<tMyHandle, 18> MyObjects = {
+    tMyHandle
```
tophyr, line 848: this initializer has 19 items for an 18-element array. what is `tMyHandle`?

scripts/LEVEL15.cpp
```diff
@@ -842,28 +844,38 @@ struct tMyHandle {
...
     {"MatCen1Peg", -1},        {"MatCen2Peg", -1},    {"MatCen3Peg", -1},    {"MatCen4Peg", -1},
     {"MatCen5Peg", -1},        {"MatCenSwitchA", -1}, {"MatCenSwitchB", -1}, {"MatCenSwitchC", -1},
     {"MatCenSwitchD", -1},     {"MatCenSwitchE", -1}, {"MatCenSwitchF", -1}, {"MatCenSwitchG", -1},
     {"MatCenPuzzleRESET", -1}, {"MatCen1Spewer", -1}, {"MatCen2Spewer", -1}, {"MatCen3Spewer", -1},
     {"MatCen4Spewer", -1},     {"MatCen5Spewer", -1},
 };
 
-tMyHandle MyMatcens[] = {{"MatCen5", -1}, {"MatCen4", -1}, {"MatCen2", -1}, {"MatCen3", -1}, {"MatCen1", -1}};
+std::array<tMyHandle, 5> MyMatcens = {
+    tMyHandle
```
tophyr, line 857: 6 items

scripts/LEVEL15.cpp
```diff
@@ -842,28 +844,38 @@ struct tMyHandle {
...
   const char *mem;
 };
 
-tMyMessage MyMessages[] = {{"MagicMatCenSwitches", NULL}, {"MatCenSwitchDOn", NULL},  {"MatCenSwitchDOff", NULL},
-                           {"MatCenSwitchGOn", NULL},     {"MatCenSwitchGOff", NULL}, {"MatCenSwitchCOn", NULL},
-                           {"MatCenSwitchCOff", NULL},    {"MatCenSwitchEOn", NULL},  {"MatCenSwitchEOff", NULL},
-                           {"MatCenSwitchBOn", NULL},     {"MatCenSwitchBOff", NULL}, {"MatCenSwitchFOn", NULL},
-                           {"MatCenSwitchFOff", NULL},    {"MatCenSwitchAOn", NULL},  {"MatCenSwitchAOff", NULL}};
+std::array<tMyMessage,16> MyMessages = {
+        tMyMessage
```
tophyr, line 872: 17 items

tophyr (changes requested):
noticed what i think is a problem in `scripts/LEVEL15.cpp` - added array initialization objects. maybe some find-replace artifacts?

don't love the added `[[gnu::packed]]` attribute in `Descent3/aistruct.h` either.

beyond the above, added some suggestions but this looks largely fine to me.

tophyr:
> The purpose of the changes are to ensure that if a memory violation occurs that it will immediately be caught rather than continue

@GravisZro for what it's worth, I didn't notice any changes here (other than the use of `std::array::fill()` over `memset()`) that actually add bounds-checking. `std::array` doesn't perform bounds checking via its iterators or `operator[]`. I do think that we should move to `std::array`; I just want to call out that the type itself adds no inherent safety. The additional safety will come when we transition to iterator for loops (`for (auto const& foo : foos)` etc) and stdlib algorithms.

Descent3/WeaponFire.cpp
tophyr: ah i see the reversed `Objects - obj` instead of `obj - Objects`

Descent3/aistruct.h
pzychotic: Also [[gnu::packed]] isn't portable, MSVC will just ignore the unknown attribute with [warning C5030: attribute [[gnu::packed]] is not recognized](https://github.com/DescentDevelopers/Descent3/pull/412/checks#step:6:520).
I believe `#pragma pack` is supported by all recent compiler versions nowadays.

Descent3/Game2DLL.cpp
GravisZro: I'm afraid not. It's origin is from [Descent3/renderobject.cpp, line 1479](https://github.com/DescentDevelopers/Descent3/blob/main/Descent3/renderobject.cpp#L1479) which actually doesn't even specify the array a size.

Descent3/Game2DLL.cpp
GravisZro: 1) I don't know why it's casted at all, so I'm simply not breaking the code.
2) The ones with `std::data` are array `std::array` types while the other are pointers.

Descent3/ObjInit.cpp
GravisZro: As I stated at the beginning of the PR...
> There is more in-depth work to adapting how the arrays are used but I didn't want to make the PR too large.

I'm 100% in agreement with the suggestion but I didn't want to make a PR that required too much analysis.

Descent3/fireball.cpp
GravisZro: This is merely adapting the code to function. Anyway, the OBJNUM pattern is too hacky anyway. Moving to an iterator based approach would be better. I'm just trying to get the process started.

Descent3/multi.h
GravisZro: `Turrett_position_counter` is not defined or even referenced anywhere in the code. Removal seemed like the right move.

scripts/LEVEL15.cpp
GravisZro: No, there are only 18 items because `tMyHandle` is the type for item 0. This hint allows the compiler to know that the others are also type `tMyHandle`.

GravisZro:
> `std::array` doesn't perform bounds checking via its iterators or `operator[]`

The C++ standard doesn't not specify the behavior of accessing nonexistent elements using `operator[]`. However, common implementations like [GCC will in fact do bounds checking.](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:25,endLineNumber:6,positionColumn:25,positionLineNumber:6,selectionStartColumn:25,selectionStartLineNumber:6,startColumn:25,startLineNumber:6),source:'%23include+%3Carray%3E%0A%0Aint+main()%0A%7B%0A++++std::array%3Cint,+3%3E+arr%3B%0A++++for(int+i+%3D+0%3B+i+%3C+5%3B+i%2B%2B)%0A++++%7B%0A++++++arr%5Bi%5D+%3D+0%3B%0A++++%7D%0A++++return+0%3B%0A%7D'),l:'5',n:'1',o:'C%2B%2B+source+%231',t:'0')),k:50.42378344047672,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((g:!((h:compiler,i:(compiler:gsnapshot,filters:(b:'0',binary:'0',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'1',execute:'0',intel:'1',libraryCode:'0',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!(),options:'',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+x86-64+gcc+(trunk)+(Editor+%231)',t:'0')),k:49.57621655952329,l:'4',m:62.73339179622393,n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'x86-64+gcc+(trunk)',editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+x86-64+gcc+(trunk)+(Compiler+%231)',t:'0')),header:(),l:'4',m:37.26660820377608,n:'0',o:'',s:0,t:'0')),k:49.57621655952329,l:'3',n:'0',o:'',t:'0')),l:'2',n:'0',o:'',t:'0')),version:4)

Descent3/aistruct.h
GravisZro: @tophyr this was part of [something else I was working on. ](https://github.com/DescentDevelopers/Descent3/issues/279)
@pzychotic MSVC is perpetually behind both GCC and clang. You would be better off using clang/llvm on windows.

Descent3/Game2DLL.cpp
GravisZro: I just realized that the size is actually the maximum number of players, so `MAX_NET_PLAYERS` it is.

Descent3/AImain.cpp
GravisZro: I mean... OK but I was expecting most of these to eventually be changed to `std::vector`.

tophyr: That would be preferable, definitely.

scripts/LEVEL15.cpp
tophyr: OH, i see, i wasn't visually parsing this correctly due to the newline. This is actually
```
std::array<tMyHandle, 18> MyObjects = { tMyHandle{"MatCen1Peg", -1}, {"MatCen2Peg", -1}, ... };
```
ok. that's great then, i'm fine with this.

tophyr (approved):
This is all great, modulo the `[[gnu::packed]]` annotation.

Descent3/aistruct.h
pzychotic: It isn't about what I use, it's about what this project does. I linked to the current CI build log, because we do use MSVC here and that attribute is being ignored in our windows builds. Unless this projects decides to be GCC/Clang only, this issue remains.
Also don't try to tell me which build environment I have to use, just because you don't like it.

GravisZro: not really a problem because it's being removed.

jengelh:
> > `std::array` doesn't perform bounds checking via its iterators or `operator[]`
> 
> The C++ standard doesn't not specify the behavior of accessing nonexistent elements using `operator[]`. However, common implementations like GCC's libstdc++ will do bounds checking.

But also only in specific debug builds (`-D_GLIBCXX_DEBUG=1`). For guaranteed bounds checking, there has always been `at()`.

PR #413, opened 2024-06-01
Revamp build instructions

Jayman2000:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

Before I made this PR, I had tried building Descent 3 on Windows for the first time. The instructions for building Descent 3 on Windows were unclear, inaccurate and they messed up my installation of Windows. I started working on this PR in order to prevent that situation from happening again. I ended up revamping the entire build instructions section.

### Related Issues

Resolves problems from #332.

### Checklist

### Additional Comments

I have tested this PR on the following:

- Fedora 40 64-bit
- Ubuntu 24.04 64-bit
- Windows 11
- Windows 10 64-bit

[I tried to test it on macOS, but I couldn’t get macOS to install.](https://github.com/quickemu-project/quickemu/discussions/1262) If someone else could test this out on macOS, that would be greatly appreciated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,47 +92,82 @@ The milestone needs testing on all platforms. Please report issues when found.
     - D3 Open Source compiles level scripts in their own hogfiles. Make sure you copy and overwrite `d3-{platform}.hog`.
 
 ## Building
-Build steps below assume you have already cloned the repository and entered it locally.
-
 #### Building - Windows
-Requires Visual Studio 2022 and C++ Tools (cmake and vcpkg)
+1. Make sure that you have Git and Visual Studio 2022 with the “Desktop development with C++” workload. If you don’t already have those installed or you aren’t sure, then open an elevated Command Prompt and run:
 
-Install and configure vcpkg:
-```sh
-git clone https://github.com/microsoft/vcpkg.git
-cd vcpkg && bootstrap-vcpkg.bat
-setx VCPKG_ROOT="C:\path\to\vcpkg"
-setx PATH=%VCPKG_ROOT%;%PATH%
-```
+    <!--
+    The following code block specifies the full path to the Visual Studio Installer because the Visual Studio Installer doesn’t add itself to the user’s Path. The installer is guaranteed to be in a specific location on 64-bit systems [1]. The installer will be in a different location on 32-bit systems [2], but Visual Studio 2022 doesn’t support 32-bit systems [3] so we can ignore that detail.
 
-Build Descent 3 using the "x86 native tools command prompt"
-```sh
-cmake --preset win32 -D ENABLE_LOGGER=[ON|OFF]
-cmake --build --preset win32 --config [Debug|Release]
-```
+    [1]: <https://learn.microsoft.com/en-us/visualstudio/install/use-command-line-parameters-to-install-visual-studio?view=vs-2022>
+    [2]: <https://github.com/microsoft/vswhere/wiki#installing>
+    [3]: <https://learn.microsoft.com/en-us/answers/questions/1689898/does-visual-studio-build-tools-2022-support-32-bit>
+    -->
 
-#### Building - MacOS
-```sh
-brew bundle install
-cmake --preset mac -D ENABLE_LOGGER=[ON|OFF]
-cmake --build --preset mac --config [Debug|Release]
-```
+    ```batch
+    winget install Git.Git Microsoft.VisualStudio.2022.Community
+
+    "%ProgramFiles(x86)%\Microsoft Visual Studio\Installer\setup.exe" modify^
+        --passive^
+        --channelId VisualStudio.17.Release^
+        --productId Microsoft.VisualStudio.Product.Community^
+        --add Microsoft.VisualStudio.Workload.NativeDesktop;includeRecommended
+    ```
+
+2. Open a “x86 Native Tools Command Prompt” and run:
+
+    ```batch
+    git clone https://github.com/DescentDevelopers/Descent3
+    cd Descent3
+    cmake --preset win32 -D ENABLE_LOGGER=[ON|OFF]
+    cmake --build --preset win32 --config [Debug|Release]
+    ```
+
+Once CMake finishes, the built files will be put in `builds\win32\Descent3\Debug` or `builds\win32\Descent3\Release`.
+
+#### Building - macOS
+1. Make sure that [Xcode](https://developer.apple.com/xcode) is installed.
+
+2. Make sure that [Homebrew](https://brew.sh) is installed.
+
+3. Run these commands:
+
+    ```sh
+    git clone https://github.com/DescentDevelopers/Descent3
+    cd Descent3
+    brew bundle install
+    cmake --preset mac -D ENABLE_LOGGER=[ON|OFF]
+    cmake --build --preset mac --config [Debug|Release]
+    ```
+
+Once CMake finishes, the built files will be put in `builds/mac/Descent3/Debug` or `builds/mac/Descent3/Release`.
 
 #### Building - Linux (Ubuntu)
+Run these commands:
+
 ```sh
 sudo apt update
-sudo apt install -y --no-install-recommends ninja-build cmake g++ libsdl2-dev zlib1g-dev
+sudo apt install -y --no-install-recommends git ninja-build cmake g++ libsdl2-dev zlib1g-dev
+git clone https://github.com/DescentDevelopers/Descent3
+cd Descent3
 cmake --preset linux -D ENABLE_LOGGER=[ON|OFF]
 cmake --build --preset linux --config [Debug|Release]
 ```
 
+Once CMake finishes, the built files will be put in `builds/linux/Descent3/Debug` or `builds/linux/Descent3/Release`.
+
 #### Building - Linux (Fedora)
+Run these commands:
+
 ```sh
 sudo dnf update --refresh
-sudo dnf install -y ninja-build cmake gcc-c++ SDL2-devel zlib-devel
+sudo dnf install -y git ninja-build cmake gcc-c++ SDL2-devel zlib-devel
+git clone https://github.com/DescentDevelopers/Descent3
+cd Descent3
 cmake --preset linux -D ENABLE_LOGGER=[ON|OFF]
 cmake --build --preset linux --config [Debug|Release]
 ```
+
+Once CMake finishes, the built files will be put in `builds/linux/Descent3/Debug` or `builds/linux/Descent3/Release`.
 
 ## Contributing
 Anyone can contribute! We have an active Discord presence at [Descent Developer Network](https://discord.gg/GNy5CUQ). If you are interested in maintaining the project on a regular basis, please contact Kevin Bentley.


JeodC:
At first glance this appears wholly unnecessary, as building in windows using visual studio, cmake and vcpkg is nowhere near challenging. Can you describe what happened to prompt this in more detail?

README.md
```diff
@@ -92,47 +92,128 @@ The milestone needs testing on all platforms. Please report issues when found.
...
+    ```sh
+    cd Descent3
+    ```
+
+5. Install the rest of the prerequisites:
+
+    - **Windows:**
+
+        ```pwsh
+        powershell -ExecutionPolicy Bypass -File Install-WindowsDependencies.ps1
```
Lgt2x, line 180: I feel like this makes the process look more complicated than what it really is. For anyone that has ever compiled any C++ on Windows, there is literally nothing more to do than opening x64/x86 command prompt and run the 2 CMake commands selecting the preset and building, nothing more is required than a working compiler. IMO we should not burden ourselves with providing a script to install MSVC

Jayman2000:
> At first glance this appears wholly unnecessary, as building in windows using visual studio, cmake and vcpkg is nowhere near challenging. Can you describe what happened to prompt this in more detail?

As I wrote in the commit message:

> (See #332 for details).

Also, I don’t think whether or not the process is challenging really matters. The problem with the current instructions is not that they’re challenging. At the moment, you need to already know how to build Descent 3 on Windows in order to successfully follow the current instructions for building Descent 3 on Windows. The problem is that the current instructions is that they don’t teach unfamiliar users what they need to know.

README.md
Jayman2000: > I feel like this makes the process look more complicated than what it really is.

How would you recommend that I change the instructions?

> IMO we should not burden ourselves with providing a script to install MSVC.

I pushed some potential fixups for this PR [here](https://github.com/Jayman2000/Descent3-pr/tree/85899e647d921a3f134bed7cd3e81386a791e8b5?tab=readme-ov-file#building). Do you think that that version without the script is better?

Lgt2x: > How would you recommend that I change the instructions?

I'd suggest we start by stating requirements to build under Windows: a working VS installation that can compile C++. A link can be sufficient for that, maybe suggesting the desktop C++ dev component. This way, experienced developers can just skip it and not run a full script when they have nothing to do, and newbies can customize their installation and follow the process.

Then, keep the instructions as simple as possible. I see in #332 that you struggled with choosing the right command prompt, and to install VCPKG. I have to admit I did not follow the readme as now VCPKG is [installed with Visual Studio](https://devblogs.microsoft.com/cppblog/vcpkg-is-now-included-with-visual-studio/), so envvars are already properly set and I did not have to touch anything. Then, I just open the x86 command prompt, and run cmake with the win32 preset, build and it just works. x64 support is still experimental, we should not advertise it in the readme yet.

Lgt2x: Your feedback in #332 is very useful, because we don't know how many people have tried building the project, failed and quit without opening any issue. It's a challenge but providing concise AND sufficient information to build the project but crucial to keep the community and project alive.

JeodC: Visual Studio 2022, at the time of writing those windows build steps, included cmake and git in the C++ tools but not vcpkg. It was easy enough to right-click in file explorer, click Terminal, and use the command line parameters from there for vcpkg. I wasn't even aware VS2022 installed several other interpreter options. You're spot on regarding `set` and `setx` though--it does require a space instead of an equal sign.

But, `set` is a temporary use. The environment variable you set using `set` is only good for the current interpreter session. If `setx` is used, that variable is made permanent.

Jayman2000: I just pushed a new version of this PR. Here are the main differences between the old version and this new version:

- I removed the script and the `.vsconfig` file.
- I removed the part that tells you to manually clone vcpkg and set the `VCPKG_ROOT` environment variable. The new version of this PR relies on the fact that the “Desktop development with C++” workload sets up vcpkg for you.

I took these two suggestions:

> I'd suggest we start by stating requirements to build under Windows: a working VS installation that can compile C++. A link can be sufficient for that, maybe suggesting the desktop C++ dev component. This way, experienced developers can just skip it and not run a full script when they have nothing to do, and newbies can customize their installation and follow the process.

> Given VS2022 now includes vcpkg, though, that's irrelevant. That section could be simplified to `Requires Visual Studio 2022 and the Desktop Development with C++ workload` and it would be fine.

and implemented something that’s similar to them. In addition to telling users that they need Visual Studio 2022 with that workload, the new version of this PR gives users two commands that they can run if they don’t already have Git and Visual Studio 2022 with that workload. I provided those two commands because I wanted to have feature parity with Linux. On Ubuntu and Fedora, users can simply copy and paste two commands in order to make sure that they have everything that they need in order to build Descent 3. I wanted Windows users to have that same convenience. That being said, I can replace those two commands with a links for how to set up Git and Visual Studio 2022 if you guys think that that would be better.

README.md
```diff
@@ -92,47 +92,97 @@ The milestone needs testing on all platforms. Please report issues when found.
...
+    [2]: <https://github.com/microsoft/vswhere/wiki#installing>
+    [3]: <https://learn.microsoft.com/en-us/answers/questions/1689898/does-visual-studio-build-tools-2022-support-32-bit>
+    -->
+
+    ```batch
+    winget install Git.Git Microsoft.VisualStudio.2022.Community
+
+    "%ProgramFiles(x86)%\Microsoft Visual Studio\Installer\setup.exe" modify^
+        --passive^
+        --channelId VisualStudio.17.Release^
```
Lgt2x, line 115: A simple link like [this one](https://visualstudio.microsoft.com/downloads/) maybe be easier for us to keep up to date than a command to download a specific version of VS, what's your opinion @JeodC ?

README.md
```diff
@@ -92,47 +92,97 @@ The milestone needs testing on all platforms. Please report issues when found.
...
+5. **macOS only:** Install the rest of the prerequisites:
+
+    ```sh
+    brew bundle install
+    ```
+
+6. Run CMake:
+
+    ```sh
+    cmake --preset <platform>
```
Lgt2x, line 179: `<platform>` can directly be replaced by `[win32|mac|linux]` to avoid the extra line below

Lgt2x (commented):
From a user standpoint, I find this version harder to follow. In the current README version, a developer looking for build instructions has to 1) Find the build header 2) Find their OS 3) Copy and run the commands.

In your version: 1) Find the "building" header 2) Expand the subheader corresponding to the prerequisites of their OS 3) Read the next step, and skip it because it is not for their OS 4) Copy commands one by one, skipping the ones not for them.

I don't know if my explanation is clear but I feel like it's adding friction to the simple "copy 3 commands and done".

README.md
JeodC: That's how I installed VS anyway. I'm not at all sure what the above is supposed to do. Being user friendly > being fast, since what's faster for one person may well be slower for another. User friendliness guarantees a similar process speed for all.

Jayman2000: > A simple link like [this one](https://visualstudio.microsoft.com/downloads/) maybe be easier for us to keep up to date than a command to download a specific version of VS,

I think that it would technically be easier to keep a link up to date, but I don’t think that it would be significantly easier. Once Descent 3 switches to a newer version of Visual Studio, we would only have to change two numbers. We would have to change the 2022 to whatever the new year is, and we would have to change the 17 to whatever the new version is. I’m pretty confident that that’s all we’ll have to do because that’s all that had to be done to go from Visual Studio 2017 to 2019 and to go from 2019 to 2022. I checked the versions of [these](https://learn.microsoft.com/en-us/visualstudio/install/use-command-line-parameters-to-install-visual-studio?view=vs-2022) [three](https://learn.microsoft.com/en-us/visualstudio/install/workload-and-component-ids?view=vs-2022) [pages](https://learn.microsoft.com/en-us/visualstudio/install/workload-component-id-vs-community?view=vs-2022) for Visual Studio 2017 and 2019, and I found that those older versions supported all of the same command-line parameters and IDs that get used here. My prediction is that the next version(s) of the Visual Studio Installer will also support those same command-line parameters and IDs. Now, Microsoft could throw us a curve ball. They could release a version of Visual Studio that breaks compatibility with those command-line parameters. Even if that happens, I doubt that it will take much effort in order to update this command. We only need to install a single workload after all.

More importantly, I think that how easy it to keep that command up to date is irrelevant. We only ever need to update that command in two different situations:

1. We switch to a new version of Visual Studio. At most, this will happen every two or three years.

2. We start depending on a component that isn’t included in the “Desktop development with C++” workload. This will probably happen infrequently or never.

Those two things don’t happen very often, so the effort required to keep the command up to date is very low. At the same time, these two commands make it easier to test the build instructions. When testing build instructions, it’s important to uninstall all build dependencies and then follow the build instructions from the beginning. If you don’t do that, then you run the risk of having build instructions that either don’t work or only work for users that have already built the software. Providing two commands that install all of the dependencies makes doing tests like that easier and less error prone. I would know. I’ve done tests like that many times while creating this PR. Having those two commands was very helpful. In other words, the small burden of keeping two commands up to date is outweighed by the benefit of being able to test the build instructions more easily and accurately.

Additionally, those two commands make it easier for new users to get started. Making things easier for users is more important than making things easier for maintainers.

---

> I'm not at all sure what the above is supposed to do.

That command makes sure that the “Desktop development with C++” workload is installed.

> Being user friendly > being fast, since what's faster for one person may well be slower for another.

I agree, being user friendly is definitely more important that being fast. That being said, I think that copy and pasting two commands is more user friendly than linking to a download page. In fact, I included those two commands in order to make the instructions more user friendly. I wasn’t thinking about speed at all when I added them.

Once those commands are started, no user interaction is required. You just have to wait for the Visual Studio Installer window to close itself. If we link to a download page, then the user has to choose a download link, wait for it to download, open the Visual Studio Installer installer, click “Continue”, wait for the Visual Studio Installer to download and install, find and check the correct workload, click “Install”, wait for it to finish, and then close the Visual Studio Installer. Copy and pasting two commands is just easier. Lgt2x said previously ‘I don't know if my explanation is clear but I feel like it's adding friction to the simple "copy 3 commands and done".’ By linking to a download page, I feel like we would be adding friction to the simple “copy 2 commands and done”.

Also, you said “what's faster for one person may well be slower for another”. Wouldn’t copy and pasting commands be the same speed for everyone?

Jayman2000:
> From a user standpoint, I find this version harder to follow. In the current README version, a developer looking for build instructions has to 1) Find the build header 2) Find their OS 3) Copy and run the commands.
>
> In your version: 1) Find the "building" header 2) Expand the subheader corresponding to the prerequisites of their OS 3) Read the next step, and skip it because it is not for their OS 4) Copy commands one by one, skipping the ones not for them.
>
> I don't know if my explanation is clear but I feel like it's adding friction to the simple "copy 3 commands and done".

OK. I’ll submit a new version that splits everything back into separate headings after we wrap up the Visual Studio discussion.

README.md
Lgt2x: > Additionally, those two commands make it easier for new users to get started. Making things easier for users is more important than making things easier for maintainers.

Ok to keep the command

Jayman2000:
> OK. I’ll submit a new version that splits everything back into separate headings after we wrap up the Visual Studio discussion.

I just force pushed a new version that splits everything back into separate headings.

README.md
Jayman2000: I actually fixed this a while ago, but I forgot to mark this as resolved. Oops.

Lgt2x:
tag me when you're ready for review :)

Jayman2000:
@Lgt2x I’m ready for review.

README.md
```diff
@@ -92,47 +92,82 @@ The milestone needs testing on all platforms. Please report issues when found.
...
+        --channelId VisualStudio.17.Release^
+        --productId Microsoft.VisualStudio.Product.Community^
+        --add Microsoft.VisualStudio.Workload.NativeDesktop;includeRecommended
+    ```
+
+2. Open a “x86 Native Tools Command Prompt” and run:
+
+    ```batch
+    git clone https://github.com/DescentDevelopers/Descent3
+    cd Descent3
```
Lgt2x, line 120: Now that I see it, I wonder if these 2 lines should go under the headline "Building", with something along the lines of "First, clone the Descent 3 repository: ... Then, follow platform-specific compilation instructions". What do you think about that? Good job otherwise, I really like these new instructions

Jayman2000: > Now that I see it, I wonder if these 2 lines should go under the headline "Building", with something along the lines of "First, clone the Descent 3 repository: ... Then, follow platform-specific compilation instructions". What do you think about that?

OK, so I just tried changing the instructions to work that way, and [I don’t like how they look](https://github.com/Jayman2000/Descent3-pr/tree/532505f3f2e6f058c8dfdbf06e5fead4ba55d266?tab=readme-ov-file#building). At the moment, the instructions tell you to install prerequisites, clone the repository and then run CMake. In order to move the cloning step to the beginning, I had to tell users to install some prerequisites, then clone the repo, then install more prerequisites and then run CMake. I think that the instructions flow better when the steps for installing prerequisites are grouped together. Additionally, moving the cloning step to the beginning resulted in Windows users having to use four different Command Prompt windows instead of two.

I suppose that we could skip the “make sure that Git is installed” part, but then the instructions would fail for users that don’t already have the `git` command on their path. For example, if a user is on Windows and they only have [GitHub Desktop](https://desktop.github.com) installed, then the `git` command won’t be on their path. Additionally, Windows users would still end up using three different Command Prompt windows even if we removed that part. They would use a regular Command Prompt in order to clone the repo, an administrator Command Prompt in order to install Visual Studio 2022 and the C++ workload, and an x86 tools Command Prompt in order to run CMake.

> Good job otherwise, I really like these new instructions

Thank you.

Lgt2x: Indeed I also prefer the previous version, this one is more cluttered than I thought. Let's revert this last change and merge the PR

JeodC: Does this include the new ninja generator that windows builds use? I don't think VS includes it by default. I could be blind here but I'm not seeing it in the instructions.

Lgt2x: Isn't ninja shipped with Visual Studio when you install CMake nowadays?

JeodC: A second opinion would be nice, but when I reinstalled VS it didn't come with ninja--at least the cmake preset wouldn't take.

Jayman2000: > Does this include the new ninja generator that windows builds use? I don't think VS includes it by default. I could be blind here but I'm not seeing it in the instructions.

I had a similar concern when I was rebasing this PR after #422 was merged. I had considered doing this:

```diff
diff --git a/README.md b/README.md
index 41aee876..ae79ca5b 100644
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ The milestone needs testing on all platforms. Please report issues when found.
     -->

     ```batch
-    winget install Git.Git Microsoft.VisualStudio.2022.Community
+    winget install Git.Git Microsoft.VisualStudio.2022.Community Ninja-build.Ninja

     "%ProgramFiles(x86)%\Microsoft Visual Studio\Installer\setup.exe" modify^
         --passive^
```

But after testing, I found that it wasn’t necessary.

---

> Isn't ninja shipped with Visual Studio when you install CMake nowadays?

I think that this is correct. I wasn’t able to find any documentation that mentions that Ninja comes with Visual Studio’s CMake component, but I was able to verify that it comes with Visual Studio’s CMake component by doing the following:

1. Make sure that CMake and Ninja aren’t installed outside of Visual Studio:

    ```batch
     winget uninstall Kitware.CMake Ninja-build.Ninja
    ```

2. Uninstall the Visual Studio Installer.

3. Install Visual Studio Build Tools 2022 and the “C++ CMake tools for Windows” component. To do so, open an elevated Command Prompt and run:

    ```batch
    winget install Microsoft.VisualStudio.2022.BuildTools

    "%ProgramFiles(x86)%\Microsoft Visual Studio\Installer\setup.exe" modify^
        --passive^
        --channelId VisualStudio.17.Release^
        --productId Microsoft.VisualStudio.Product.BuildTools^
        --add Microsoft.VisualStudio.Component.VC.CMake.Project
    ```

4. Open “x86 Native Tools Command Prompt for VS 2022”.

5. Make sure that Ninja is installed:

    ```batch
    where ninja
    ```

---

> I receive the following when using a fresh install of VS 2022 community per instructions, along with C++ desktop tools.
> 
> ```
> [cmake] CMake Error: CMake was unable to find a build program corresponding to "Ninja Multi-Config".  CMAKE_MAKE_PROGRAM is not set.  You probably need to select a different build tool.
> [cmake] CMake Error: CMAKE_C_COMPILER not set, after EnableLanguage
> [cmake] CMake Error: CMAKE_CXX_COMPILER not set, after EnableLanguage
> [cmake] -- Configuring incomplete, errors occurred!
> ```

I can’t reproduce that error. The best that I can think of is that somehow Ninja isn’t ending up on your `Path`. What happens when you run these commands?

```batch
where ninja
set Path
```

PR #414, opened 2024-06-02
Fix segfault on missing critical files

winterheart:
## Pull Request Type

- [ ] GitHub Workflow changes
- [x] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [x] Other changes

### Description

If no d3voide1.hog or d3voice2.hog in search path was found, Descent3 fails with segmentation fault after pressing cancel on "Please insert Descent 3 CD" dialogue on Linux. Added guard checks which gracefully handle this issue.

Refactored GetMultiCDPath() function which greatly simplifies it's logic and remove rare and little of use cmd-line options:

* Removed `-moviedir` option as is only available on DEBUG build and useless in development environment (we can define movies location in many other ways).
* Removed `-intro` option as is there a little use to redefine intro.mve location (again, we can control movies location in other ways).

### Checklist


Descent3/descent.cpp
```diff
@@ -515,7 +505,7 @@ void Descent3() {
           // a lowercase extension
           char extension[16], *p;
           p = extension;
-          ddio_SplitPath(GameArgs[intro_movie_arg + 1], NULL, base_intro_movie_name, extension);
+          ddio_SplitPath(GameArgs[intro_movie_arg + 1], nullptr, base_intro_movie_name, extension);
```
Lgt2x, line 508: You said in description that the "-intro" argument was removed? 8 lines higher we still search for it

Descent3/descent.cpp
```diff
@@ -807,103 +792,53 @@ struct file_vols {
...
-
-int num_cd_files = sizeof(file_volumes) / sizeof(file_vols);
-
 // This function figures out whether or not a file needs to be loaded off of
 // CD or off of the local drive. If it needs to come from a CD, it figures out
 // which CD and prompts the user to enter that CD. If they hit cancel, it
 // returns NULL.
 const char *GetMultiCDPath(const char *file) {
-  static char filepath[_MAX_PATH * 2];
+  std::vector<file_vols> file_volumes = {
```
Lgt2x, line 800: mark const

Lgt2x (commented):
works on Linux, good cleanup

Descent3/descent.cpp
winterheart: Yeah, missed that part. Good thing, I was able refactor a little bit intro playback.

PR #423, opened 2024-06-06
Create new 'win' preset for Ninja (native) builds

tophyr:
Keep win32 and win64 using MSBuild; do not add build parallelism (it apparently causes races)

## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,7 +16,7 @@ on:
 
 jobs:
   build:
-    name: ${{ matrix.os.runner }}, ${{ matrix.os.cxx }}, ${{ matrix.build_type }}
+    name: ${{ matrix.os.name }}, ${{ matrix.build_type }}
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


tophyr:
@Lgt2x for your consideration

CI config not yet updated

tophyr:
ok, `Windows-Ninja` CI config added and passing

Lgt2x:
Why would we keep msbuild? ninja seems better from every perspective

tophyr:
@Lgt2x Ninja appears incapable of producing a 32-bit binary on a 64-bit build host (or a 64-bit binary on a 32-bit build host), is why

Lgt2x:
> @Lgt2x Ninja appears incapable of producing a 32-bit binary on a 64-bit build host (or a 64-bit binary on a 32-bit build host), is why

Actually you can. I figured out through #422  that when using the ninja generator, the only difference is that you need to load the compiler toolchain manually, whereas msbuild does it automatically when you provide the `-A` option or `architecture` Cmake preset parameter.

In order to load the toolchain, can either 1) Use [the VS script that does it for you](https://learn.microsoft.com/en-us/cpp/build/building-on-the-command-line?view=msvc-170#developer_command_file_locations), that's what the CI build do under the hood with the `msvc-dev-cmd` action. or 2) Launch the x64/x86 command prompt for VS in the start menu, which prepares the MSVC toolchain for either 32bit or 64bit.

So to get the 32bit build on a 64bit computer: Launch x86 native command prompt, and run `cmake --preset win32`, and use x64 command prompt/preset to get the 64bit build

tophyr:
Ah, much better then.

PR #424, opened 2024-06-06
PXO online: More const char* fixes

MaddTheSane:
This fixes a few compiler warnings discovered in Xcode.

### DIFF
--- a/netcon/includes/con_dll.h
+++ b/netcon/includes/con_dll.h
@@ -404,7 +404,7 @@ DoUI_fp DLLDoUI;
 typedef void (*Debug_ConsolePrintf_fp)(int n, const char *format, ...);
 Debug_ConsolePrintf_fp DLLDebug_ConsolePrintf;
 
-typedef void (*DedicatedServerPrintf_fp)(char *format, ...);
+typedef void (*DedicatedServerPrintf_fp)(const char *format, ...);
 DedicatedServerPrintf_fp DLLPrintDedicatedMessage;
 /*
 typedef int( *ValidateUser_fp) (validate_id_request *valid_id, char *trackerid);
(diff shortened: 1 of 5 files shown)


netcon/includes/con_dll.h
```diff
@@ -404,7 +404,7 @@ DoUI_fp DLLDoUI;
 typedef void (*Debug_ConsolePrintf_fp)(int n, const char *format, ...);
 Debug_ConsolePrintf_fp DLLDebug_ConsolePrintf;
 
-typedef void (*DedicatedServerPrintf_fp)(char *format, ...);
+typedef void (*DedicatedServerPrintf_fp)(const char *format, ...);
```
Lgt2x, line 407: Already covered by #434 but it's fine

netcon/mtclient/mtclient.cpp
```diff
@@ -1351,8 +1348,10 @@ int MainMultiplayerMenu() {
           char *pcount = strchr(tokp, ' ');
           // if(pcount)
           {
-            if (!pcount)
-              pcount = "";
+            if (!pcount) {
+              static char quietWarning[5] = "";
```
Lgt2x, line 1352: this fix does not seem optimal. How about simply `pcount = (char *)"";` ? My compiler does not warn when doing that

MaddTheSane: While `pcount = (char *)"";` does quiet the warning, the line under it will try to modify it. On most systems, strings like that are constant and it might crash the app.
An alternative is to not modify the string if `pcount` is nil.

Lgt2x: The alternative of not setting the string when pcount is nullptr seems better in this context

Lgt2x (approved):

PR #426, opened 2024-06-06
Make Debug config use ASAN

tophyr:
The Debug variant includes various asserts and sanity checks, and Address Sanitizer is really just another type of assert/sanity check. So, it follows that Debug should just have ASAN turned on. RelWithDebInfo can be used to produce a debugger-friendly yet error-tolerant build.

## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -83,6 +83,9 @@ add_custom_target(get_git_hash ALL
   DEPENDS ${PROJECT_BINARY_DIR}/Descent3/d3_version.h
 )
 
+add_compile_options("$<$<CONFIG:Debug>:-fsanitize=address>")
+add_link_options("$<$<CONFIG:Debug>:-fsanitize=address>")
+
 if(UNIX)
   add_compile_options("$<$<COMPILE_LANGUAGE:CXX>:-Wno-multichar;${BITS};${EXTRA_CXX_FLAGS}>")
   add_compile_options("$<$<COMPILE_LANGUAGE:C>:${BITS}>")


tophyr:
MSVC-based Windows Debug builds are failing because something in CMake is trying to be too smart about ASAN. Switching to Ninja (via https://github.com/DescentDevelopers/Descent3/pull/422) solves that failure, so I'm not going to try to address it here.

Lgt2x:
#422 introducing ninja for Windows is now merged, you can rebase on main branch. The flag seems to be slightly [different for MSVC](https://learn.microsoft.com/en-us/cpp/sanitizers/asan?view=msvc-170#ide-cmake). I'd also prefer ASAN to be hidden behind a CMake option such as `USE_ASAN` defaulting to `OFF`, but enabled in CI. It may not be something you want all the time when debugging

tophyr:
Using a flag is a good idea.. I had tried to chase down making a new config
for it but that's surprisingly difficult.

What do you think of making the flag available but make asan still default
to on? I do want to push people into using asan while debugging, but I
agree that we shouldn't *force* its use.

Lgt2x:
ASAN can have a big impact on runtime performance, I'd leave the default to OFF

tophyr:
I'll figure out some benchmarks to quantify that. I don't imagine that most
computers people will be developing with will have any trouble handling the
extra load, but that's something that'd be on me to justify for sure.

For actual perf-sensitive use I would expect people to use Release or
RelWithDebInfo, which wouldn't use ASAN.

Lgt2x:
I agree that a benchmark is necessary.

Lgt2x:
Closing for inactivity after 3 weeks. I still wish we can add a CMake option to enable ASan, anyone is welcome to open a PR to add it

PR #428, opened 2024-06-07
Eliminate warnings

tophyr:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Checklist

### DIFF
--- a/win32/winapp.cpp
+++ b/win32/winapp.cpp
@@ -214,7 +214,7 @@ oeWin32Application::oeWin32Application(const char *name, unsigned flags, HInstan
     oeWin32Application::first_time = false;
   }
 
-  HICON dicon = ExtractIcon(NULL, "descent 3.exe", 0);
+  HICON dicon = ExtractIcon(nullptr, "descent 3.exe", 0);
   wc.hCursor = NULL;
   wc.hIcon = dicon;
   wc.lpszMenuName = NULL;
@@ -487,6 +487,11 @@ tWin32OS oeWin32Application::version(int *major, int *minor, int *build, char *s
 
   osinfo.dwOSVersionInfoSize = sizeof(osinfo);
 
+  // GetVersionEx is deprecated and triggers a couple warnings. However, it appears to still be
+  // fully-functional, and the recommended alternatives (IsWindows* macros) don't provide the
+  // info we want (major/minor/build) nor do they provide tests for EOL products like Win95. On
+  // modern Windows, this will just return a version describing Win8 (even if OS is 10/11/etc).
+  // TODO: re-evaluate whether or not we actually even care about this information
 #pragma warning(suppress: 28159 4996)
   if (!GetVersionEx(&osinfo)) {
     return NoWin32;
--- a/win32/winapp.cpp
+++ b/win32/winapp.cpp
@@ -693,17 +693,15 @@ LRESULT WINAPI MyWndProc(HWND hWnd, UINT msg, WPARAM wParam, LPARAM lParam) {
     break;
   }
 
-  oeWin32Application *winapp = Win32_AppObjects[i].app;
-
   //	if this window not on list, then run default window proc.
-  if (i == -1 || winapp == NULL || force_default)
+  if (i < 0 || Win32_AppObjects[i].app == NULL || force_default)
     return DefWindowProc(hWnd, msg, wParam, lParam);
 
-  if (!winapp->run_handler((HWnd)hWnd, (unsigned)msg, (WParam)wParam, (LParam)lParam))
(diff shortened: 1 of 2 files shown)


win32/winapp.cpp
```diff
@@ -214,7 +214,7 @@ oeWin32Application::oeWin32Application(const char *name, unsigned flags, HInstan
     oeWin32Application::first_time = false;
   }
 
-  HICON dicon = ExtractIcon((HINSTANCE)hinst, "descent 3.exe", 0);
+  HICON dicon = ExtractIcon(NULL, "descent 3.exe", 0);
```
Lgt2x, line 217: how are you sure `hinst` is always null? And prefer `nullptr` to `NULL` in C++

win32/winapp.cpp
```diff
@@ -487,6 +487,7 @@ tWin32OS oeWin32Application::version(int *major, int *minor, int *build, char *s
 
   osinfo.dwOSVersionInfoSize = sizeof(osinfo);
 
+#pragma warning(suppress: 28159 4996)
```
Lgt2x, line 490: This is pretty opaque, please add a comment saying what this warning is, and why we disable it

win32/winapp.cpp
```diff
@@ -693,50 +694,48 @@ LRESULT WINAPI MyWndProc(HWND hWnd, UINT msg, WPARAM wParam, LPARAM lParam) {
...
+  fprintf(fp, "OS: %s %d.%d.%d %s\n",
           (os == Win9x)   ? "Win9x"
           : (os == WinNT) ? "WinNT"
           : (os == WinCE) ? "WinCE"
                           : "Non standard Win32",
           maj, min, build, desc);
 
   // get system memory info
-  MEMORYSTATUS mem_stat;
+  MEMORYSTATUSEX mem_stat{};
```
Lgt2x, line 728: So what's the difference between the 2 APIs ? are they ABI compatible?

tophyr: > On computers with more than 4 GB of memory, the GlobalMemoryStatus function can return incorrect information, reporting a value of –1 to indicate an overflow. For this reason, applications should use the [GlobalMemoryStatusEx](https://learn.microsoft.com/en-us/windows/desktop/api/sysinfoapi/nf-sysinfoapi-globalmemorystatusex) function instead.

`GlobalMemoryStatus` is a different method endpoint from `GlobalMemoryStatusEx`, so no ABI worries passing the different structure type.

This _does_ come with a minimum supported version of Windows XP, however. I don't know if Win95 support is something we want to maintain.

win32/winapp.cpp
tophyr: roger, wasn't sure if it was still just windows-dev practice to pass `NULL` for `HINSTANCE`-esque handles

> how are you sure hinst is always null?

i'm actually fairly sure that it _isn't_. however, the `ExtractIconA` method header marks that parameter as `_Reserved_`, indicating that it should always be set to null:
```
SHSTDAPI_(HICON) ExtractIconA(_Reserved_ HINSTANCE hInst, _In_ LPCSTR pszExeFileName, UINT nIconIndex);
```
```
// Reserved pointer parameters, must always be NULL.
#define _Reserved_                      _SAL2_Source_(_Reserved_, (), _Pre1_impl_(__null_impl))
```
The result of passing a real value here was a `C6388 'hinst' might not be '0': this does not adhere to the specification for the function 'ExtractIconA'` warning.

win32/winapp.cpp
winterheart: Guess it's about deprecation warning about GetVersion() call. I think this can be greatly simplified as we assume that we running on Windows 10 or later.

tophyr: yep, exactly. added a comment in https://github.com/DescentDevelopers/Descent3/pull/428/commits/ef33834a2a7c1b0f26a446de550d856bfb8c9825 to address the opacity of it tho

win32/winapp.cpp
Lgt2x: Ok I got it, I'm not too familiar with Windows APIs :)

win32/winapp.cpp
Lgt2x: thanks for the comment

win32/winapp.cpp
Lgt2x: Win95 does not matter to us, and I'm pretty sure we're already incompatible, we can safely ignore that

win32/winapp.cpp
tophyr: same lol. thankfully MSDN is Documentation Heaven

Lgt2x (approved):
All good, merging

PR #431, opened 2024-06-11
Fix server crash during client connection

notimaginative:
## Pull Request Type

- [ ] GitHub Workflow changes
- [x] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [x] Network changes
  - [ ] Other changes

### Description

The serverconn variable identifies the reliable socket used by the server. However this variable is not yet initialized at the very start of the first client connection which causes the server to crash. This fixes a bad safety check for this issue and a further instance of the bug.

### Checklist

### DIFF
--- a/networking/networking.cpp
+++ b/networking/networking.cpp
@@ -1235,7 +1235,7 @@ void nw_SendReliableAck(SOCKADDR *raddr, uint32_t sig, network_protocol link_typ
   network_address send_address;
   memset(&send_address, 0, sizeof(network_address));
 
-  send_address.connection_type = reliable_sockets[serverconn].connection_type;
+  send_address.connection_type = link_type;
 
   if (NP_TCP == link_type) {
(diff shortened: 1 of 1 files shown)


networking/networking.cpp
```diff
@@ -1281,8 +1281,8 @@ void nw_WorkReliable(uint8_t *data, int len, network_address *naddr) {
 
   reliable_socket *rsocket = NULL;
   // Check to see if we need to send a packet out.
-  if ((reliable_sockets[serverconn].status == RNF_LIMBO) &&
-      ((serverconn != -1) && (timer_GetTime() - last_sent_iamhere) > NETRETRYTIME)) {
+  if ((serverconn != -1) && (reliable_sockets[serverconn].status == RNF_LIMBO) &&
```
Lgt2x, line 1284: could even be `serverconn >= 0` to be even safer here, but it doesn't really matter

Lgt2x (approved):

PR #432, opened 2024-06-11
fix PXO chat lobby connection failure on Linux/Mac

notimaginative:
## Pull Request Type

- [ ] GitHub Workflow changes
- [x] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [x] Network changes
  - [ ] Other changes

### Description

On Linux & Mac `connect()` returns `EINPROGRESS` when the socket is set to non-blocking. This return value was not checked for with the PXO chat lobby connection, which caused the connection to ultimately fail. This change simply adds that missing value check.

### Checklist

### DIFF
--- a/netcon/mtclient/chat_api.cpp
+++ b/netcon/mtclient/chat_api.cpp
@@ -205,7 +205,7 @@ int ConnectToChatServer(const char *serveraddr, int16_t chat_port, char *nicknam
 #ifndef __LINUX__
       if (WSAEWOULDBLOCK == WSAGetLastError())
 #else
-      if (WSAEWOULDBLOCK == ret || EINPROGRESS == ret || 0 == ret)
+      if (EINPROGRESS == ret || 0 == ret)
 #endif
       {
         DLLmprintf(0, "Beginning socket connect\n");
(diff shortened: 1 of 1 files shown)


netcon/mtclient/chat_api.cpp
```diff
@@ -205,7 +205,7 @@ int ConnectToChatServer(const char *serveraddr, int16_t chat_port, char *nicknam
 #ifndef __LINUX__
       if (WSAEWOULDBLOCK == WSAGetLastError())
 #else
-      if (WSAEWOULDBLOCK == ret || 0 == ret)
+      if (WSAEWOULDBLOCK == ret || EINPROGRESS == ret || 0 == ret)
```
Lgt2x, line 208: `WSAEWOULDBLOCK` seems to be a windows-only error code, we shouldn't need to test for it in this block that is only compiled on *NIX

notimaginative: It's defined to a similar valid error code on non-Windows, but it's a code not returned by `connect()`. So it is a useless test here. However other such instances in the code base do check for all three error codes on non-Windows so I wanted to keep it similar to existing code.

Lgt2x: if it's useless then we should remove it

notimaginative: Done

Lgt2x: thanks

PR #429, opened 2024-06-09
Fixed failing MSVC build

pzychotic:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Documentation or Wiki changes
- [x] Build and Dependency changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

When starting from a completely new/empty build directory where `$<TARGET_FILE_DIR:Descent3>` doesn't exist yet, the copy command for `LICENSE` will create a **file** named `Debug` or `Release`, depending on the current configuration, in the Descent3 directory. This then prevents the creation of the output directory for the Descent3 target with the same name and thus failing the build.
I'm not quite sure why this doesn't happen on our CI builds. Maybe it's some difference between IDE and command line builds or just luck in case of parallel builds...

### Checklist

- [x] I have tested my changes locally and verified that they work as intended.
- [x] I have documented any new or modified functionality.
- [x] I have reviewed the changes to ensure they do not introduce any unnecessary complexity or duplicate code.
- [x] I understand that by submitting this pull request, I am agreeing to license my contributions under the project's license.

PR #435, opened 2024-06-12
Create THIRD_PARTY.md for external code handling

winterheart:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

Added all external code and license notices. Install THIRD_PARTY.md on installation step. 

### Related Issues

Closes #429.

### Checklist


THIRD_PARTY.md
```diff
@@ -0,0 +1,56 @@
+# Third party components
+
+This file contains information about third party components, required for
+building Descent 3.
```
Jayman2000, line 4: That comma looks out of place. I would change it to this:

```suggestion
This file contains information about third party components that are required
for building Descent 3.
```

Jayman2000 (changes requested):
This PR would benefit from a more in depth commit message. At the moment, the commit message explains what the changes are, but it doesn’t explain why the changes were made. Here’s a list of things that I would include in the commit message:

- I would include an explanation of what the problem in #429 was and an explanation of what the commit does to fix it.
- I would also include an explanation of why the `third_party_licenses/` directory was replaced with `THIRD_PARTY.md`. (Personally, I would rather keep each license in its own file, but if there’s a good reason to put them all in one file, then I’m all ears).
- I would also include an explanation of why the MIT License was chosen for stb. [The upstream stb repo says “This software is available under 2 licenses -- choose whichever you prefer”](https://github.com/nothings/stb/blob/013ac3beddff3dbffafd5177e7972067cd2b5083/LICENSE#L1), and this PR chooses one of the licenses over the other. I would add an explanation of why one of the two licenses was included in `THIRD_PARTY.md` but not the other.

winterheart:
STB licensed under MIT because MIT is compatible with GPLv3, while public domain may require additional explanations for re-licensing. I prefer to use clear statement on license compatibility.

Jayman2000:
> > What makes the MIT License’s GPL compatibility more clear than the Unlicense’s GPL compatibility?
> 
> GPL is less problematic than the unlicense (not strictly equivalent to public domain) for a number of reasons [[¹]](https://softwareengineering.stackexchange.com/questions/147111/what-is-wrong-with-the-unlicense)

If that’s the reason why this PR only uses the MIT License for stb, then I would put that reason into the commit message. I think that that reason is stronger than the clarity thing, although I still don’t fully understand the clarity thing.

Lgt2x (approved):
LGTM, files install correctly

PR #436, opened 2024-06-13
Changes to memory management functions

winterheart:
## Pull Request Type

- [ ] GitHub Workflow changes
- [x] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [x] Other changes

### Description

Enable RTL (real-time library) memory functions by default which improves overall performance. Replace Global* memory functions with mem_* equivalents from mem.h, remove compatibility functions on Linux / macOS platforms.

### Checklist

### DIFF
--- a/lib/mem.h
+++ b/lib/mem.h
@@ -74,14 +74,12 @@
 #ifndef MEM_H
 #define MEM_H
 
-// #define MEM_USE_RTL	1
-
-
+// Memory management debugging
 #ifdef MEM_USE_RTL
 #define mem_malloc(d) malloc(d) // Use this if your going to run BoundsChecker
 #define mem_free(d) free(d)
 #define mem_strdup(d) strdup(d)
-#define mem_size(d) _msize(d)
+#define mem_size(d) mem_size_sub(d)
 #define mem_realloc(d, e) realloc(d, e)
 #else
 // Use this if your going to NOT run BoundsChecker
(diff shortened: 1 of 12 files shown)


lib/mem.h
```diff
@@ -74,14 +74,14 @@
 #ifndef MEM_H
 #define MEM_H
 
-// #define MEM_USE_RTL	1
+#define MEM_USE_RTL	1
```
Lgt2x, line 77: Do you think there is a case where we would want to go back? If not, we should remove the define completely. Otherwise make it a CMake option (but I personally think it's unnecessary)

mem/mem.cpp
```diff
@@ -276,29 +278,44 @@ void mem_free_sub(void *memblock) {
...
-#if defined(MACOSX)
+#ifdef __LINUX__
+#ifdef MACOSX
   return malloc_size(memblock);
 #else
   return malloc_usable_size(memblock);
 #endif
+#elif WIN32
+  return _msize(memblock);
+#else
```
Lgt2x, line 308: we can safely assume that platform is either LINUX/MACOS or WIN32, it's already what we do in many other places

lib/mem.h
winterheart: This switch was for debugging memory allocations, but it was enabled for some reason. Personally, I don't find any use for this functionality, but who knows, maybe this can be somewhere useful. CMake option is good idea.

mem/mem.cpp
winterheart: Actually, there is `ifdef` on global level for platform check, so this part of code actually unreachable. Removed.

Lgt2x (approved):

PR #437, opened 2024-06-14
opengl: use SDL_WINDOW_FULLSCREEN_DESKTOP and an FBO.

icculus:
This now renders to an OpenGL Framebuffer Object at the game's resolution, and blits it to the window at whatever resolution it is currently using, scaling and letterboxing if necessary.

Which is to say: display resolutions are now imaginary, and we never change the physical display mode now. A smaller resolution is simply drawing less pixels and scaling them up with the GPU for display. This solves a few problems: no more resizing background windows or desktop icons shuffling around, no more being stuck in a weird resolution when debugging or if the game crashes, no more waiting on monitors to click over to a new mode, and no more weird rendering when the display didn't exactly support the requested mode.

This also means the game doesn't have to drop down to 640x480 for the config menu screen when it was otherwise using some other resolution.

Some caveats:

- This _requires_ OpenGL Framebuffer Object support; there is currently no fallback if it's missing and the game will refuse to start. But any desktop hardware of the last ~20 years should support it. For weird embedded things or whatnot, it will be possible to add a fallback.

- This currently requires SDL. The OpenGL pieces should work on Windows, but someone would need to adjust the existing win32 code to create a fullscreen window and not change the physical display mode. It should still compile on windows and work as before (untested by me, though).

- This is only OpenGL; it does not touch the Direct3D renderer, which should continue to work as before (again, untested by me).

### DIFF
--- a/ddio/lnxkey_sdl.cpp
+++ b/ddio/lnxkey_sdl.cpp
@@ -386,10 +386,11 @@ int sdlKeyFilter(const SDL_Event *event) {
       if ((kc == KEY_ENTER) || (kc == KEY_PADENTER)) {
         extern SDL_Window *GSDLWindow;
         Uint32 flags = SDL_GetWindowFlags(GSDLWindow);
+        // (In SDL2, SDL_WINDOW_FULLSCREEN_DESKTOP is SDL_WINDOW_FULLSCREEN plus an extra bit set, so just check for _any_ fullscreen in this bitwise AND.)
         if (flags & SDL_WINDOW_FULLSCREEN) {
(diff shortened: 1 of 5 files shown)


icculus:
(Sorry this took so long to get around to! Also, this has only had extremely light testing here after I merged it into this codebase, so please feel free to try it out and let me know if something is hosed and I'll fix things up.)

bryanperris (approved):

winterheart:
Tested on Linux, works fine.

Lgt2x (approved):
Works perfectly on Linux

PR #440, opened 2024-06-16
Fixed custom target dependency for d3_version.h

pzychotic:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

_cmake/CheckGit.cmake_ uses `configure_file()` to create _d3_version.h_ in `${TARGET_DIR}/lib` aka `${PROJECT_BINARY_DIR}/lib`. But `add_custom_command()` and `add_custom_target()` calls in the root CMakeLists.txt use `${PROJECT_BINARY_DIR}/Descent3` as path to that file, expecting it in a different place. While the build still works, since the compiler looks in the right include paths for it, the dependencies between these rules/targets are technically broken.
At least msbuild will generate warnings about missing build outputs. You can see it on CI builds before we switched to Ninja: https://github.com/DescentDevelopers/Descent3/pull/429/checks#step:6:405

Also removed the wrong `ALL` parameter from `add_custom_command()`, it's only a valid parameter for `add_custom_target()`.
See: https://github.com/DescentDevelopers/Descent3/pull/429/checks#step:6:406

### Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -69,7 +69,7 @@ if(BUILD_TESTING)
 endif()
 
 add_custom_command(
-  OUTPUT ${PROJECT_BINARY_DIR}/Descent3/d3_version.h ALL
+  OUTPUT ${PROJECT_BINARY_DIR}/lib/d3_version.h
   COMMAND ${CMAKE_COMMAND}
     -D SOURCE_DIR=${PROJECT_SOURCE_DIR}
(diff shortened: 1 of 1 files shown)


winterheart:
Thanks for pointing it out.

PR #443, opened 2024-06-16
Migrating some of ddio functions to use std::filesystem::path

winterheart:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

This is initial PR for migrating some of FS operations to C++17 std::filesystem. Converting some of global variables to std::filesystem::path, greatly refactored LockFile functions (now using with C++ IOOps / IOStreams), unify some platform-dependent function to cross-platform variants.

### Checklist


ddio/ddio.h
```diff
@@ -401,51 +402,62 @@ bool ddio_GetFullPath(char *full_path, const char *rel_path);
...
+ * @return
+ * 1: Lock file doesn't exist.
+ * 2: Lock file was in a directory, but it belonged to a process that no longer exists, so a lock file _can_ be made
+ * in the directory.
+ * 3: Lock file for this process already exists.
+ * 0: Lock file currently exists in directory.
+ * -1: Illegal directory.
+ * -2: There is a lock file in the directory, but it is in an illegal format.
+ */
+int ddio_CheckLockFile(const std::filesystem::path& dir, int pid);
```
Lgt2x, line 435: this function is only used in a single place, and the result is only used as a bool:
```cpp
  int result = ddio_CheckLockFile(lock_filename, curr_pid);
  switch (result) {
  case 0:
  case -1:
  case -2:
  case 3:
    return result;
  };
```

maybe we can simplify the return type, and set it  `[[nodiscard]]`

ddio/ddio.h
```diff
@@ -401,51 +402,62 @@ bool ddio_GetFullPath(char *full_path, const char *rel_path);
...
+ * Deletes a lock file (for the current process) in the specified directory
+ * @param dir Directory for which the lock file should be deleted from
+ * @return
+ * 1: Lock file deleted.
+ * 0: Lock file not deleted, the lock file in the directory does not belong to our process.
+ * -1: Illegal directory.
+ * -2: A lock file exists in the directory, but wasn't deleted...illegal format.
+ * -3: Unable to delete file.
+ */
+int ddio_DeleteLockFile(const std::filesystem::path& dir);
```
Lgt2x, line 461: the only use of this function discards the returned value

ddio/ddio.h
```diff
@@ -401,51 +402,62 @@ bool ddio_GetFullPath(char *full_path, const char *rel_path);
...
+ * Creates a lock file in the specified directory
+ * @param dir Directory for which the lock file should be created in
+ * @return
+ * 1: Lock file created.
+ * 2: Lock file created (there was a lock file in that directory, but it belonged to a process that no longer exists)
+ * 3: Lock file for this process already exists
+ * 0: Lock file not created, a lock file currently exists in the directory
+ * -1: Illegal directory
+ * -2: There is a lock file in the directory, but it is in an illegal format
+ * -3: Unable to create lock file
```
Lgt2x, line 447: instead of a simple int, the result could be an enum or enum class.`return 3` is very opaque

ddio/winfile.cpp
```diff
@@ -599,220 +589,24 @@ const char *ddio_GetCDDrive(const char *vol) {
...
-      if (obj_ret == WAIT_TIMEOUT) {
-        // this process exists still
-        ddio_SetWorkingDir(old_directory);
-        return 0;
-      }
+/**
+ * Check process existence by PID
+ * @param pid
+ * @return true if process exists, false otherwise
+ */
```
Lgt2x, line 596: documentation is already ddio.h

ddio/file.cpp
```diff
@@ -0,0 +1,155 @@
...
+  D3::bin_read(input, header.pid);
+  return input;
+}
+
+int ddio_CheckLockFile(const std::filesystem::path& lock_filename, int pid) {
+  if(!std::filesystem::is_directory(lock_filename.parent_path())) {
+    return -1;
+  }
+
+  if (!std::filesystem::exists(lock_filename)) {
```
Lgt2x, line 50: the std::filesystem API is so nice :)

ddio/file.cpp
```diff
@@ -0,0 +1,155 @@
...
+  LockFileContent lock_content;
+  try {
+    std::ifstream lockfile(lock_filename, std::ios::binary);
+    lockfile >> lock_content;
+    lockfile.close();
+  } catch (std::exception &e) {
+    return -2;
+  }
+
+  if (lock_content.tag != std::array{'L', 'O', 'C', 'K'}) {
```
Lgt2x, line 63: the "LOCK" array should be a const array (defined in an anonymous namespace?) so you only have to declare it once

ddio/ddio.h
winterheart: Reworked functions to simply return true or false on success / failure.

Lgt2x (approved):

Issue #418, opened 2024-06-03
[Runtime Issue]: OpenGL ChangeDisplaySettings error when launching


hurricanejuiceboxing:
### Build Version

4795388500777b78b5053e163d0869bf8644faae

### Operating System Environment

- [ ] Microsoft Windows (32-bit)
- [X] Microsoft Windows (64-bit)
- [ ] Mac OS X
- [ ] Linux (specify distribution and version below)

### CPU Environment

- [ ] x86 (32-bit Intel/AMD)
- [X] x86\_64 (64-bit Intel/AMD)
- [ ] ARM (32-bit)
- [ ] ARM64 (64-bit; sometimes called AArch64)
- [ ] Other (RISC V, PPC...)

### Game Modes Affected

- [X] Single player
- [ ] Anarchy
- [ ] Hyper-Anarchy
- [ ] Robo-Anarchy
- [ ] Team Anarchy
- [ ] Capture the Flag
- [ ] Bounty
- [ ] Entropy
- [ ] Hoard
- [ ] Monsterball
- [ ] Cooperative

### Description

After following usage instructions in the readme, D3-open-source\Descent3.exe fails to launch with error:
Descent 3 Error
Error: OGL: ChangeDisplaySettings failed. Make sure your desktop is set to 16bit mode!

### Regression Status

I don't know

### Steps to Reproduce

Expected behavior: game launches as normal.
Steps to reproduce: follow "usage" section [here](https://github.com/DescentDevelopers/Descent3?tab=readme-ov-file#usage) and then try to launch the game.
Frequency: happens every time.

Lgt2x:
The game needs to be run in Windows 95 compatibility mode

JeodC:
> The game needs to be run in Windows 95 compatibility mode

Setting the default BPP to 32 will fix this fyi.

hurricanejuiceboxing:
> The game needs to be run in Windows 95 compatibility mode

This works for the 32 bit build, but is not an option for the 64 bit build.
Launching the 32 bit exe, setting the graphics bit setting to 32, then replacing the 32 bit build with the 64 bit build also results in the same error message, so it seems like the default BPP being 16 means only the 32 bit build works for now?

JeodC:
> > The game needs to be run in Windows 95 compatibility mode
> 
[...]

Issue #373, opened 2024-05-20
[Feature Request] Linux: use XDG directories


twolife:
Hi,
Currently, the configfile on linux is saved in `$XDG_DATA_HOME/Outrage Entertainment/Descent 3/` (as returned by `SDL_GetPrefPath()`), that's cool !

In addition to that, it would be nice if:
- the pilote files (*.plt) could be saved in `SDL_GetPrefPath()` too
- .`/custom/cache/` could be relocated to `$XDG_CACHE_HOME/Outrage Entertainment/Descent 3/`
- screenshots could be saved in _XDG_PICTURES_DIR_ (not an env variable, it's defined in `~/.config/user-dirs.dirs`, there is code that can be cut&past from https://gitlab.freedesktop.org/xdg/xdg-user-dirs/-/blob/master/xdg-user-dir-lookup.c to handle that)

Thanks!

JeodC:
Hi there!

As outlined by #282 and #326, the eventual goal is to confine everything to the game folder itself. Personally I fail to see an argument for keeping Registry (in windows) and XDG (in Linux) when we can consolidate it all and reduce overhead by aligning all platforms to use the same process. I'll expand on this reasoning:

Regarding screenshots, I would like to allow the user to define where they're saved. As for the pilot files, that logic is performed in [pilot.cpp](https://github.com/DescentDevelopers/Descent3/blob/3c74b79667a9fb9de1aca5ac583302bbd0a528da/Descent3/pilot.cpp#L1856) where the variable `Base_directory` is searched for any `*.plt` files. The temp directory (`custom/cache`) is also io'd based on that variable, so in the end it would be more effort than it's worth to change it for a specific platform (whereas Windows would have to use the `Documents/` library folder).

twolife:
I completely see your point to reduce maintenance overhead.
But let me explain why I see value in splitting `write()` operations to the user directory and leaving the rest of the commercial game data elsewhere (in a readonly directory):
- the game engine, GPL3 code, could be packaged by a linux distribution (the binary would be installed as `/usr/games/descent3`)
- the non-free game data could by installed by the user with something like g-d-p https://wiki.debian.org/Games/GameDataPackager in `/usr/share/games/descent3` (this is not hypothetical, I wrote the descent3 support to g-d-p some months ago, targeting the old Loki port from 2000) 

If everything is confined in the game folder, that "workflow" dies

Jayman2000:
[...]

PR #449, opened 2024-06-20
Stop calling `HeapSize()` on `malloc()`’d memory

Jayman2000:
## Pull Request Type

- [ ] GitHub Workflow changes
- [x] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [x] Other changes

### Description

This makes the Windows version of Descent 3 less dependent on Windows 95 compatibility mode. Users will still need to run the game in Windows 95 compatibility mode for the moment, but this change is required for us to eventually make Descent 3 not require Windows 95 compatibility mode. See the commit message for details.

Additionally, [I’m working on another PR](https://github.com/DescentDevelopers/Descent3/issues/373#issuecomment-2121090450) that uses `std::filesystem::directory_iterator`. Unfortunately, Microsoft’s implementation of `std::filesystem::directory_iterator` crashes when the program is run in Windows 95 compatibility mode.

### Related Issues

This PR is related to #418, but it doesn’t actually fix that issue.

### Checklist

### Additional Comments

I’ve tested the `linux` preset on NixOS 23.11 x86_64, and I’ve tested the `win32` preset on Windows 11 x86_64. [I haven’t been able to test this PR on macOS](https://github.com/quickemu-project/quickemu/discussions/1262), but it changes macOS-specific code.

### DIFF
--- a/mem/mem.cpp
+++ b/mem/mem.cpp
@@ -297,11 +297,7 @@ char *mem_strdup_sub(const char *string, const char *file, int line) {
 void *mem_realloc_sub(void *mem, int size) { return realloc(mem, size); }
 
 int mem_size_sub(void *memblock) {
-#ifdef MACOSX
-  return malloc_size(memblock);
-#else
-  return malloc_usable_size(memblock);
-#endif
+  return native_mem_size(memblock);
 }
 
 bool mem_dumpmallocstofile(char *filename) { return false; }
--- a/mem/mem.h
+++ b/mem/mem.h
@@ -74,12 +74,28 @@
 #ifndef MEM_H
 #define MEM_H
 
+#ifdef MACOSX
+#include <malloc/malloc.h>
+#else
+#include <malloc.h>
+#endif
+
+#ifdef __LINUX__
+#ifdef MACOSX
+#define native_mem_size(d) malloc_size(d)
+#else
+#define native_mem_size(d) malloc_usable_size(d)
+#endif
+#else
+#define native_mem_size(d) _msize(d)
+#endif
+
 // Memory management debugging
 #ifdef MEM_USE_RTL
 #define mem_malloc(d) malloc(d) // Use this if your going to run BoundsChecker
 #define mem_free(d) free(d)
 #define mem_strdup(d) strdup(d)
-#define mem_size(d) mem_size_sub(d)
+#define mem_size(d) native_mem_size(d)
 #define mem_realloc(d, e) realloc(d, e)
 #else
 // Use this if your going to NOT run BoundsChecker


winterheart:
I think is better just simplify WIN32's variant of `mem_size_sub()` in mem.cpp just like you want it in macro:

```cpp
int mem_size_sub(void *memblock) {
  return _msize(memblock);
}
```

Jayman2000:
> I think is better just simplify WIN32's variant of `mem_size_sub()` in mem.cpp just like you want it in macro:
>
> ```c++
> int mem_size_sub(void *memblock) {
>   return _msize(memblock);
> }
> ```

I think that doing that would fix one bug but introduce another bug. [According to the Microsoft C runtime library (CRT) reference:](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/msize?view=msvc-170)

> The `_msize` function returns the size, in bytes, of the memory block allocated by a call to `calloc`, `malloc`, or `realloc`.

When `MEM_USE_RTL` is defined, stuff like this will happen:

```c++
void *pointer = mem_malloc(100);  // This ends up calling malloc().
size_t size = mem_size(pointer);  // This ends up calling _msize().
// Calling _msize() on a pointer returned by malloc() is OK.
```

When `MEM_USE_RTL` is not defined, stuff like this will happen:

```c++
void *pointer = mem_malloc(100);  // This ends up calling HeapAlloc().
size_t size = mem_size(pointer);  // This ends up calling _msize().
// Calling _msize() on a pointer returned by HeapAlloc() is bad.
// You’re supposed to call HeapSize() when dealing with pointers returned by HeapAlloc().
```

winterheart:
Well, if main goal is get rid of Heap calling, it would be better get rid them in another mem_* functions too. There lot of code in mem.cpp for WIN32 can be cleaned up.

Jayman2000:
> Well, if main goal is get rid of Heap calling, it would be better get rid them in another mem_* functions too.

That’s not the main goal. As I wrote in the commit message,

> The main motivation behind this change is to make it so that Windows
> users can run Descent 3 in the DWM8And16BitMitigation compatibility
> mode.

I don’t really know anything about Windows’s `Heap*` functions, so I don’t know if it would be a good idea to stop calling them. I hadn’t heard about them until I started trying to run Descent 3 without Windows 95 compatibility mode.

JeodC:
I have been able to avoid using compatibility settings for Descent 3 ever since I made the default BPP 32 instead of 16. I am not sure if this is explicitly needed in the compatibility context.

Jayman2000:
> I have been able to avoid using compatibility settings for Descent 3 ever since I made the default BPP 32 instead of 16. I am not sure if this is explicitly needed in the compatibility context.

Did you set the default BPP to 32 before or after [`ENABLE_MEM_RTL`](https://github.com/DescentDevelopers/Descent3/blob/fa3f379e41477b3786f700339a413c4163bf61f8/CMakeLists.txt#L16) was added?  I didn’t start encountering this problem until after `ENABLE_MEM_RTL` was added.

JeodC:
> > I have been able to avoid using compatibility settings for Descent 3 ever since I made the default BPP 32 instead of 16. I am not sure if this is explicitly needed in the compatibility context.
> 
> 
> 
> Did you set the default BPP to 32 before or after [`ENABLE_MEM_RTL`](https://github.com/DescentDevelopers/Descent3/blob/fa3f379e41477b3786f700339a413c4163bf61f8/CMakeLists.txt#L16) was added?  I didn’t start encountering this problem until after `ENABLE_MEM_RTL` was added.

Before. Like a month ago. I haven't recompiled for windows yet to test the recent merges.

Lgt2x:
I have not taken the time to understand D3 memory management system yet as others did, but `HeapSize` call in `mem_size_sub` mem.cpp has been causing crashes/memory corruption in my windows branch, could this PR solve this issue?

pzychotic:
> I have not taken the time to understand D3 memory management system yet as others did, but `HeapSize` call in `mem_size_sub` mem.cpp has been causing crashes/memory corruption in my windows branch, could this PR solve this issue?

It does fix the memory corruption problems on Windows for me.

Jayman2000:
> I have not taken the time to understand D3 memory management system yet as others did, but `HeapSize` call in `mem_size_sub` mem.cpp has been causing crashes/memory corruption in my windows branch, could this PR solve this issue?

I dropped the “Hack to get the Heap memory size function working” commit from your `win-sdl2` branch and then I cherry picked this PR on top of it. After doing that, I was able to load into a level without it crashing.

As a stopgap solution, you can compile the `win-sdl2` with `ENABLE_MEM_RTL=OFF`.

Lgt2x:
Thanks for taking the time to test, waiting on @winterheart opinion to merge this

winterheart:
I'll recheck the code related to the problem and try to rework it taking into account the fact that Windows part can use different memory allocation methods.

winterheart:
Please see #456 for proposed fix.

PR #450, opened 2024-06-21
Clang-format entire codebase

tophyr:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [x] Other changes

### Description

Adds lint scripts that run `clang-format`, and formats the codebase.

### Checklist

### Additional Comments

Am curious why the existing `.github/workflows/clang-format-check.yml` job isn't yelling about things. Want to understand that before merging this; this will be useless and just noisy unless CI enforces clang-format going forward.

### DIFF
--- a/lint.bat
+++ b/lint.bat
@@ -1,0 +1,56 @@
+@echo off
+setlocal enabledelayedexpansion
+
+rem Runs clang-format on all *.cpp/*.h/*.json files in the project, excluding builds/
+set "FILE_TYPES=*.cpp *.h *.json"
+set "EXCLUDED=builds"
+
+rem Make sure we have a clang-format to use.
+rem Prefer any clang-format already in the user's %PATH%.
+set "PATH=%PATH%;%VCINSTALLDIR%\Tools\Llvm\bin\"
+where /q clang-format
+if %ERRORLEVEL% neq 0 (
+  echo Unable to find clang-format. Are you in a Developer command prompt?
+  exit /b 1
+)
+
+rem Figure out how parallel we're gonna be
+if /i "%1" == "-j" (
+  shift
+  set /a JOBS=%1
+) else if "%1" == "" (
+  set /a JOBS=%NUMBER_OF_PROCESSORS%
+) else (
+  echo Usage: lint.bat ^[-j ^<NUM_JOBS^>^]
+  echo     If -j is not specified, NUM_JOBS defaults to the number of processors ^(%NUMBER_OF_PROCESSORS%^)
+  exit /b 1
+)
+
+rem Generate a temporary filename to store the list of files
+:gen_temp_file
+set CPP_FILES="%TMP%\lint~%RANDOM%.tmp"
+if exist %CPP_FILES% goto :gen_temp_file
+
+rem Find all the files, and split them up amongs %JOBS% different files, round-robin style
+set /a i=1
+for /f "usebackq tokens=*" %%A in (`dir %FILE_TYPES% /s /b ^| findstr /v "%EXCLUDED%"`) do (
+  rem !i! syntax here is "immediate" expansion: https://ss64.com/nt/delayedexpansion.html
+  echo %%A >> %CPP_FILES%-!i!
+  set /a i=!i! + 1
+  if !i! gtr %JOBS% (set /a i=1)
+)
+
+rem start a clang-format task for each filelist created
+for /l %%A in (1,1,%JOBS%) do (
+  rem Use `start /b` to asynchronously spawn a new cmd.exe session running a clang-format and then a delete
+  rem The clang-format cmd is stdout/stderr redirected to NUL, and the ^& is an "and then" joiner
+  rem In batch, `foo & bar` is a way to indicate both foo and bar on one line, one after another. The ^ here
+  rem is an escape, indicating that the & character should be passed to cmd, *not* start
+  start "" /b cmd /c clang-format -i --files %CPP_FILES%-%%A >NUL 2>&1 ^& del %CPP_FILES%-%%A
+)
+
+rem Loop until all split filelists are deleted by their asynchronous cmds. Poor man's semaphore, file-style.
+:wait_for_done
+for /l %%A in (1,1,%JOBS%) do (
+  if exist %CPP_FILES%-%%A goto :wait_for_done
+)
--- a/lint.sh
+++ b/lint.sh
@@ -1,0 +1,74 @@
+#!/bin/sh
+
+# Runs clang-format on all *.cpp/*.h/*.json files in the project, excluding builds/
+FILE_TYPES="*.cpp *.h *.json"
+EXCLUDED="*/builds/*"
+
+# Make sure we have a clang-format to use.
+# Prefer any clang-format already in the user's $PATH.
+# PATH=$PATH:/usr/bin/
+which -s clang-format || (
+    echo Unable to find clang-format in \$PATH.
+    exit 1
+)
+
+# parse arguments
+while [ "$1" != "" ]; do
+    case $1 in
+        -j | --jobs )
+            shift
+            JOBS="$1"
+            ;;
+        * )
+            echo "Usage: lint.sh [-j|--jobs <NUM_JOBS>]"
+            echo "    If -j is not specified, NUM_JOBS defaults to the number of CPU threads."
+            exit 1
+            ;;
+    esac
+    shift
+done
+
+# Figure out how parallel we're gonna be
+# default to simple single-threaded xargs
+EXECUTOR=xargs
+if [ "$JOBS" != "1" ]; then
+    # if xargs supports parallelism, use that
+    if (echo foo bar | xargs -lkj -P 0$JOBS true 2>/dev/null); then
+        # prefix the $JOBS with 0 so that unspecified JOBS becomes "0" and specified becomes "02" which is still valid
+        # -n 1 so that xargs doesn't simply collect all the args into one job and break parallelism regardless
+        EXECUTOR="xargs -P 0$JOBS -n 1"
+    # if xargs doesn't support it, see if we can use GNU parallel
+    elif which -s parallel; then
+        # -X to spread the parameters evenly among jobs
+        EXECUTOR="parallel -X"
+        # if JOBS is explicitly set, add it (don't use 0-prefix trick; it triggers some buggy behavior in the "-j 0" case)
+        if [ "$JOBS" != "" ]; then
+            EXECUTOR="$EXECUTOR -j $JOBS"
+        fi
+    # if nothing supports parallelism, warn if the user appears to expect it
+    elif [ "$JOBS" != "" ]; then
+        echo "Warning: Parallel jobs specified, but installed xargs does not support -P and GNU parallel not found. Defaulting to single-threaded." >&2
+    fi
+fi
+
+# takes the FILE_TYPES and EXCLUDED vars and concats them into find parameters
+function path_args() {
+    for i in $1; do
+        echo -path "$i" -or
+    done
+}
+
+# finally, run the job
+find .                              \
+    -type f                         \
+    \(                              \
+        $(path_args "$FILE_TYPES")  \
+        -false                      \
+    \)                              \
+    -not \(                         \
+        $(path_args "$EXCLUDED")    \
+        -false                      \
+    \)                              \
+    -print0                         |
+# pipe the file list into the executor (xargs or parallel) to run clang-format
+$EXECUTOR --null clang-format -i


Jayman2000 (changes requested):
In general, I like the idea behind this PR, but I’m not so sure about the implementation. Personally, I would use [pre-commit](https://pre-commit.com) here. Instead of creating scripts that run `clang-format`, pre-commit would allow us to do this:

```diff
diff --git a/.pre-commit-config.yaml b/.pre-commit-config.yaml
new file mode 100644
index 00000000..17cce664
--- /dev/null
+++ b/.pre-commit-config.yaml
@@ -0,0 +1,6 @@
+repos:
+  -
+    repo: https://github.com/pre-commit/mirrors-clang-format
+    rev: v18.1.6
+    hooks:
+      - id: clang-format
```

If we do decide to stick with the scripts, then I have some additional suggestions about how `lint.sh` could be improved.

Also, there was already a script that ran `clang-format` in `tools/formatter.sh`. I would update this PR so that it deletes `tools/formatter.sh`.

---

> Adds lint scripts that run `clang-format`, and formats the codebase.

I see the commits that add the lint scripts, but I don’t see a commit that formats the code base.

---

> Am curious why the existing `.github/workflows/clang-format-check.yml` job isn't yelling about things.

It looks like 97a22f5bf2c2e16d204acacc4002e5c5b9fb8cdf made `clang-format-check.yml` not run automatically.

Lgt2x:
I really dislike the pre-commit hook idea. When I do a quick draft PR/commit, I don't want to be bothered by git preventing me from committing because the formatting is wrong. There are extensive articles about pre-commit hooks slowing down the development process. It also creates a dependency to clang-format for all developers. Imagine you just want to update some doc, even though you don't edit the code at all.

Automatic formatting check on CI has been disabled because it was causing problems, I don't remember the exact reason, but you can probably look back into past PRs for it.

For me scripts we can run before releases that don't bother developers are a possible solution (but please don't put them in project root directly)

Jayman2000:
> I really dislike the pre-commit hook idea. When I do a quick draft PR/commit, I don't want to be bothered by git preventing me from committing because the formatting is wrong.

What if we told developers to run `pre-commit run clang-format` instead of `pre-commit install`? What if we used [`stages: [manual]`](https://pre-commit.com/#confining-hooks-to-run-at-certain-stages)?

> There are extensive articles about pre-commit hooks slowing down the development process.

Could you link to some of them? I use [pre-commit](https://pre-commit.com) on pretty much all of my projects, so I’m interested if it’s causing me problems that I’m not aware of.

> It also creates a dependency to clang-format for all developers. Imagine you just want to update some doc, even though you don't edit the code at all.

I disagree. I don’t think that this would create a dependency on `clang-format` for all developers. For one thing, [pre-commit will automatically install `clang-format` for you if it’s not installed](https://pre-commit.com/#introduction):

> We built pre-commit to solve our hook issues. It is a multi-language package manager for pre-commit hooks. You specify a list of hooks you want and pre-commit manages the installation and execution of any hook written in any language before every commit. pre-commit is specifically designed to not require root access. If one of your developers doesn’t have node installed but modifies a JavaScript file, pre-commit automatically handles downloading and building node to run eslint without root.

Additionally, developers don’t have to run `pre-commit` if they don’t want to. It would be more accurate to say “It also creates a dependency on the `pre-commit` command for developers who want to use pre-commit.”

tophyr:
[pre-commit](https://pre-commit.com/) looks like an interesting concept. My guess is that @Lgt2x is talking strictly about _git_ pre-commit hooks, which don't have any of the fancy [pre-commit](https://pre-commit.com/) _tool_ capabilities. I have all three of Windows, Mac and Linux dev envs - I'll try the pre-commit _tool_ and see how easy it is to "bootstrap".

The primary objection I have to automatically formatting during a precommit is that I've seen situations where I explicitly want something formatted a certain way, and having that get changed underneath my feet (as opposed to failing a CI check, or giving me a warning about it pre/post commit) would piss me off.

@Jayman2000 would a commit-time _warning_, with a manually-invoked script, satisfy you? fwiw, i also get super pissed off at work when linters fail my diffs in CI - so having a local warning would be, i think, the best of both worlds.

tophyr:
> I see the commits that add the lint scripts, but I don’t see a commit that formats the code base.
@Jayman2000 good catch, looks like in my various rebases to get this onto upstream tip I actually didn't re-run the format diff.

Jayman2000:
> @Jayman2000 would a commit-time _warning_, with a manually-invoked script, satisfy you? fwiw, i also get super pissed off at work when linters fail my diffs in CI - so having a local warning would be, i think, the best of both worlds.

Yeah, that would satisfy me. That being said, I don’t even want it to be triggered before a commit happens. I would be fine if it was triggered before a commit happens, but I wasn’t trying to make that suggestion. I was only suggesting [pre-commit](https://pre-commit.com) because I thought that a 6-line YAML file looked a lot simpler than a 56-line Batch script and a 74-line shell script.

Here’s one way that the pre-commit tool could be used without it actually being triggered by `git commit`:

```diff
diff --git a/tools/.pre-commit-config.yaml b/tools/.pre-commit-config.yaml
new file mode 100644
index 00000000..37d3cd34
--- /dev/null
+++ b/tools/.pre-commit-config.yaml
@@ -0,0 +1,8 @@
+repos:
+  -
+    repo: https://github.com/pre-commit/mirrors-clang-format
+    rev: v18.1.6
+    hooks:
+      -
+        id: clang-format
+        stages: [manual]
diff --git a/tools/lint.bat b/tools/lint.bat
new file mode 100644
index 00000000..cd8ac880
--- /dev/null
+++ b/tools/lint.bat
@@ -0,0 +1,4 @@
+@echo off
+
+for /f "delims=" %%i in ('git rev-parse --show-toplevel') do set WORKTREE_ROOT=%%i
+pre-commit run --config "%WORKTREE_ROOT%\tools\.pre-commit-config.yaml" --hook-stage manual --all
diff --git a/tools/lint.sh b/tools/lint.sh
new file mode 100644
index 00000000..d7009fd8
--- /dev/null
+++ b/tools/lint.sh
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+WORKTREE_ROOT="$(git rev-parse --show-toplevel)"
+exec pre-commit run --config "$WORKTREE_ROOT/tools/.pre-commit-config.yaml" --hook-stage manual --all
```

tophyr:
> I was only suggesting [pre-commit](https://pre-commit.com/) because I thought that a 6-line YAML file looked a lot simpler than a 56-line Batch script and a 74-line shell script.

Ahh, gotcha! Ok cool, I'll check it out. The scripts are kind of hilariously onerous, for what they do (just invoke `clang-format` on all files). The complexity is really just in making them parallelize the execution. Takes the formatting job from ~15 seconds to ~2. ¯\_(ツ)_/¯

Jayman2000:
> [[1]](https://dev.to/afl_ext/are-pre-commit-git-hooks-a-good-idea-i-dont-think-so-38j6) [[2]](https://www.thoughtworks.com/insights/blog/pre-commit-don-t-git-hooked) Are you using in distributed open-source teams like ours or only for your solo projects? Even for closed teams, I've never been a fan, mostly for the reasons mentioned here.

I mainly use it for personal projects. All of my personal projects are open-source so technically anyone can contribute, but not a lot of people use my projects so I get very few contributions. I think that I’ve also encountered [pre-commit](https://pre-commit.com) when contributing to other open-source projects, but I can’t remember which ones.

tophyr:
@Jayman2000 I've tried it out, and I have mixed feelings on the `pre-commit` tool. On one hand, it seems very nice and convenient - but on the other, I don't necessarily love the idea of requiring a `pip install` and I'm getting strong "npm + left-pad" vibes... it seems like a whole lot of infrastructure to rely on just to basically run `find . -name "*.cpp" -exec clang-format -i {} \+`.

(That last part is definitely unfair and reductive; pre-commit can clearly do a _lot_ more than that. But on the other hand, we're not currently using anything else from it.)

Even my `lint` scripts here are kind of massively-overengineered, really. They're both a whole bunch of complexity in order to run clang-format optimally parallelized, when realistically the job completes serially in under a minute.

Thoughts?

Jayman2000:
> @Jayman2000 I've tried it out, and I have mixed feelings on the `pre-commit` tool. On one hand, it seems very nice and convenient - but on the other, I don't necessarily love the idea of requiring a `pip install` and I'm getting strong "npm + left-pad" vibes... it seems like a whole lot of infrastructure to rely on just to basically run `find . -name "*.cpp" -exec clang-format -i {} \+`.
>
> (That last part is definitely unfair and reductive; pre-commit can clearly do a _lot_ more than that. But on the other hand, we're not currently using anything else from it.)
>
> Even my `lint` scripts here are kind of massively-overengineered, really. They're both a whole bunch of complexity in order to run clang-format optimally parallelized, when realistically the job completes serially in under a minute.
>
> Thoughts?

When I first read your post, my thought was “Yeah, [pre-commit](https://pre-commit.com) is probably not the best tool to use in this situation. It would probably be better to just write a new, simpler clang-format script.” I created [an alternative PR that contains a simpler script](https://github.com/DescentDevelopers/Descent3/pull/480), but as I was finishing the commits in that PR’s branch, I realized something. Having a script that runs `clang-format` on everything is probably the wrong way to go about doing things.

For one thing, I’m assuming that the goal isn’t just to `clang-format` the codebase one time. I’m assuming that the goal is to keep the codebase properly formatted, so I updated and re-enabled the clang-format Check workflow. Here’s where I ran into a problem. That workflow uses [`jidicula/clang-format-action`](https://github.com/marketplace/actions/clang-format-check). [The latest version of `jidicula/clang-format-action` (v4.13.0) uses version 18.1.3 of `clang-format`.](https://github.com/Jayman2000/Descent3-pr/actions/runs/9733683888/job/26860884469#step:3:27) [At the moment, the version of Linux that I use (NixOS 24.05) ships `clang-format` 18.1.7.](https://github.com/NixOS/nixpkgs/blob/d032c1a6dfad4eedec7e35e91986becc699d7d69/pkgs/development/compilers/llvm/18/default.nix#L29) When I run the script, it formats everything successfully, but when when I push my changes, the GitHub Action complains that the code isn’t formatted correctly. `clang-format` 18.1.3 and `clang-format` 18.1.7 produce different results.

I can think of a few potential solutions to this problem:

- Turn the script into a [`#!/usr/bin/env nix-shell`](https://nix.dev/tutorials/first-steps/reproducible-scripts.html) script that uses a [`shell.nix`](https://nix.dev/tutorials/first-steps/declarative-shell#declarative-reproducible-envs) file. That way we could know for sure that the user is using `clang-format` 18.1.3. [This is basically what I did in order to finish that alternative PR](https://github.com/DescentDevelopers/Descent3/pull/480/commits/3fb2925b59a524ae637e82fba16384185a616ff1), but I took a long time to wait for clang to compile. [Also, Nix doesn’t really work on Windows yet.](https://lastlog.de/blog/libnix_mingw_status.html)

- Instead of having a workflow that runs `clang-format`, have a workflow that runs `pre-commit`. That way, the version of `clang-format` will always be the same regardless of whether its being run locally or run in CI.

- Instead of running `clang-format` directly, tell users to run the `clang-format` workflow locally via [Act](https://nektosact.com). This has two disadvantages. For one thing, it would be much slower than running `clang-format` normally because [`jidicula/clang-format-action` is really inefficient](https://github.com/jidicula/clang-format-action/issues/147). Additionally, it’s kind of a hack. You’re not supposed to run GitHub Actions locally because GitHub Actions are supposed to create vendor lock-in. Instead of trying to workaround the vendor lock-in, I think that it would be better to choose something like pre-commit that doesn’t try to create vendor lock-in in the first place.

- Instead of running `clang-format` directly, write a script that uses [the same Docker container that `jidicula/clang-format-action` uses](https://github.com/jidicula/clang-format-action/blob/c74383674bf5f7c69f60ce562019c1c94bc1421a/check.sh#L94).

- Continue to have a script that runs whatever version of `clang-format` users happen to have on their path. If the `clang-format` Check workflow fails on a PR, have contributors manually apply the formatting changes.

> I'm getting strong "npm + left-pad" vibes...

I agree. pre-commit is effectively a platform-neutral CI system, and every CI system that I’ve ever encountered gives me npm + left-pad vibes. If you know of any CI systems that are different, the I would be interested in learning more about them.

Lgt2x:
Hi @Jayman2000 , thanks for taking the time to investigate and report possible solutions to the formatting issue. As you figured out, Github actions for formatting only tell us whether the code is properly formatted or not, and will unfortunately not format the code for you. Enforcing this will be an additional constraint for contributors, that need to install clang-format themselves and figure out a way to make clang-format happy.

Let's keep in mind that we're not Google/Facebook with 150 engineers working full time on the project, but a small team of dedicated hobbyists fond of the game, and attracting new contributors by lowering the barrier to entry should be a higher priority than keeping the code perfectly formatted at all time. Forcing potential contributors to install dedicated tools just for formatting code is not a good solution whatsoever, let alone formater version conflicts. The only case where enforcing clang-format for contributions is acceptable is when you have an automated tool that can run on the PR to reformat and push fixes automatically to your branch (see for example the [VTK gitlab bot](https://docs.vtk.org/en/latest/developers_guide/develop_quickstart.html#robot-checks)). When it's not automated, it's just a hassle for devs.

I suggest we stop wasting time on an issue as minor as this one. We'll just run clang-format manually before each release on the whole codebase, that'll be good enough for the scale of this project.

PR #452, opened 2024-06-21
Remove the CD-ROM asset loading logic and -cdrom option

Lgt2x:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

CD-ROM code needs specific hardware to be tested and maintained, and does not fit with the modernization this project envisions. I doubt the Mac hardware we now target have built-in CDrom readers anymore.

To play using data from a CD-ROM, it needs to be mounted, and the mount address to be provided to the `-setdir` argument


winterheart:
There also GetMultiCDPath() function that can be removed.

Overall PR is good, tested on Linux, no problems spotted.

Lgt2x:
> There also GetMultiCDPath() function that can be removed.
> 
> Overall PR is good, tested on Linux, no problems spotted.

In my understanding `GetMultiCDPath` does more than retrieve from a CD, it's a generic file loader. You've edited this function yourself recently. It can maybe be renamed though

PR #458, opened 2024-06-23
Fix crash in DumpTextToClipboard()

pzychotic:
## Pull Request Type

- [ ] GitHub Workflow changes
- [x] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [x] Other changes

### Description

`SetClipboardData()` expects its second parameter to be allocated with the `GMEM_MOVEABLE` flag ([see here](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setclipboarddata#parameters)) and will call `GlobalSize()` on it internally, so we can't replace it with a plain malloc.
Also the assignment of `h_text` to `ptr` was missing, and `GlobalUnlock()` was still called on h_text even after the change to malloc.

### Related Issues

Regression introduced in c41c3a7 by #436.

### Checklist

### DIFF
--- a/ddebug/windebug.cpp
+++ b/ddebug/windebug.cpp
@@ -937,9 +937,10 @@ void DumpTextToClipboard(char *text) {
   // Length of string with CRs added
   int len = strlen(text) + extra + 1;
 
-  char *h_text = (char *)mem_malloc(len);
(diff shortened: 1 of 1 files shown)


Lgt2x (approved):
windows allocations are a pain... thanks for the fix

PR #459, opened 2024-06-24
Create an Android build

tophyr:
## Pull Request Type
- [x] GitHub Workflow changes
- [x] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [x] Other changes

### Description
Adds Android as a supported platform. Uses Gradle to perform the build overall.

### Checklist


Lgt2x:
Just a suggestion... You can add the __LINUX__ define for android platforms instead of adding it to every check

tophyr:
> Just a suggestion... You can add the `__LINUX__` define for android platforms instead of adding it to every check

SDL specifically undefines `__LINUX__` when `ANDROID` is defined, unfortunately.

.gitmodules
```diff
@@ -0,0 +1,4 @@
+[submodule "SDL"]
```
Lgt2x, line 1: already discussed in #466, not the solution we'd like to go for

CMakeLists.txt
```diff
@@ -91,23 +91,27 @@ add_custom_target(get_git_hash ALL
   DEPENDS ${PROJECT_BINARY_DIR}/lib/d3_version.h
 )
 
-if(UNIX)
+if(UNIX AND NOT HOST_TOOLS_ONLY)
```
Lgt2x, line 94: What is `HOST_TOOLS_ONLY`? I could not find any doc

Descent3/CMakeLists.txt
```diff
@@ -292,18 +292,25 @@ endif()
 
 file(GLOB_RECURSE INCS "../lib/*.h")
 
-add_executable(Descent3
-  WIN32
-  ${HEADERS} ${CPPS} ${PLATFORM_CPPS} ${INCS}
-)
+if(ANDROID)
+  add_library(Descent3
```
Lgt2x, line 296: Oh, interesting. You're supposed to build it as a lib for Android?

Descent3/LoadLevel.cpp
```diff
@@ -3240,7 +3240,7 @@ void ReadTerrainSkyAndLightChunk(CFILE *fp, int version) {
 
   Terrain_sky.flags = cf_ReadInt(fp);
 
-#ifdef __LINUX__
+#if defined(__LINUX__) || defined(ANDROID)
```
Lgt2x, line 3243: I know SDL undefs `__LINUX__` on Android, but this is an ugly solution. I'd rather prefer to have a distinct define for all *NIX platforms rather than a condition on the define

android/build.gradle
```diff
@@ -0,0 +1,61 @@
+apply plugin: 'com.android.application'
+
+def hostBuildDir = new File(rootProject.buildDir, "host")
```
Lgt2x, line 3: I'm honestly very uneducated about Android builds and have no idea what's going on here, but you seem to know what you're doing

android/proguard-rules.pro
```diff
@@ -0,0 +1,17 @@
+# Add project specific ProGuard rules here.
+# By default, the flags in this file are appended to flags specified
+# in [sdk]/tools/proguard/proguard-android.txt
+# You can edit the include path and order by changing the proguardFiles
+# directive in build.gradle.
+#
+# For more details, see
+#   http://developer.android.com/guide/developing/tools/proguard.html
+
+# Add any project specific keep options here:
```
Lgt2x, line 10: all comments in this file?

android/src/main/res/mipmap-hdpi/ic_launcher.png
Lgt2x, line 1: do we need to embed all the SDL logos?

build.gradle
```diff
@@ -0,0 +1,25 @@
+// Top-level build file where you can add configuration options common to all sub-projects/modules.
+
+buildscript {
```
Lgt2x, line 3: can't this go in the android directory? I'd like to avoid having to many files lying in repository root

scripts/data/fullhog/d3-android-fullhog.txt
```diff
@@ -0,0 +1,112 @@
+AIGAMEi.STR
```
Lgt2x, line 1: I guess this file should not have been committed?

tools/CMakeLists.txt
```diff
@@ -1,7 +1,12 @@
-add_executable(
-	HogMaker
-	HogMaker/HogFormat.cpp
-	HogMaker/HogMaker.cpp
-)
-add_dependencies(HogMaker get_git_hash)
-target_include_directories(HogMaker PRIVATE ${PROJECT_BINARY_DIR}/lib)
+if (ANDROID)
+	find_package(HogMaker REQUIRED)
```
Lgt2x, line 2: redundant with #465

Lgt2x (changes requested):

.gitmodules
tophyr: got this working i think! great find earlier.

CMakeLists.txt
tophyr: this is an external define i set in the android build so that it would skip configuring SDL for the host tools. this may be redundant now that SDL will be provided by vcpkg - i'll experiment to see if it configures/builds sdl regardless of whether i `find_package()` it.

if i do keep this "skip all configuration except for HogMaker" var, is there a more cmake-idiomatic naming convention to use?

Descent3/CMakeLists.txt
tophyr: yep. The Android application lifecycle is pretty different from what we normally think of. I won't dive into the details unless you're interested, but the bottom line is that any application-provided native code must be `dlopen`-ed (effectively) into an existing Java process. That Java process can then call into functions exported from the library and we can begin our work from there.

Descent3/LoadLevel.cpp
tophyr: i don't really disagree, but the problem here is that [SDL2 contains literally this](https://github.com/libsdl-org/SDL/blob/SDL2/include/SDL_platform.h#L63-L67):
```
#ifdef ANDROID
#undef __LINUX__
#endif
```

SDL3, for what it's worth, does _not_ contain this. might be worth just chasing down migration to SDL3? i'd be happy to drive that; I'm gonna have to get real familiar with the SDL innards anyway for OpenXR integration.

android/build.gradle
tophyr: yeah, this is all just super-standard android app stuff.

android/proguard-rules.pro
tophyr: copypasta for the time being. ProGuard is a Java optimizer and obfuscator. i would be extremely surprised if we ended up needing it at all - obfuscation would be pointless since we're OSS, and optimization will likely be equally pointless because the java's only purpose is to load and call into cpp.

android/src/main/res/mipmap-hdpi/ic_launcher.png
tophyr: eep, no, these are copypasta for the time being. we do need some sort of icon, but it should definitely be D3-related, not SDL.

build.gradle
tophyr: unfortunately some of the Gradle bits need to live in the root dir. Android Studio (an IDE based upon IntelliJ) treats the "project root" as the directory that contains the root `build.gradle` and `settings.gradle`. They _could_ live in the `android/` dir, but then AS would end up ignoring all of the cpp code - which would be extremely inconvenient for development.

scripts/data/fullhog/d3-android-fullhog.txt
tophyr: this i just copied from `d3-linux-fullhog.txt` actually. currently it just satisfies CMake's target-graph. haven't actually tested any of it.

tools/CMakeLists.txt
tophyr: this PR has a bunch of those other PRs merged into it :)

Descent3/LoadLevel.cpp
pzychotic: Maybe we can move the SDL includes into a small wrapper header file where we` #define __LINUX__` in case of `ANDROID` after their includes and only include the wrapper in our sources.

tophyr: I thought about this, but depending on include order it could end up introducing some ODR violations. (Not that SDL's behavior isn't introducing the same risk 👎 )

@Lgt2x @pzychotic @winterheart how about this - particularly in light of Louis' "prefer to have a distinct *NIX marker" idea: This appears to be using `__LINUX__` _really_ as a _POSIX_ API indicator. What if I went through (in a separate PR) and replaced any POSIX-indicative-but-not-actually-Linux-exclusive conditional to use a `POSIX` define instead? Then we could use `__LINUX__`, `__MACOSX__` and `ANDROID` (and presumably `__IPHONE__` etc in the future) to disambiguate _specific_ OS differences, while still be able to lump all "*nix" things together.

pzychotic: fine with me

tophyr: Split this particular bit out into https://github.com/DescentDevelopers/Descent3/pull/492

tophyr:
Ready for review.

Includes https://github.com/DescentDevelopers/Descent3/pull/493, https://github.com/DescentDevelopers/Descent3/pull/492, https://github.com/DescentDevelopers/Descent3/pull/469, and https://github.com/DescentDevelopers/Descent3/pull/465 as dependencies (see merge commit at base). Those PRs should be reviewed and merged prior to this one.

tophyr:
@Lgt2x this is finally ready! The game builds, has a mechanism to easily (and legally) put gamedata on the device, and will start up a level.

There's still a fair bit of work to do in order get the game _playable_, but at this point I think it's stable enough to start relying on build signals from GitHub CI.

Do note that this now PR includes the `ExternalProject` change from #465.

android/src/main/assets/index.html
```diff
@@ -0,0 +1,16 @@
+<html>
+<head>
+  <link href="css/style.css" rel="stylesheet"/>
+  <script src="js/jquery-3.7.1.js" type="text/javascript"></script>
```
Lgt2x, line 4: I don't see anywhere jquery is actually used in the HTML files

tophyr:
It's not, I had high hopes of making nice pretty reactive upload screens
but then remembered I can't UI a paper bag

android/src/main/assets/index.html
tophyr: It's not, I had high hopes of making nice pretty reactive upload screens but then remembered I can't UI a paper bag

Lgt2x:
Overall this is a pretty large and impactful port, I will need time before deciding whether to merge it. I especially want to make sure that I'm able to build, run and debug it, so I can independently fix the build ~if~ when it will break, with the goal of make it more sustainable. I'm not familiar at all with Java and Android, so wish me luck with that Android Studio thingy :)

tophyr:
Sounds like you've managed to build and run it, at least! Hit me up with any questions about this specifically or Android development in general.

PR #461, opened 2024-06-24
Define OutrageMessageBox regardless of DEBUG

tophyr:
## Pull Request Type
- [ ] GitHub Workflow changes
- [x] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [x] Other changes

### Description
`OutrageMessageBox()` doesn't depend on anything debug-specific - it appears to be conditionally-defined basically just to avoid interrupting the user in release builds. However, the codebase appears to think that `DEBUG` and `RELEASE` are perfect opposites, which is not true. (The `RelWithDebInfo` build, in particular, has neither set.)

Here we define `OutrageMessageBox` regardless of build type, and let the `RELEASE` check determine whether or not it wants to interrupt the user.

### Checklist

### DIFF
--- a/lib/pserror.h
+++ b/lib/pserror.h
@@ -164,11 +164,9 @@ void Int3MessageBox(const char *file, int line);
 #define MBOX_YESNO 2
 #define MBOX_YESNOCANCEL 3
 #define MBOX_ABORTRETRYIGNORE 4
-#if (defined(EDITOR) || defined(NEWEDITOR) || defined(_DEBUG))
 //	prints out a standard OS messagebox
 void OutrageMessageBox(const char *str, ...);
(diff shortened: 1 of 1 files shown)


Lgt2x (changes requested):
Better fix that does not require compile-time conditionals: Define `OutrageMessageBox` functions for release with empty function body in `error.cpp`, and remove all compile guards on function usage

tophyr:
> Better fix that does not require compile-time conditionals: Define `OutrageMessageBox` functions for non-debug builds with empty function body in `error.cpp`, and remove all compile guards on function usage

re-worked

Lgt2x (approved):

PR #467, opened 2024-06-24
Set default BPP to 32

JeodC:
## Pull Request Type

- [ ] GitHub Workflow changes
- [x] Runtime changes
  - [x] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

In windows x64, on trying to run the game, ChangeDisplaySettings will fail because the game does not use the new windows api to change these settings. Since Win95 compatibility mode is not an option for Win64, instead change the default BPP to 32.

There may be a more proper way to handle this, and make it more uniform to remove BPP 16 entirely, but this has worked for me since putting it in a fork over a month ago.

### Related Issues

#418

### DIFF
--- a/Descent3/game.cpp
+++ b/Descent3/game.cpp
@@ -966,7 +966,7 @@ void SetScreenMode(int sm, bool force_res_change) {
     } else {
       scr_width = FIXED_SCREEN_WIDTH;
       scr_height = FIXED_SCREEN_HEIGHT;
-      scr_bitdepth = 16;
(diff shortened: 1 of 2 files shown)


JeodC:
#457, if merged, will make this obsolete, but #457 is still in progress and does not have audio after movies.

Lgt2x (approved):
This code will (hopefully) not last long, but thanks for submitting the PR!

PR #471, opened 2024-06-27
Add `-additionaldir` command-line option

Jayman2000:
## Pull Request Type

- [ ] GitHub Workflow changes
- [x] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [x] Other changes

### Description

This PR adds a command-line argument named `-additionaldir`. `-additionaldir` allows users to store game assets in multiple different directories. For example, if a user installs Descent 3 using a package manager, then they’ll be able to do this:

```sh
Descent3 -setdir <path-to-proprietary-assets> -additionaldir <path-to-assets-installed-by-package-manager>
```

### Related Issues

This is related to #373 and #534, but it doesn’t fully implement either of those feature requests.

### Checklist

### Additional Comments

This PR is meant to be a first step. If it gets accpeted, then I’ll open PRs to do the following:
- Add a CMake option called `DEFAULT_ADDITIONAL_DIRS`. This way, package maintainers will be able to create `Descent3` executables that automatically find the other files that get installed by their Descent 3 package.
- Add documentation for `-additionaldir` and how it compares to `-setdir`.
- Allow Descent 3 to load `.pld` files from `-additionaldir` directories.


winterheart:
Hi. I'm working on  cfile / ddio refactoring and some of your changes already done in my branch. I'd suggest to wait it since after that path handling and general filesystem operations will be much simpler.

Still, there some comments on general idea (as I see it, regarding to #373 and similar issues). We have read-only and writable paths. Read-only are some preconfigured on build-time paths (like `/usr/share/Descent3` in Linux or values from Registry in Windows from preinstalled D3) and defined by configuration/command line options. Writable paths are paths that known to belong to current user (like $HOME/.config, $HOME/.local in Linux, and C:\users\USER\AppData in Windows). All of these paths together forms search path list, that should be searched from writable to read-only. Descent3 should use relative path notation for resource lookup, and cfile / ddio will handle all work to search and open requested file. There no `current dir` in search for reason. Unless we going to make "portable" mode for game, Descent3 should use proposed scheme to write user-data (cache/temp files included) to writeable paths that belongs to user. Current dir may not belong to user, so we should avoid overuse it.

Jayman2000:
> Hi. I'm working on cfile / ddio refactoring and some of your changes already done in my branch. I'd suggest to wait it since after that path handling and general filesystem operations will be much simpler.

So you’re saying that I should rebase this PR on main once your PR gets merged?

> Still, there some comments on general idea (as I see it, regarding to #373 and similar issues). We have read-only and writable paths. Read-only are some preconfigured on build-time paths (like `/usr/share/Descent3` in Linux or values from Registry in Windows from preinstalled D3) and defined by configuration/command line options. Writable paths are paths that known to belong to current user (like $HOME/.config, $HOME/.local in Linux, and C:\users\USER\AppData in Windows). All of these paths together forms search path list, that should be searched from writable to read-only. Descent3 should use relative path notation for resource lookup, and cfile / ddio will handle all work to search and open requested file. There no `current dir` in search for reason. Unless we going to make "portable" mode for game, Descent3 should use proposed scheme to write user-data (cache/temp files included) to writeable paths that belongs to user. Current dir may not belong to user, so we should avoid overuse it.

I agree. This PR implements all of what you said, except for these things:

- This PR makes it so that there’s only one writable path.
- There’s no way to specify additional/read-only paths at compile time yet. Like I mentioned before, I’ll do that in a later PR.
- This PR doesn’t add any code that looks in the Windows registry.
- At the moment, the default value for the writable directory is still the current working directory. A future PR can make it default to something else.

Jayman2000:
I just pushed a new version of this PR. Here are the changes:

- It’s been rebased onto d3659bb5 (Merge pull request #499 from vlohacks/main, 2024-07-22). I started this rebase a while ago, and me a long time to finish it. That’s the only reason why this PR hasn’t been rebased onto something newer yet.
- I had to move the `Base_directory` variable from `Descent3/init.cpp` into a its own new module. I had to move it because of the tests added in 1a3c0122 (Refactor cf_FindRealFileNameCaseInsensitive() function, 2024-07-02). Those tests were for the `cf_FindRealFileNameCaseInsensitive()` function. This PR has always deleted that function and replaced it with a new one. Here’s the problem: that new function needs access to the `Base_directory` variable. For the regular game, accessing `Base_directory` is fine because `Base_directory` is delcared in `Descent3/init.cpp`. For tests, accessing `Base_directory` is a problem. Tests don’t currently link to `Descent3/init.cpp`, and I found it really difficult to make them link to `Descent3/init.cpp`. That’s why I created a separate module to hold the `Base_directory` variable.
- On the bright side, I was able to use that separate module in order to keep all of the new functions that this PR adds in one single place. While the new module started as a workaround, I actually like it better this way. It makes things more compartmentalized.

Here are some things that I could use some help with:

- I ended up deleting the new implementation of `cf_FindRealFileNameCaseInsensitive()` that was added in 1a3c0122 (Refactor cf_FindRealFileNameCaseInsensitive() function, 2024-07-02). One of the commits from this PR replaces `cf_FindRealFileNameCaseInsensitive()` with a new function named `gd_LocatePath()`. I haven’t run any benchmarks, but I think that my `gd_LocatePath()` is probably slower than `cf_FindRealFileNameCaseInsensitive()`. I don’t really understand the code that was in `cf_FindRealFileNameCaseInsensitive()`, but it would be nice if that code could somehow be reused in the implementation of `gd_LocatePath()`.
- One of the commits introduces a regression on Windows. Specifically, if you try to start a multiplayer server, then the game will page fault.

Jayman2000:
I just pushed another new version of this PR. Here are the changes:

- It’s been rebased onto 3cb1e891 (Merge pull request #479 from DescentDevelopers/editor-fix-deprecation, 2024-08-18).
- I fixed the uninitialized memory access on Windows.
- I took some of the changes from the “Add -aditionaldir option” commit and put them in their own commit named “Add start parameter to FindArg()”. I think that this makes the commit history cleaner, and it also gave me an opportunity to write a new commit message where I justify some of my changes.

I could still use some help with the whole `cf_FindRealFileNameCaseInsensitive()` thing.

Lgt2x:
Tell us if/when you need feedback or help with anything :)

Jayman2000:
> Tell us if/when you need feedback or help with anything :)

Like I mentioned earlier, I could use some help with the `cf_FindRealFileNameCaseInsensitive()` thing. I know that @winterheart put in a significant amount of effort into making `cf_FindRealFileNameCaseInsensitive()` faster. At the moment, this PR deletes `cf_FindRealFileNameCaseInsensitive()` and replaces it with `gd_LocatePath()`. It would be nice if that faster code could somehow be incorporated into `gd_LocatePath()`, but I don’t know how to do that because I don’t really understand the code that’s in `cf_FindRealFileNameCaseInsensitive()`.

winterheart:
I would like to correct some points of PR. It's rather big feature request.

Let's try to implement feature in one of already exist submodule. As it's more file access and discovery thing, I suggest to use `cfile`. We don't have much writable paths in to be defined (actually is one - location in user's home dir), so we can inject proper path in cfopen() with checking "w" flag option. For reading we can use list of paths and hog-files specified in not yet implemented cf_Init() function. In this way cfopen() should _always_ accepts relative paths - "d3.hog", "missions/d3.msn", "movies/intro.mve" etc.

I thinking about using some cache path resolver that adds already resolved paths into LRU cache or similar by lowercased relative path filename. This cache can be used to keep tracking already opened resources and avoiding excessive case-insensitive look-ups, like cfile's `Libraries` variable does for entries of hog-files.

Regarding `cf_FindRealFileNameCaseInsensitive` -  let's keep it under cfile submodule as is, with suggested design alterations it may be used with minimal changes.

Jayman2000:
> I would like to correct some points of PR. It's rather big feature request.

Finally, someone is reading my code!

---

Before I start implementing your suggestions, I want to make sure that I fully understand them.

> Let's try to implement feature in one of already exist submodule. As it's more file access and discovery thing, I suggest to use `cfile`.

When I move all of the stuff from the `gamedata` module to the `cfile` module, should the “Move Base_directory into separate module” commit be changed so that it moves `Base_directory` into the `cfile` module?

> We don't have much writable paths in to be defined (actually is one - location in user's home dir), so we can inject proper path in cfopen() with checking "w" flag option. For reading we can use list of paths and hog-files specified in not yet implemented cf_Init() function. In this way cfopen() should _always_ accepts relative paths - "d3.hog", "missions/d3.msn", "movies/intro.mve" etc.

I think that you’re saying two things here:

1. `cfopen()` should do something like this:

    ```cpp
    CFILE *cfopen(const std::filesystem::path &relative_path, const char *mode) {
        // …
        std::filesystem::path full_path;
        if (mode[0] == 'w') {
            // cf_GetWritableBaseDirectory is the new name for gd_GetWritableBaseDirectory.
            full_path = cf_GetWritableBaseDirectory() / relative_path;
        } else if (mode[0] == 'r') {
            // cf_LocatePath is the new name for gd_LocatePath.
            full_path = cf_LocatePath(relative_path);
        }
        // …
    }
    ```

2. `gd_Init()` should be moved to the `cfile` module and renamed to `cf_Init()`.

Am I understanding correctly?

> I thinking about using some cache path resolver that adds already resolved paths into LRU cache or similar by lowercased relative path filename. This cache can be used to keep tracking already opened resources and avoiding excessive case-insensitive look-ups, like cfile's `Libraries` variable does for entries of hog-files.

So you’re saying that you might open another PR that optimizes Descent 3’s case insensitive path code further?

> Regarding `cf_FindRealFileNameCaseInsensitive` - let's keep it under cfile submodule as is, with suggested design alterations it may be used with minimal changes.

Yeah, but how am I going to make those minimal changes? Like I said previously:

> It would be nice if that faster code could somehow be incorporated into `gd_LocatePath()`, but I don’t know how to do that because I don’t really understand the code that’s in `cf_FindRealFileNameCaseInsensitive()`.

winterheart:
1. and 2. are correct. Only thing is that we should get away sometime from `Base_directory` to list of base directories (readonly and writable).

I'm planning heavily rewrite cfile / ddio modules (related to file operations), I still see some optimization points here to improve overall performance. Path cache is one of them. Currently we are about on halfway of this refactoring :).

Main issue of case insensitive discovering is that can't be optimized much. On each level of directories directory iterator loses so much time, making whole traversal incredibly slow. Optimization of `cf_FindRealFileNameCaseInsensitive()` is based in fact that `Base_directory` is already known to us, and we only have to resolve one-two sublevels of directory path. So that why I'm stressing that we should use relative paths on resource loading, this is one of ways to optimize case-insensitive mess.

Jayman2000:
> 1. and 2. are correct.

OK. In that case, I’ll move everything to the `cfile` module like you asked, but I’m not going to make the requested `cfopen()` change. It’s just too much work to hunt down and change every absolute-path `cfopen()` call, especially since I already have a working implementation of `-additionaldir` that doesn’t depend on that change. I say that it’s too much work because I’m fed up with all of the extra stuff that I’ve had to do in order to implement `-additionaldir`. For example, I had to rebase this PR because you decided that [your other PR](https://github.com/DescentDevelopers/Descent3/pull/475) should come before this one. As part of that rebase, I had to create a new module and move all of my code out of an existing module into that new module. I now have to move all of my code out of that new module and back into an existing module. As part of that previous rebase, I also had to change every instance of `example_path.string().c_str()` to `example_path.u8string().c_str()` in order to match the code that you introduced in [that other PR](https://github.com/DescentDevelopers/Descent3/pull/475). I also had file, argue for and fix [this issue](https://github.com/DescentDevelopers/Descent3/issues/483) in order to make sure that adding additional calls to the `u8string()` method doesn’t introduce additional regressions.

Those are just some examples. There’s a lot of other stuff that I’ve had to deal with in order to finish this PR. The thought of having to hunt down and change every absolute-path `cfopen()` call in this PR makes me want to start crying and vomiting, especially when I look back at how unsupportive the Descent 3 development community has been. I do think that the `cfopen()` change is worth making, but I think that it should be made in a separate PR.

---

> Only thing is that we should get away sometime from `Base_directory` to list of base directories (readonly and writable).

It sounds like you’re saying that you agree with the changes in this PR. Specifically, it sounds like you agree with the parts of this PR that replace `char Base_directory[_MAX_PATH]` with `std::vector<std::filesystem::path> Base_directories`. It sounds like you also agree with parts of this PR that cause the path to the writable base directory and the paths to all of the read-only base directories in the `Base_directories` variable. That being said, I don’t think that you successfully answered this question:

> When I move all of the stuff from the `gamedata` module to the `cfile` module, should the “Move Base_directory into separate module” commit be changed so that it moves `Base_directory` into the `cfile` module?

---

> I'm planning heavily rewrite cfile / ddio modules (related to file operations), I still see some optimization points here to improve overall performance. Path cache is one of them. Currently we are about on halfway of this refactoring :).
>
> Main issue of case insensitive discovering is that can't be optimized much. On each level of directories directory iterator loses so much time, making whole traversal incredibly slow.

I don’t think that you understood this question that I asked:

> So you’re saying that you might open another PR that optimizes Descent 3’s case insensitive path code further?

Here’s my motivation for asking that question. Sometimes, you mention things that could be done but haven’t been done yet. Here are some examples:

- [You said that you were working on refactroring the `cfile` and `ddio` modules.](https://github.com/DescentDevelopers/Descent3/pull/471#issuecomment-2194400627)

- [You said that it might be good to add a LRU cache or something that’s similar to an LRU cache.](https://github.com/DescentDevelopers/Descent3/pull/471#issuecomment-2324453127)

- [You mentioned that case-insensitive path lookup can’t be optimized much.](https://github.com/DescentDevelopers/Descent3/pull/471#issuecomment-2325374261)

When someone comments on my PR, I expect the comments to fall into one of four categories: approvals, disapprovals, questions about the PR or suggestions for how the PR can be improved. Some of your comments don’t fall into any of those categories so I don’t know what to do with them. [In the past, I tried asking you how one of those comments affects my PR](https://github.com/DescentDevelopers/Descent3/pull/471#issuecomment-2195292283), but you never responded.

---

> Optimization of `cf_FindRealFileNameCaseInsensitive()` is based in fact that `Base_directory` is already known to us, and we only have to resolve one-two sublevels of directory path.

This is really helpful. Part of the reason why I didn’t understand `cf_FindRealFileNameCaseInsensitive()` is because I expected it to handle an arbitrary number of subdirectories. The code makes a lot more sense now that I know that it’s not supposed to handle an arbitrary number of subdirectories on purpose.

---

> So that why I'm stressing that we should use relative paths on resource loading, this is one of ways to optimize case-insensitive mess.

If that’s the case, then I think that you’ll like the changes in this PR’s “[WIP] Consolidate case-sensitive filesystem functions” commit.

winterheart:
> I pushed a new version of this PR that resolves merge conflicts. I haven’t gotten rid of the `gamedata` module yet. I’ll start work on moving everything from `gamedata` to `cfile` once I get an answer to my question about the “Move Base_directory into separate module” commit.

`Base_directory` (or list of them) should be in cfile and should be accessed (when possible) via submodule's functions. Direct access to variable should be avoidable.

Jayman2000:
> `Base_directory` (or list of them) should be in cfile[…]

OK. I just pushed a new version of this PR that restores the code that was in `cf_FindRealFileNameCaseInsensitive()`, moves everything that was in `gamedata` to `cfile`, fixes the merge conflicts, replaces the `mprintf()` calls with `LOG_WHATEVER` and makes a few improvements to comments and commit messages.

> […]and should be accessed (when possible) via submodule's functions. Direct access to variable should be avoidable.

I feel like this PR already did that by the time that you left that comment. When you left that comment, this PR added functions named `gd_LocatePath()`, `gd_LocateMultiplePaths()` and `gd_GetWritableBaseDirectory()`, and this PR always accessed `Base_directories` through those functions (with only one exception). The new version of this PR does the same thing, but those functions have been renamed to `cf_LocatePath()`, `cf_LocateMultiplePaths()` and `cf_GetWritableBaseDirectory()`.

What do you think? Do you think that the way that this PR uses those functions counts as avoiding direct access to the `Base_directories` variable?

cfile/cfile.cpp
```diff
@@ -69,6 +79,112 @@ cfile_error cfe;
...
+ *
+ * @param relative_path A relative path that we’ll hopefully find in
+ *                      one of the Base_directories. You don’t have to get the
+ *                      capitalization of relative_path correct, even on macOS
+ *                      and Linux.
+ *
+ * @return Either an absolute path that’s inside a base directory or an empty
+ *         path if nothing is found.
+ */
+std::filesystem::path cf_LocatePath(std::filesystem::path relative_path) {
```
winterheart, line 157: add const qualifier to arg.

cfile/cfile.cpp
```diff
@@ -69,6 +79,112 @@ cfile_error cfe;
...
+ *
+ * @param relative_path A relative path that we’ll hopefully find in
+ *                      one or more of the Base_directories. You don’t have to
+ *                      get the capitalization of relative_path correct, even on
+ *                      macOS and Linux.
+ *
+ * @return A list of absolute paths. Each path will be inside one of the
+ *         Base_directories.
+ */
+std::vector<std::filesystem::path> cf_LocateMultiplePaths(std::filesystem::path relative_path) {
```
winterheart, line 177: add const qualifier to arg.

cfile/cfile.cpp
```diff
@@ -69,6 +79,112 @@ cfile_error cfe;
...
+ *
+ * @return Either an absolute path that’s inside a base directory or an empty
+ *         path if nothing is found.
+ */
+std::filesystem::path cf_LocatePath(std::filesystem::path relative_path) {
+  auto return_value_list = cf_LocatePathMultiplePathsHelper(relative_path, true);
+  if (return_value_list.empty()) {
+    return "";
+  } else {
+    return return_value_list[0];
```
winterheart, line 162: Use `.front()` instead accessing by index.

Descent3/d3movie.cpp
```diff
@@ -357,17 +320,7 @@ void CallbackShowFrameNoFlip(unsigned char *buf, unsigned int bufw, unsigned int
...
-  real_name = mve_FindMovieFileRealName(mvename);
-  if (real_name.empty()) {
-    LOG_WARNING.printf("MOVIE: No such file %s", mvename);
-    fhandle = nullptr;
-    return 0;
-  }
-#else
-  real_name = mvename;
-#endif
+  std::filesystem::path real_name = cf_LocatePath(std::filesystem::path("movies") / mvename);
```
winterheart, line 323: Probably `mvename` should be inside `cf_LocatePath()`? It can be non-lowercased.

Descent3/gamespy.cpp
```diff
@@ -124,7 +124,7 @@ int gspy_Init() {
   }
 
   // Read the config, resolve the name if needed and setup the server addresses
-  cfgpath = std::filesystem::path(Base_directory) / gspy_cfgfilename;
+  cfgpath = cf_GetWritableBaseDirectory() / gspy_cfgfilename;
```
winterheart, line 127: gamespy file can be not only in write-path

Descent3/init.cpp
```diff
@@ -1406,16 +1404,30 @@ void InitIOSystems(bool editor) {
...
+
+  ddio_SetWorkingDir(writable_base_directory.u8string().c_str());
+  initial_base_directories.insert(initial_base_directories.begin(), writable_base_directory);
+
+  // Set any additional base directories
+  auto additionaldirarg = 0;
+  while (0 != (additionaldirarg = FindArg("-additionaldir", additionaldirarg))) {
+    const auto dir_to_add = GetArg(additionaldirarg + 1);
+    if (dir_to_add == NULL) {
+      LOG_WARNING << "-additionaldir was at the end of the argument list. It should never be at the end of the argument list.";
```
winterheart, line 1422: Probably this should means that user passed `-additionaldir` without a value?

Descent3/init.cpp
```diff
@@ -1406,16 +1404,30 @@ void InitIOSystems(bool editor) {
...
+      writable_base_directory = executablePath.parent_path();
+      LOG_INFO << "Using working directory of " << writable_base_directory.u8string().c_str();
     }
   } else {
-    ddio_GetWorkingDir(Base_directory, sizeof(Base_directory));
+    writable_base_directory = std::filesystem::current_path();
+  }
+
+  ddio_SetWorkingDir(writable_base_directory.u8string().c_str());
+  initial_base_directories.insert(initial_base_directories.begin(), writable_base_directory);
```
winterheart, line 1415: Use `.push_back()` on std::vector filling, this is more fast than `.insert()`.

Descent3/multi_dll_mgr.cpp
```diff
@@ -595,7 +598,7 @@ int LoadMultiDLL(const char *name) {
   if (MultiDLLHandle.handle)
     FreeMultiDLL();
 
-  std::filesystem::path dll_path_name = std::filesystem::path(Base_directory) / "online";
+  std::filesystem::path dll_path_name = cf_GetWritableBaseDirectory() / "online";
```
winterheart, line 601: netcon modules can be placed not only in write-path.

ddio/sdlcontroller.cpp
```diff
@@ -1401,7 +1401,7 @@ int CTLLex(const char *command) {
 void sdlgameController::parse_ctl_file(int devnum, const char *ctlname) {
   // parse each file until we find a name match, no name match, just return
   ddio_DoForeachFile(
-      Base_directory, std::regex(".*\\.ctl"), [this, &devnum, &ctlname](const std::filesystem::path &path) {
+      cf_GetWritableBaseDirectory(), std::regex(".*\\.ctl"), [this, &devnum, &ctlname](const std::filesystem::path &path) {
```
winterheart, line 1404: Files can be not only in write-path.

netcon/includes/con_dll.h
```diff
@@ -768,7 +777,7 @@ int DLLGame_mode;
 char *DLLTracker_id;
 int *DLLNum_directplay_games;
 netgame_info *DLLNetgame;
-char *DLLLocalD3Dir;
+std::vector<std::filesystem::path> *DLLBase_directories;
```
winterheart, line 780: This variable should not be exported and accessible directly to netcon modules.

netcon/includes/mdllinit.h
```diff
@@ -286,7 +289,7 @@ DLLTracker_id = (char *)API.vp[1];
 DLLGame_is_master_tracker_game = API.vp[2];
 DLLGame_mode = *API.vp[3];
 // DLLCurrent_pilot = (pilot *)API.vp[4];
-DLLLocalD3Dir = (char *)(pilot *)API.vp[5];
+DLLBase_directories = (std::vector<std::filesystem::path> *)API.vp[5];
```
winterheart, line 292: This variable should not be exported and accessible directly to netcon modules.

Descent3/init.cpp
```diff
@@ -1406,16 +1404,30 @@ void InitIOSystems(bool editor) {
       Error("Failed to get executable path\n");
     } else {
       std::filesystem::path executablePath(exec_path);
-      std::string baseDirectoryString = executablePath.parent_path().string();
-      strncpy(Base_directory, baseDirectoryString.c_str(), sizeof(Base_directory) - 1);
-      Base_directory[sizeof(Base_directory) - 1] = '\0';
-      LOG_INFO << "Using working directory of " << Base_directory;
+      writable_base_directory = executablePath.parent_path();
+      LOG_INFO << "Using working directory of " << writable_base_directory.u8string().c_str();
```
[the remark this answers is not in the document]
Jayman2000: OK. I removed the `.u8string().c_str()` part.

Descent3/multi_dll_mgr.cpp
```diff
@@ -518,14 +518,17 @@ void GetMultiAPI(multi_api *api) {
...
+  api->fp[114] = (int *)cf_GetWritableBaseDirectory;
 
   // Variable pointers
   api->vp[0] = (int *)&Player_num;
   api->vp[1] = (int *)Tracker_id;
   api->vp[2] = (int *)&Game_is_master_tracker_game;
   api->vp[3] = (int *)&Game_mode;
   api->vp[4] = (int *)NULL; // Current_pilot; no longer a struct
-  api->vp[5] = (int *)Base_directory;
+  api->vp[5] = (int *)&Base_directories;
```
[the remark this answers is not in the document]
Jayman2000: I just pushed a new version of this PR that changes that line to this:

```cpp
  api->vp[5] = (int *)NULL; // Base_directory; no longer exists
```

cfile/cfile.cpp
```diff
@@ -69,6 +79,112 @@ cfile_error cfe;
 // The message for unexpected end of file
 const char *eof_error = "Unexpected end of file";
 
+// This function must be called before you use anything else from this module.
+void cf_Init(std::vector<std::filesystem::path> initial_base_directories) {
+  Base_directories = initial_base_directories;
+}
```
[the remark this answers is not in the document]
Jayman2000: I made the requested change. I decided to call the function `cf_AddBaseDirectory()` because I really dislike the name `cf_AddSearchPath()`. The variable is called `Base_directories`, not `Search_paths`, so the name `cf_AddBaseDirectory()` is more clear. Additionally, there’s already two functions that are declared in the same file that are named `cf_SetSearchPath()` and `cf_ClearAllSearchPaths()`, and those two functions don’t change the `Base_directories` variable. It would be really confusing if `cf_AddSearchPath()` affected `Base_directories` but `cf_SetSearchPath()` did not.

cfile/cfile.cpp
```diff
@@ -69,6 +79,112 @@ cfile_error cfe;
...
+
+#ifdef WIN32
+std::filesystem::path cf_LocatePathCaseInsensitiveHelper(const std::filesystem::path &relative_path, const std::filesystem::path &starting_dir) {
+  return starting_dir / relative_path;
+}
+#else
```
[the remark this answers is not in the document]
Jayman2000: OK, done.

cfile/cfile.cpp
```diff
@@ -69,6 +79,112 @@ cfile_error cfe;
...
+    // Component not found, mission failed
+    return {};
+  }
+
+  return result;
+}
+#endif
+
+std::vector<std::filesystem::path> cf_LocatePathMultiplePathsHelper(std::filesystem::path relative_path, bool stop_after_first_result) {
+  ASSERT(("realative_path should be a relative path.", relative_path.is_relative()));
```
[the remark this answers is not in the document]
Jayman2000: I don’t understand how I could replace the `ASSERT`s with code that uses `std::filesystem::absolute()`. Could you post a diff showing what you would do here?

cfile/cfile.cpp
```diff
@@ -69,6 +79,112 @@ cfile_error cfe;
...
+  std::vector<std::filesystem::path> return_value = { };
+  for (auto base_directory : Base_directories) {
+    ASSERT(("base_directory should be an absolute path.", base_directory.is_absolute()));
+    auto to_append = cf_LocatePathCaseInsensitiveHelper(relative_path, base_directory);
+    ASSERT(("to_append should be either empty or an absolute path.", to_append.empty() || to_append.is_absolute()));
```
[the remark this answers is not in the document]
Jayman2000: I strongly object to this. If `cf_LocatePathCaseInsensitiveHelper()` ever returns a non-empty relative path, then there’s a serious bug with `cf_LocatePathCaseInsensitiveHelper()`. I would rather find out about the bug sooner rather than later, and this `ASSERT` will help us find the bug sooner rather than later. I’ve already made mistakes like that while creating this PR, and that `ASSERT` helped me find and fix my mistakes.

cfile/cfile.cpp
```diff
@@ -69,6 +79,112 @@ cfile_error cfe;
...
+  } else {
+    // Component not found, mission failed
+    return {};
+  }
+
+  return result;
+}
+#endif
+
+std::vector<std::filesystem::path> cf_LocatePathMultiplePathsHelper(std::filesystem::path relative_path, bool stop_after_first_result) {
```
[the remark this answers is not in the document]
Jayman2000: OK, done.

cfile/cfile.cpp
```diff
@@ -69,6 +79,112 @@ cfile_error cfe;
...
+  ASSERT(("realative_path should be a relative path.", relative_path.is_relative()));
+  std::vector<std::filesystem::path> return_value = { };
+  for (auto base_directory : Base_directories) {
```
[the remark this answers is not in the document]
Jayman2000: In order to switch to `.push_back()` like you asked in another review comment, I had to change that for loop so that it iterates over `Base_directories` backwards. What do you think of that for loop now?

cfile/cfile.cpp
Jayman2000: Done.

cfile/cfile.cpp
Jayman2000: Done.

cfile/cfile.cpp
Jayman2000: Done.

Descent3/d3movie.cpp
Jayman2000: I’m confused. To me, it looks like `mvename` is inside of the `cf_LocatePath()` call. Could you post a diff of what you’re suggesting here?

Descent3/gamespy.cpp
Jayman2000: OK. I changed that line to this:

```cpp
  cfgpath = cf_LocatePath(gspy_cfgfilename);
```

Descent3/init.cpp
Jayman2000: So, you’re saying that it would be better if the error message said something along the lives of “There was no value for the final -additionaldir option”?

Descent3/init.cpp
Jayman2000: OK. I changed it so that if fills `Base_directories` with `.push_back()` and then iterates over `Base_directories` backwards instead of forwards.

Descent3/multi_dll_mgr.cpp
Jayman2000: I agree, netcon modules can indeed be placed in read-only base directories. I’ve tested my code extensively, and I can confirm that my code works properly if you put the `online` directory inside one of the read-only base directories.

The line of code that you’re commenting on is part of a chunk of code that deletes temporary files. I think that it’s OK to delete temporary files that are in the writable base directory, but I don’t think that it’s OK to delete temporary files that are in any of the read-only base directories. It would be really confusing if we decided to call a list of directories “read-only” but then tried to write to those directories.

ddio/sdlcontroller.cpp
Jayman2000: OK. I changed that code so that it looks for `.ctl` files in each of the base directories. In order to do so, I added code to `ddio/sdlcontroller.cpp` that accesses the `Base_directories` variable directly. Is that OK?

netcon/includes/con_dll.h
Jayman2000: OK. I removed that variable.

netcon/includes/mdllinit.h
Jayman2000: OK. I removed that variable.

Descent3/multi_dll_mgr.cpp
winterheart: Seems to be here old logic left after switching to use `Descent3_temp_directory`. This part of code now does nothing useful, so we can skip this request.

ddio/sdlcontroller.cpp
winterheart: Yes, that OK.

winterheart (approved):

winterheart:
Tested on Linux and Windows, and generally changes works as intended. Thanks for patience and contribution.

Lgt2x:
Thanks for this improvement, big step forward for file management!

PR #474, opened 2024-06-28
Enable all compilation warning on unix platforms

Lgt2x:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

Enable `-Wall` on Unix compilers, in order to catch more problems in the code. Cleanup global compile options, that referred to unset variables

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -90,9 +90,8 @@ install(
   DESTINATION ${CMAKE_INSTALL_DOCDIR}
 )
 
-if(UNIX)
-  add_compile_options("$<$<COMPILE_LANGUAGE:CXX>:-Wno-multichar;${BITS};${EXTRA_CXX_FLAGS}>")
-  add_compile_options("$<$<COMPILE_LANGUAGE:C>:${BITS}>")
+if(CMAKE_CXX_COMPILER_ID MATCHES "Clang" OR CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+  add_compile_options("-Wno-multichar;-Wall")
 endif()
 
 find_package(SDL2 REQUIRED)


winterheart:
Need resolve conflicts, otherwise is good to merge.

Lgt2x:
rebased

CMakeLists.txt
```diff
@@ -91,8 +91,7 @@ install(
 )
 
 if(UNIX)
-  add_compile_options("$<$<COMPILE_LANGUAGE:CXX>:-Wno-multichar;${BITS};${EXTRA_CXX_FLAGS}>")
-  add_compile_options("$<$<COMPILE_LANGUAGE:C>:${BITS}>")
+  add_compile_options("-Wno-multichar;-Wall")
 endif()
```
winterheart, line 95: On second thought, I think it's better to use `CMAKE_<LANG>_COMPILER_ID` check instead of platform check, like

```
if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang" OR CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
  add_compile_options(-Wno-multichar -Wall)
elseif()
```

Lgt2x: does "Clang" also cover "AppleClang" ?

Lgt2x: cf https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER_ID.html

winterheart: Probably not, better to use `MATCHES` instead `STREQUAL` then.

Lgt2x: done

winterheart:
Ready to merge.

PR #475, opened 2024-06-29
cfile / ddio: use most of functions to std::filesystem::path

winterheart:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

This PR updates cfile and partially ddio modules to use C++17's `std::filesystem::path` feature, which simplifies overall filesystem operations.

Some parts of code was migrated, and there some performance improvements was achieved. Internally cfile maintains some sort search paths. Old code was excessively added some non-existent paths, and on opening files these paths was pointlessly used. Also there some code optimization in `cf_FindRealFileNameCaseInsenstive()` was performed.

Added unittest framework for cfile.

### Checklist


Lgt2x:
You can do something like that to fix mac os CI (but cleaner). `strncpy` is problematic https://github.com/Lgt2x/Descent3/commit/15a50ca126b7da7504f6f2328bd6622d65463b71#diff-be34739991a103c73033e6358b31b2a8d7859b7a5c84616dcf1409176168f55d

winterheart:
After some refactoring cf_FindRealFileNameCaseInsensitive() now works much faster and more reliable. Here some benchmarks:

Release:
```
2024-07-02T20:08:56+03:00
Running ./cfile_bench
Run on (12 X 4300 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x6)
  L1 Instruction 32 KiB (x6)
  L2 Unified 256 KiB (x6)
  L3 Unified 12288 KiB (x1)
Load Average: 2.82, 1.81, 1.42
***WARNING*** CPU scaling is enabled, the benchmark real time measurements may be noisy and will incur extra overhead.
-------------------------------------------------------------------
Benchmark                         Time             CPU   Iterations
-------------------------------------------------------------------
cfile_no_exist_old            11384 ns        11065 ns        63154
cfile_no_exist_new             9812 ns         9613 ns        74990
cfile_exist_relative_old       2682 ns         2452 ns       258570
cfile_exist_relative_new        926 ns          925 ns       784549
cfile_exist_absolute_old       3422 ns         3132 ns       222331
cfile_exist_absolute_new       1222 ns         1220 ns       585813
```
Debug:
```
2024-07-02T20:09:47+03:00
Running ./cfile_bench
Run on (12 X 4300 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x6)
  L1 Instruction 32 KiB (x6)
  L2 Unified 256 KiB (x6)
  L3 Unified 12288 KiB (x1)
Load Average: 2.21, 1.81, 1.45
***WARNING*** CPU scaling is enabled, the benchmark real time measurements may be noisy and will incur extra overhead.
-------------------------------------------------------------------
Benchmark                         Time             CPU   Iterations
-------------------------------------------------------------------
cfile_no_exist_old            11768 ns        11426 ns        61555
cfile_no_exist_new            13682 ns        13423 ns        53839
cfile_exist_relative_old       3147 ns         2907 ns       224424
cfile_exist_relative_new       1591 ns         1588 ns       427936
cfile_exist_absolute_old       4179 ns         3893 ns       183101
cfile_exist_absolute_new       2132 ns         2129 ns       319614
```

cfile/cfile.cpp
```diff
@@ -394,363 +363,165 @@ CFILE *open_file_in_lib(const char *filename) {
...
+  }
+
   // set read or write mode
   tmode[0] = mode[0];
-  // if mode is "w", then open in text or binary as requested.  If "r", alsway open in "rb"
+  // if mode is "w", then open in text or binary as requested.  If "r", always open in "rb"
   tmode[1] = (mode[0] == 'w') ? mode[1] : 'b';
   // try to open file
-  fp = fopen(path, tmode);
+  fp = fopen(using_filename.u8string().c_str(), tmode);
```
Jayman2000, line 427: How do you know that UTF-8 is the correct encoding to use for `fopen()`’s `filename` parameter?

winterheart: This way how std::filesystem::path represent native encoding on systems. POSIX filesystem is native `char` representation, but Windows' native representation is `wchar`. If you want use paths with C functions, you need convert them with `c_str()` using `u8string()`.

Descent3/gamesave.cpp
```diff
@@ -395,11 +397,10 @@ void SaveGameDialog() {
   //	ddio_MakePath(pathname, savegame_dir, "*.sav", NULL); -unused
 
   // create savegame directory if it didn't exist before.
-  if (!ddio_DirExists(savegame_dir)) {
-    if (!ddio_CreateDir(savegame_dir)) {
-      DoMessageBox(TXT_ERROR, TXT_ERRCREATEDIR, MSGBOX_OK);
-      return;
-    }
+  std::error_code ec;
+  if (!std::filesystem::create_directories(savegame_dir, ec)) {
```
Lgt2x, line 401: `ec` unused, use the overload without error code (https://en.cppreference.com/w/cpp/filesystem/create_directory (1))

cfile/tests/cfile_tests.cpp
```diff
@@ -0,0 +1,123 @@
...
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+* GNU General Public License for more details.
+*
+* You should have received a copy of the GNU General Public License
+* along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include <filesystem>
+#include <vector>
+#include <gtest/gtest.h>
```
Lgt2x, line 21: I need `#include <algorithm>` to compile on my machine, please add it

Descent3/gamesave.cpp
Lgt2x, line 401: ec unused

cfile/cfile.cpp
```diff
@@ -103,42 +94,38 @@ static CFILE *open_file_in_lib(const char *filename);
...
   uint32_t offset;
   static int first_time = 1;
-  tHogHeader header;
-  tHogFileEntry entry;
+  tHogHeader header{};
+  tHogFileEntry entry{};
 
   // allocation library structure
   std::shared_ptr<library> lib = std::make_shared<library>();
 #ifdef __LINUX__
```
Lgt2x, line 107: Didnt you say Mac is also case insensitive? OSX defines `__LINUX__`

cfile/cfile.cpp
```diff
@@ -103,42 +94,38 @@ static CFILE *open_file_in_lib(const char *filename);
...
-  if (resolve_dir != nullptr)
-    ddio_MakePath(lib->name, resolve_dir, t_out, nullptr);
+  if (!resolve_dir.empty())
+    lib->name = resolve_dir / t_out;
   else
-    strcpy(lib->name, t_out);
+    lib->name = t_out;
 #else
-  strcpy(lib->name, libname);
+  lib->name = libname;
```
Lgt2x, line 126: we could have the same code path for case sensitive and case insensitive fs, if`cf_FindRealFileNameCaseInsensitive` works for both. It does not have a big performance impact

cfile/cfile.cpp
```diff
@@ -292,8 +258,9 @@ CFILE *cf_OpenFileInLibrary(const char *filename, int libhandle) {
 
   do {
     i = (first + last) / 2;
-    c = stricmp(filename, lib->entries[i]->name); // compare to current
-    if (c == 0) {                                 // found it
+    c = stricmp(filename.u8string().c_str(), lib->entries[i]->name); // compare to current
+    if (c == 0) {
+      // found it
```
Lgt2x, line 263: redundant comment

cfile/cfile.cpp
```diff
@@ -292,8 +258,9 @@ CFILE *cf_OpenFileInLibrary(const char *filename, int libhandle) {
...
       found = 1;
```
Lgt2x, line 264: could make found a `bool`

cfile/cfile.cpp
```diff
@@ -394,363 +363,165 @@ CFILE *open_file_in_lib(const char *filename) {
...
-  bool found_match = false;
+  search_path = directory / fname.parent_path();
+  search_file = fname.filename();
 
-  if ((real_file[0] >= 'a' && real_file[0] <= 'z') || (real_file[0] >= 'A' && real_file[0] <= 'Z')) {
-    // alpha first letter...we need to do 2 iterations
-    iterations = 2;
+  // If directory does not exist, nothing to search.
+  if (!is_directory(search_path) || search_file.empty()) {
+    return {};
```
Lgt2x, line 380: Personal preference for explicit default constructor  `std::filesystem::path{}`

Lgt2x (commented):
Really good job, tested successfully on Linux. Will also test Windows before merging. If someone can do OSX don't hesitate

cfile/cfile.cpp
Jayman2000: > This way how std::filesystem::path represent native encoding on systems. POSIX filesystem is native `char` representation, but Windows' native representation is `wchar`. If you want use paths with C functions, you need convert them with `c_str()` using `u8string()`.

I’m a little bit confused. [According to cppreference.com, `std::filesystem::path::u8string()` returns a `std::u8string`](https://en.cppreference.com/w/cpp/filesystem/path/string). [`std::u8string` is the same as `std::basic_string<char8_t>`.](https://en.cppreference.com/w/cpp/string/basic_string) [`std::basic_string::<whatever>::c_str()` returns a `whatever *`](https://en.cppreference.com/w/cpp/string/basic_string/c_str) which means that `my_u8string.c_str()` should return a `char8_t *`, right? It seems to me like `using_filename.u8string().c_str()` is always going to evaluate to a `char8_t *`. I just don’t see how it could evaluate to a `wchar_t *`.

(The `std::u8string` type was added in C++20. In C++17, a similar argument can be made, but you have to replace `std::u8string` with `std::string` and `char8_t` with `char`.)

Passing a UTF-8 encoded string to `fopen()` is potentially a problem. [According to Microsoft’s documentation,](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/fopen-wfopen?view=msvc-170#remarks)

> The **`fopen`** function opens the file specified by *`filename`*. By default, a narrow *`filename`* string is interpreted using the ANSI codepage (`CP_ACP`). In Windows Desktop applications, it can be changed to the OEM codepage (`CP_OEMCP`) by using the [`SetFileApisToOEM`](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-setfileapistooem) function. You can use the [`AreFileApisANSI`](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-arefileapisansi) function to determine whether *`filename`* is interpreted using the ANSI or the system default OEM codepage.

Is there anything that’s going to guarantee that we’re using a UTF-8 codepage? [Windows won’t use a UTF-8 codepage by default unless the user has “Beta: Use Unicode UTF-8 for worldwide language support” enabled, or the application explicitly specifies a UTF-8 codepage.](https://learn.microsoft.com/en-us/windows/apps/design/globalizing/use-utf8-code-page)

Additionally, if `using_filename.u8string().c_str()` sometimes does evaluate to a `wchar_t *`, then I think that we have a problem. [According to Microsoft’s documentation, you’re supposed to pass a `char *` to `fopen()`, not a `wchar_t *`. If you want to use a `wchar_t *` for the `filename` parameter, then you’re supposed to call `_wfopen()`.](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/fopen-wfopen?view=msvc-170)

Descent3/gamesave.cpp
winterheart: `ec` is unused, but this prevents throwing exceptions.

cfile/cfile.cpp
winterheart: This gray area that was discovered after unittests implementation. Still, there nothing bad happens on macos, effectively that logic works. Removed here ifdef

cfile/cfile.cpp
winterheart: We don't even interested here what encoding is actually in filesystem. We effectively receive C-string (and don't bother in what encoding is it, it's just sequence of chars) and we ensuring that C-string is actually valid path in native filesystem. We do not using UTF-8 here at all. It's just way to convert native encoding to C-string.

Jayman2000: > We don't even interested here what encoding is actually in filesystem.

If we’re going to call `fopen()` on Windows, then I think that we need to be interested in the `filename` parameter’s encoding because Windows is interested in the `filename` parameter’s encoding.

---

> We effectively receive C-string (and don't bother in what encoding is it, it's just sequence of chars) and we ensuring that C-string is actually valid path in native filesystem. We do not using UTF-8 here at all.

That last sentence is very surprising to me. I thought that `std::filesystem::path::u8string` was supposed to return UTF-8, that `std::filesystem::path::u16string` was supposed to return UTF-16 and that `std::filesystem::path::u32string` was supposed to return UTF-32. [cppreference.com says the following about those functions:](https://en.cppreference.com/w/cpp/filesystem/path/string)

> Returns the internal pathname in native pathname format, converted to specific string type. Conversion, if any, is performed as follows:
>
> […]
>
> 3\) The result encoding in the case of `u8string()` is always UTF-8.

Additionally, [Microsoft’s documentation for `std::filesystem::path::u8string` says:](https://learn.microsoft.com/en-us/cpp/standard-library/path-class?view=msvc-170#u8string)

> Converts the sequence stored in `mypath` to UTF-8 and returns it stored in an object of type `u8string`.

What makes you say “We do not using UTF-8 here at all”?

---

> It's just way to convert native encoding to C-string.

I agree. `u8string().c_str()` does indeed convert a `path` (which uses its native encoding internally) into a C-string. My concern is about the encoding of the resulting C-string.

Lgt2x (approved):

bryanperris (approved):

Lgt2x:
No problems in game in either Linux or Windows, we're good to merge. I hope we did not break anything for MacOS, but I don't have the hardware to test it. If UTF-8 proves to be a problem in some situations we can fix it later.

PR #480, opened 2024-07-02
Make sure that everything is Clang formatted

Jayman2000:
## Pull Request Type

- [x] GitHub Workflow changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

This PR does multiple `clang-format` related things:

1. It revamps `tools/formatter.sh`.

2. It fixes and re-enables the `clang-format` workflow.

3. It makes sure that all files obey the style settings in `.clang-format`.

### Related Issues

Closes #450.

### Checklist

### Additional Comments

This PR is meant to be an alternative to #450. Here are the differences between this PR and that other PR.

- The script that this PR introduces is much shorter than the two scripts that other PR introduces.
- The script that this PR introduces has less dependencies that the shell script that the other PR introduces. This PR’s script depends on `/bin/sh`, `git` and `clang-format`. The other PR’s shell script depends on `/bin/sh`, `which`, `clang-format`, `xargs`/`parallel` and `find`.
- The script that this PR introduces does not depend on `which`. The shell script that the other PR introduces depends on a version of `which` that support the `-s` flag. [GNU `which` does not support the `-s` flag.](https://carlowood.github.io/which). I don’t know how common GNU `which` is compared to other implementations of `which`, but I do know that [GNU `which` is a required package on NixOS](https://github.com/NixOS/nixpkgs/blob/00d80d13810dbfea8ab4ed1009b09100cca86ba8/nixos/modules/config/system-path.nix#L40) [and a mandatory package on Fedora Workstation](https://pagure.io/fedora-comps/blob/b58a1bbbcad3456e1d0acc38f4d24b3c9d5e280f/f/comps-f40.xml.in#_5800).
- The script that this PR introduces doesn’t bother checking whether or not its dependencies are present before trying to execute them. The shell script that the other PR introduces checks for its dependencies before trying to execute them. As far as I know, there’s no real consequences if this PR’s script gets run by someone who doesn’t have `git` or `clang-format`, so checking to see if they’re available is not necessary.
- This PR introduce a single shell script. The other PR introduces a shell script and a batch script. Windows users can just run the shell script in Git Bash. After all, building Descent 3 already depends on Git, and [Git depends on “A POSIX-compliant shell”](https://git.kernel.org/pub/scm/git/git.git/tree/INSTALL?h=v2.44.2#n118).

> [!IMPORTANT]  
> I’m not so sure that this PR should actually be merged. I have a concern that I wrote about [here](https://github.com/DescentDevelopers/Descent3/pull/450#issuecomment-2204526022). Depending on how we decide to resolve that concern, I might end up canceling this PR.


Lgt2x:
Closing, see comments in #450

PR #482, opened 2024-07-03
Better project structure for IDEs

pzychotic:
Enabled the CMake property USE_FOLDERS to provide a more structured project list in IDEs with the help of set(CMAKE_FOLDER "...") calls.

## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

When using CMake to generate project files for an IDE like VS, the CMake property `USE_FOLDERS` enables the possibility to group targets into folders specified with `set(CMAKE_FOLDER "...")` calls. This allows to get some structure inside the IDE instead of just having an endless long list of targets on the root level, see screenshot.

Screenshot 2024-07-03 200447

### Checklist


Lgt2x (approved):
I did not know about this CMake property, but no problem to enable that if it makes it easier for VS users

PR #485, opened 2024-07-08
Put core campaign missions at top, use listbox for level select

JeodC:
## Pull Request Type

- [ ] GitHub Workflow changes
- [x] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [x] Input changes
  - [ ] Network changes
  - [x] Other changes

### Description

This PR can be treated as a proof-of-concept, since it may not be desired on PC. I made these changes in fork to provide accessibility for controller-only setups, but I like the look of it better than the current "type a number" setup. It does the following:

- Puts Retribution and Mercenary at the top of the mission list
- Changes the old level warp dialog which requires keyboard input for a listbox of the number of levels in the mission
- - The mission name is titled at the top of the level select listbox to inform the player which mission they're choosing levels from
- - Level names are retrieved from the level file, but the core mission levels don't have names inside, so they use an array in `menu.h`
- If a mission is flagged as `finished` in the pilot file, shows all levels like Debug mode, which allows selecting secret levels
- - Tested locally with Mercenary, where a new pilot loaded a savegame for the final level and completed it

image

### Checklist


Descent3/menu.cpp
```diff
@@ -1398,6 +1455,76 @@ int DisplayLevelWarpDlg(int max_level) {
   hwnd.Destroy();
   return chosen_level;
 }
+
+// DisplayLevelSelectDlg
```
JeodC, line 1459: Unless the `DisplayLevelWarpDlg` function can be used elsewhere, it could be removed entirely since it is now unused. I haven't tested the cheat to warp while in a level though.

JeodC:
The level select box is currently the same dimensions as the mission list box, but for some level strings this can overflow. See screenshot.

image

I personally don't think it's important enough to warrant changing the dialog box dimensions, since they match the dimensions of the mission diaog, and the `char` for the strings also match the length allowed for the mission names.

Descent3/menu.cpp
```diff
@@ -1398,6 +1455,76 @@ int DisplayLevelWarpDlg(int max_level) {
...
+// displays a list of levels associated with the selected mission
+int DisplayLevelSelectDlg(int max_level) {
```
winterheart, line 1461: Probably you can add campaign name as second argument, so you don't have to create global variable `msnname`.

Descent3/menu.cpp
```diff
@@ -1398,6 +1455,76 @@ int DisplayLevelWarpDlg(int max_level) {
...
+  // Loop through the levels
+  for (int level = 1; level <= max_level; ++level) {
+    // Check if msnname matches core mission names, if so use the arrays for level names
+    // Otherwise, try to get the level name from the level info
+    if (strcmp(msnname, "Descent 3: Retribution") == 0) {
+      if (level >= 1 && level <= sizeof(RetributionNames) / sizeof(RetributionNames[0])) {
+        if (level > 15) { // If they're the secret levels we use a different format
+          snprintf(level_name, sizeof(level_name), "Secret Lv - %s", RetributionNames[level - 1]);
+        } else {
+          snprintf(level_name, sizeof(level_name), "Lv %02d - %s", level, RetributionNames[level - 1]);
```
winterheart, line 1489: Better to avoid `snprintf()` usage here, just place "Lv <num> - <name>" or "Secret Lv - <name>" into mission names lists, and this code will be simpler.

Descent3/menu.cpp
```diff
@@ -1398,6 +1455,76 @@ int DisplayLevelWarpDlg(int max_level) {
...
+  level_lb = select_sheet->AddListBox(352, 256, UID_LVLB, UILB_NOSORT);
+  select_sheet->NewGroup(NULL, 160, 280, NEWUI_ALIGN_HORIZ);
+  select_sheet->AddButton(TXT_OK, UID_OK);
+  select_sheet->AddButton(TXT_CANCEL, UID_CANCEL);
+
+  // Loop through the levels
+  for (int level = 1; level <= max_level; ++level) {
+    // Check if msnname matches core mission names, if so use the arrays for level names
+    // Otherwise, try to get the level name from the level info
+    if (strcmp(msnname, "Descent 3: Retribution") == 0) {
```
winterheart, line 1484: Maybe there can be more elegant way to determine on what campaign we choose?

Descent3/menu.cpp
JeodC: I'm not going to attempt that, since I don't have experience with the newer standard, and there are other loops in this file that are similar. If you want to simplify the entire menu.cpp file this way, by all means feel free to merge this pr into a new one that actually cleans up the whole file.

JeodC: I looked into this some more, and while I can add Lv## to the array easily to simplify this, the problem comes when the array is not used and we then need snprintf for the level number. I think this is easier. If you refactor the whole menu.cpp file in the future, this could certainly qualify for a pass.

Descent3/menu.cpp
JeodC: Sure, if we ever need the mission name in other files we can declare it earlier and make it a global option.

JeodC:
I'm going to close this, seems quite a bit has changed and I don't have time to look into it for correction. Feel free to use this as a point of reference though.

PR #488, opened 2024-07-10
Fix crash with -nointro

pzychotic:
## Pull Request Type

- [ ] GitHub Workflow changes
- [x] Runtime changes
  - [ ] Render changes
  - [x] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

When `-no-intro` is used on the command line, the game will crash in https://github.com/DescentDevelopers/Descent3/blob/3c66b13200e929835870cc9445e774cd0742a798/libmve/mveplay.cpp#L327 because the `mve_setpalette` callback was never assigned.

When intros are played, we arrive in `mve_PlayMovie()` which sets the callback with `MVE_palCallbacks(CallbackSetPalette)`. Without intros, we arrive in `mve_SequenceStart()` which doesn't call `MVE_palCallbacks(...)`.

This change will assign the missing callback before we start playing a video.

### Related Issues

#289 introduced the problem

### Checklist


Lgt2x (approved):
Good catch, thanks!

PR #494, opened 2024-07-17
Ensure UTF-8 is used in more places

### LINKED ISSUE #483: [Runtime Issue]: Unable to set temporary directory to a path that contains Japanese characters
### Build Version

v1.5 de75a80ae7564563961c4635ef839fce70944a4c

- [X] Microsoft Windows (64-bit)
- [ ] Mac OS X

### CPU Environment

- [ ] x86 (32-bit Intel/AMD)
- [ ] ARM (32-bit)
- [ ] Other (RISC V, PPC...)

### Game Modes Affected

- [X] Single player
- [X] Anarchy
- [X] Hyper-Anarchy
- [X] Robo-Anarchy
- [X] Team Anarchy
- [X] Capture the Flag
- [X] Bounty
- [X] Entropy
- [X] Hoard
- [X] Monsterball
- [X] Cooperative

### Game Environment

I’m using game data from the US Windows Dual-Jewel version of Descent 3 v1.4.

### Description

If `Descent3.exe` is stored in a directory that has Japanese characters in its name, then it will immediately crash if you try launching it:

Descent 3 Error&NewLine;Error: Unable to set temporary directory to: "D:\VC\Partially mine\Descent3\Encoding bug test\ディセント3\custom\cache"&NewLine;&NewLine;&NewLine;Press OK to exit, CANCEL to ignore this error and continue.

If you press “Cancel”, then the game will try to continue, but it will immediately give you another similar error. That second error does not give you an option to continue.

You can workaround this problem by enabling [Windows’s “Beta: Use Unicode UTF-8 for worldwide language support” setting](https://learn.microsoft.com/en-us/windows/apps/design/globalizing/use-utf8-code-page#set-a-process-code-page-to-utf-8).

### Regression Status

Last commit that doesn’t have this problem: 9263055e730c381b7b2fb9fb32d7cea439daddc2
First commit that does have this problem: 82084e9176bdc36d2c88ec7caab1a0ed69769c25

### Steps to Reproduce

1. Set your system locale to a non-Unicode Japanese locale.

    1. Go to Windows **Settings > Time & language > Language & region > Administrative language settings > Change system locale**.

    2. Make sure that “Beta: Use Unicode UTF-8 for worldwide language support” is unchecked.

    3. In the “Current system locale” dropdown box, select “Japanese (Japan)”.

    4. Click “OK”.

    5. Reboot.

2. Follow [the README’s build instructions](https://github.com/DescentDevelopers/Descent3#building---windows).

3. Create a `D3-open-source` folder with the required files and folders (See [the README’s usage section](https://github.com/DescentDevelopers/Descent3#usage) for details).

4. Try running `Descent3.exe`. You should be able to get to the title screen without any errors.

5. Rename the `D3-open-source` folder to `ディセント3`.

6. Try running `Descent3.exe`. Before the Dolby ad starts, you’ll get an error about not being able to set the working directory to `ディセント3\custom\cache`.

Jayman2000:
## Pull Request Type

- [ ] GitHub Workflow changes
- [x] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [x] Other changes

### Description

At the moment, Descent 3 doesn’t always specify which character encoding should be used. As a result, the default character encoding gets used. The default character encoding can change depending on what OS you’re using and how the OS is configured. This PR makes Descent 3 use UTF-8 instead of the default character encoding in some situations. The idea is that if we use UTF-8 everywhere, then users won’t have to change their system configuration or avoid using certain characters in order to make Descent 3 work properly.

### Related Issues

Fixes #483.

### Checklist

### Additional Comments

@winterheart [had a different idea for how to fix #483](https://github.com/DescentDevelopers/Descent3/issues/483#issuecomment-2225622869). I think that it would be valuable to implement both the fix proposed by this PR and the fix proposed by winterheart. The changes in this PR affect a lot more than just `fopen()` which is why I think that they are valuable. Calling `_wfopen()` instead of `fopen()` on Windows would still be valuable even if this PR gets merged because if would be more efficient. Specifically, if you call `fopen()`, then the `filename` parameter has to be converted to UTF-16. If you call `_wfopen()`, then the `filename` parameter doesn’t have to be converted to UTF-16 because it’s already in UTF-16.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -5,6 +5,7 @@ indent_style = space
 indent_size = 2
 tab_width = 8
 end_of_line = lf
+charset = utf-8
 spelling_language = en-US
 
 [*.md]
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,6 +28,12 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
 set_property(GLOBAL PROPERTY USE_FOLDERS ON)
 
+if(MSVC)
+  add_compile_options(/source-charset:UTF-8)
+else()
+  add_compile_options(-finput-charset=UTF-8)
+endif()
+
 if(FORCE_COLORED_OUTPUT)
   if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.24)
     set(CMAKE_COLOR_DIAGNOSTICS ON)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,9 +29,18 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 set_property(GLOBAL PROPERTY USE_FOLDERS ON)
 
 if(MSVC)
-  add_compile_options(/source-charset:UTF-8)
+  add_compile_options(/source-charset:UTF-8 /execution-charset:UTF-8)
 else()
   add_compile_options(-finput-charset=UTF-8)
+  # Unfortunately, Clang doesn’t support -fexec-charset yet so this next part
+  # is GCC only. Luckily, Clang defaults to using UTF-8 for the execution
+  # character set [1], so we’re fine. Once Clang gets support for
+  # -fexec-charset, we should probably start using it.
+  #
+  # [1]: <https://discourse.llvm.org/t/rfc-enabling-fexec-charset-support-to-llvm-and-clang-reposting/71512>
+  if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+    add_compile_options(-fexec-charset=UTF-8)
+  endif()
 endif()
 
 if(FORCE_COLORED_OUTPUT)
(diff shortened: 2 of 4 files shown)


pzychotic:
Confirmed working on a German Win 11 with D3 running out of a Japanese named directory.

I had no idea that a simple manifest entry can switch the ANSI WinApi functions to suddenly except UTF-8. No need to muck around with UTF-16 or other wchar_t based APIs looks like a big win for cross-platform projects.

Lgt2x:
How does the manifest file work? Does it have to be next to the executable? Should we install & package it?

pzychotic:
> How does the manifest file work? Does it have to be next to the executable? Should we install & package it?

It will be embedded into the executable, so nothing to do for install.

Lgt2x (approved):
Really nice indeed, I did not know that CMake could treat manifest files as sources to embed them into the executable. 

>  Calling _wfopen() instead of fopen() on Windows would still be valuable even if this PR gets merged because if would be more efficient

With the relatively low number of files the game opens, I don't think this is really performance-critical

PR #439, opened 2024-06-16
Add 1920x1080 resolution supoort

jopadan:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Documentation or Wiki changes
- [ ] Build and Dependency changes
- [x] Runtime changes
  - [x] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

Add 1920x1080 resolution support affecting #205

### Checklist

- [x] I have tested my changes locally and verified that they work as intended.
- [x] I have documented any new or modified functionality.
- [x] I have reviewed the changes to ensure they do not introduce any unnecessary complexity or duplicate code.
- [x] I understand that by submitting this pull request, I am agreeing to license my contributions under the project's license.

PR #497, opened 2024-07-19
Eliminate glBegin/glEnd usage, unify color calculation, cleanup dead code

tophyr:
## Pull Request Type

- [ ] GitHub Workflow changes
- [x] Runtime changes
  - [x] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description
Eliminates `glBegin` and `glEnd` calls, bringing us a step closer to eliminating Immediate-Mode OpenGL. In the process, color calculations are unified into one method (which does still need further simplification) and a bunch of dead code is removed.

### Checklist

### DIFF
--- a/renderer/HardwareBaseGPU.cpp
+++ b/renderer/HardwareBaseGPU.cpp
@@ -517,7 +517,7 @@ void rend_DrawPolygon2D(int handle, g3Point **p, int nv) {
   gpu_RenderPolygon(&vArray[0], nv);
 }
 
-static color_array DeterminePointColor(g3Point const* pnt) {
+color_array DeterminePointColor(g3Point const* pnt) {
   auto alpha = gpu_Alpha_multiplier * gpu_Alpha_factor;
   if (gpu_state.cur_alpha_type & ATF_VERTEX) {
     alpha *= pnt->p3_a;
@@ -528,13 +528,10 @@ static color_array DeterminePointColor(g3Point const* pnt) {
     if (gpu_state.cur_light_state == LS_FLAT_GOURAUD) {
       return {GR_COLOR_RED(gpu_state.cur_color) / 255.0f, GR_COLOR_GREEN(gpu_state.cur_color) / 255.0f,
               GR_COLOR_BLUE(gpu_state.cur_color) / 255.0f, alpha};
+    } else if (gpu_state.cur_color_model == CM_MONO) {
+      return {pnt->p3_l, pnt->p3_l, pnt->p3_l, alpha};
     } else {
-      // Do lighting based on intesity (MONO) or colored (RGB)
-      if (gpu_state.cur_color_model == CM_MONO) {
-        return {pnt->p3_l, pnt->p3_l, pnt->p3_l, alpha};
-      } else {
-        return {pnt->p3_r, pnt->p3_g, pnt->p3_b, alpha};
-      }
+      return {pnt->p3_r, pnt->p3_g, pnt->p3_b, alpha};
     }
   } else {
     return {1, 1, 1, alpha};
--- a/renderer/HardwareBaseGPU.cpp
+++ b/renderer/HardwareBaseGPU.cpp
@@ -574,39 +574,6 @@ void rend_DrawPolygon3D(int handle, g3Point **p, int nv, int map_type) {
     // all points should be original
     ASSERT(pnt->p3_flags & PF_ORIGPOINT);
 
-    ////////////////////////////////////////////
-    if (pnt->p3_flags & PF_ORIGPOINT) {
-      if (!(pnt->p3_flags & PF_PROJECTED)) {
-        g3_ProjectPoint(pnt);
-      }
-
-      // get the original point
-      float origPoint[4];
-      origPoint[0] = pnt->p3_vecPreRot.x;
-      origPoint[1] = pnt->p3_vecPreRot.y;
-      origPoint[2] = pnt->p3_vecPreRot.z;
-      origPoint[3] = 1.0f;
-
-      // transform by the full transform
-      float view[4];
-      g3_TransformVert(view, origPoint, gTransformFull);
-
-      vector tempv = pnt->p3_vecPreRot - View_position;
-      vector testPt = tempv * Unscaled_matrix;
-
-      float screenX = pnt->p3_sx + gpu_state.clip_x1;
-      float screenY = pnt->p3_sy + gpu_state.clip_y1;
-
-      // normalize
-      float oOW = 1.0f / view[3];
-      view[0] *= oOW;
-      view[1] *= oOW;
-      view[2] *= oOW;
-
-      oOW *= 1.0f;
-    }
-    ////////////////////////////////////////////
-
     if (gpu_state.cur_alpha_type & ATF_VERTEX) {
       alpha = pnt->p3_a * gpu_Alpha_multiplier * gpu_Alpha_factor;
     }
--- a/renderer/HardwareBaseGPU.cpp
+++ b/renderer/HardwareBaseGPU.cpp
@@ -450,18 +450,8 @@ void rend_DrawPolygon2D(int handle, g3Point **p, int nv) {
   int xAdd = gpu_state.clip_x1;
   int yAdd = gpu_state.clip_y1;
 
-  float fr, fg, fb;
-  if (gpu_state.cur_light_state == LS_FLAT_GOURAUD || gpu_state.cur_texture_quality == 0) {
-    float scale = 1.0f / 255.0f;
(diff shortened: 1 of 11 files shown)


winterheart:
Is there any noticeable visible changes in the rendered picture?

tophyr:
> Is there any noticeable visible changes in the rendered picture?

No, there should be absolutely no difference in the rendered result.

Lgt2x:
No noticeable difference in game on a first look. I'll try diving into the code this week-end. However, I'm not very confident pushing these rendering changes into the main branch before our first release, as we don't have the whole rendering pipeline figured out. How about merging these into another target branch `rendering` that we'll merge into main after our first "stable" release? I hope the release to come in the next month or so.

tophyr:
> How about merging these into another target branch rendering that we'll merge into main after our first "stable" release?

Seems reasonable. It's a bit out of scope for this PR, but in terms of branching and "keep stable stable but let devs dev" strategies, I'm more used to `main` being a shared development tip and then having stabilization branches forked off of it to be more selective about what gets into a particular release. If I remember correctly, the "git flow" model follows this pretty well.

It doesn't seem like this project has enough code churn (yet?) to really need to be dogmatic about any particular workflow, but if you start noticing more PRs which seem reasonable but you're not comfortable merging into 1.5, having a think about setting up a release-branch pattern might be worth it.

Either way, I'm happy to merge this into a separate "future" branch too. I mostly just don't want my fork to diverge too much and end up in the same boat as Piccu, where it's so divergent that reintegration becomes prohibitively difficult.

Lgt2x:
> > How about merging these into another target branch rendering that we'll merge into main after our first "stable" release?
> 
> Seems reasonable. It's a bit out of scope for this PR, but in terms of branching and "keep stable stable but let devs dev" strategies, I'm more used to `main` being a shared development tip and then having stabilization branches forked off of it to be more selective about what gets into a particular release. If I remember correctly, the "git flow" model follows this pretty well.
> 
> It doesn't seem like this project has enough code churn (yet?) to really need to be dogmatic about any particular workflow, but if you start noticing more PRs which seem reasonable but you're not comfortable merging into 1.5, having a think about setting up a release-branch pattern might be worth it.
> 
> Either way, I'm happy to merge this into a separate "future" branch too. I mostly just don't want my fork to diverge too much and end up in the same boat as Piccu, where it's so divergent that reintegration becomes prohibitively difficult.

You're right. In this case most PRs right now should go into the next release, PRs like these ones (and probably #439  #485 ) are the exception. I created the `release` branch on this remote, I'll backport all necessary MRs from `main` regularly until the release. This way, people won't have to wonder which branch they should target, this is the maintainer's charge

renderer/HardwareOpenGL.cpp
```diff
@@ -1819,17 +1784,17 @@ void rend_FillRect(ddgr_color color, int x1, int y1, int x2, int y2) {
...
+  PosColorUVVertex vertex{
+      vector{static_cast<float>(x), static_cast<float>(y), 0},
+      color_array{GR_COLOR_RED(color) / 255.0f, GR_COLOR_GREEN(color) / 255.0f, GR_COLOR_BLUE(color) / 255.0f, 1},
+      tex_array{}
+  };
 
-  dglBegin(GL_POINTS);
-  dglVertex2i(x, y);
-  dglEnd();
+  dglVertexPointer(3, GL_FLOAT, sizeof(PosColorUVVertex), &vertex.pos);
```
Lgt2x, line 1795: to avoid costly conversions from int to float, could we use `GL_INT` as input?

renderer/HardwareBaseGPU.cpp
```diff
@@ -528,6 +476,26 @@ void rend_DrawPolygon2D(int handle, g3Point **p, int nv) {
   gpu_RenderPolygon(&vArray[0], nv);
 }
 
+color_array DeterminePointColor(g3Point const* pnt, bool disableGouraud, bool checkTextureQuality, bool flatColorForNoLight) {
```
Lgt2x, line 479: the function is used in critical paths, consider inlining it with `inline`

renderer/HardwareBaseGPU.cpp
```diff
@@ -528,6 +476,26 @@ void rend_DrawPolygon2D(int handle, g3Point **p, int nv) {
...
+  auto alpha = gpu_Alpha_multiplier * gpu_Alpha_factor;
+  if (gpu_state.cur_alpha_type & ATF_VERTEX) {
+    alpha *= pnt->p3_a;
+  }
+
+  // If we have a lighting model, apply the correct lighting!
+  if ((gpu_state.cur_light_state == LS_FLAT_GOURAUD && !disableGouraud) ||
```
Lgt2x, line 486: Maybe we should create templates and specializations of this function to replace branching by static dispatch to increase performance

renderer/HardwareBaseGPU.cpp
```diff
@@ -528,6 +476,26 @@ void rend_DrawPolygon2D(int handle, g3Point **p, int nv) {
...
+      (gpu_state.cur_light_state == LS_NONE && flatColorForNoLight)) {
+    return {GR_COLOR_RED(gpu_state.cur_color) / 255.0f, GR_COLOR_GREEN(gpu_state.cur_color) / 255.0f,
+            GR_COLOR_BLUE(gpu_state.cur_color) / 255.0f, alpha};
+  } else if (gpu_state.cur_light_state == LS_NONE) {
+    return {1, 1, 1, alpha};
+  } else if (gpu_state.cur_color_model == CM_MONO) {
+    return {pnt->p3_l, pnt->p3_l, pnt->p3_l, alpha};
+  } else {
+    return {pnt->p3_r, pnt->p3_g, pnt->p3_b, alpha};
+  }
```
Lgt2x, line 497: this would benefit being unit-tested

renderer/HardwareInternal.h
```diff
@@ -91,4 +91,6 @@ void gpu_RenderPolygonUV2(PosColorUV2Vertex *vData, uint32_t nv);
 void gpu_DrawFlatPolygon3D(g3Point **p, int nv);
 void rend_DrawMultitexturePolygon3D(int handle, g3Point **p, int nv, int map_type);
 
+color_array DeterminePointColor(g3Point const* pnt, bool disableGouraud = false, bool checkTextureQuality = false, bool flatColorForNoLight = false);
```
Lgt2x, line 94: add a documentation comment

Lgt2x (commented):
Really good work, I gave some feedback to possibly improve performance. I did not check correctness in depth compared to the previous version

tophyr:
Converting this to draft; I'll either be significantly reshaping this PR or just rolling its changes into a completely new one.

renderer/HardwareOpenGL.cpp
```diff
@@ -1860,13 +1820,27 @@ void rend_DrawLine(int x1, int y1, int x2, int y2) {
...
+  color_array color{
+      GR_COLOR_RED(gpu_state.cur_color) / 255.0f,
+      GR_COLOR_GREEN(gpu_state.cur_color) / 255.0f,
+      GR_COLOR_BLUE(gpu_state.cur_color) / 255.0f,
+  };
+  std::array<PosColorUVVertex, 2> vertices{
+      PosColorUVVertex{
+          vector{static_cast<float>(x1 + gpu_state.clip_x1), static_cast<float>(y1 + gpu_state.clip_y1), 0},
+          color,
+          tex_array{ /* unused */ }
```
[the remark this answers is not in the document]
tophyr: not a bad idea. they could even inherit from one another in order to be more flexible with typing

renderer/HardwareBaseGPU.cpp
tophyr: `inline` doesn't actually do what it's commonly thought to do: since c++11, it's a _linker_ flag indicating that multiple definitions are allowed, not a _compiler_ flag indicating that the function should be inlined into its callsites.

optimizers got smart enough (and LTO became enough of a thing) that the previous meaning of "plop this code directly in place instead of generating a callsite" became a bit useless

renderer/HardwareBaseGPU.cpp
tophyr: ooh, great idea. if i'm not able to pretty shortly turn this into shader logic, i'll convert it to template logic.

renderer/HardwareBaseGPU.cpp
tophyr: also great idea

renderer/HardwareBaseGPU.cpp
Lgt2x: I've been lied to! :smile: Thanks for the heads-up

PR #499, opened 2024-07-22
fix path creation in gamesave.cpp

vlohacks:
std::filesystem::create_directories returns false but no error code if the path already exists. Hence no longer saving was possible as soon as the path exists. Added check for the error code (ec) to be non-zero for the error message

### DIFF
--- a/Descent3/gamesave.cpp
+++ b/Descent3/gamesave.cpp
@@ -399,8 +399,10 @@ void SaveGameDialog() {
   // create savegame directory if it didn't exist before.
   std::error_code ec;
   if (!std::filesystem::create_directories(savegame_dir, ec)) {
-    DoMessageBox(TXT_ERROR, TXT_ERRCREATEDIR, MSGBOX_OK);
-    return;
+    if (ec != 0) {
+      DoMessageBox(TXT_ERROR, TXT_ERRCREATEDIR, MSGBOX_OK);
+      return;
+    }
   }
 
   // open window
--- a/Descent3/gamesave.cpp
+++ b/Descent3/gamesave.cpp
@@ -399,7 +399,7 @@ void SaveGameDialog() {
   // create savegame directory if it didn't exist before.
   std::error_code ec;
   if (!std::filesystem::create_directories(savegame_dir, ec)) {
-    if (!ec) {
+    if (ec) {
       DoMessageBox(TXT_ERROR, TXT_ERRCREATEDIR, MSGBOX_OK);
       return;
     }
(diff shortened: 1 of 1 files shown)


Lgt2x:
Good catch, probably an oversight of #475 . The MacOS build is failing though, you should probably use `if (!ec) {` in the condition

vlohacks:
> Good catch, probably an oversight of #475 . CI builds are failing though, you should probably use `if (!ec) {` in the condition

Most likely a comparison with some kind of std::error::no_error constant is the most correct way? Did not find anything like this however, so I changed it according to your suggestion.

Lgt2x:
> > Good catch, probably an oversight of #475 . CI builds are failing though, you should probably use `if (!ec) {` in the condition
> 
> Most likely a comparison with some kind of std::error::no_error constant is the most correct way? Did not find anything like this however, so I changed it according to your suggestion.

StackOverflow is fine with using `operator bool()` in this case https://stackoverflow.com/questions/41699343/how-do-i-test-that-an-stderror-code-is-not-an-error

vlohacks:
> > > Good catch, probably an oversight of #475 . CI builds are failing though, you should probably use `if (!ec) {` in the condition
> > 
> > 
> > Most likely a comparison with some kind of std::error::no_error constant is the most correct way? Did not find anything like this however, so I changed it according to your suggestion.
> 
> StackOverflow is fine with using `operator bool()` in this case https://stackoverflow.com/questions/41699343/how-do-i-test-that-an-stderror-code-is-not-an-error

Alright, thanks. And CI is also happy now.

Lgt2x (approved):

Lgt2x:
wait `!ec` is success, shouldn't the condition be reversed?

vlohacks:
> wait `!ec` is success, shouldn't the condition be reversed?

Doh.. absolutely... fixed.

Descent3/gamesave.cpp
```diff
@@ -399,8 +399,10 @@ void SaveGameDialog() {
   // create savegame directory if it didn't exist before.
   std::error_code ec;
   if (!std::filesystem::create_directories(savegame_dir, ec)) {
-    DoMessageBox(TXT_ERROR, TXT_ERRCREATEDIR, MSGBOX_OK);
-    return;
+    if (ec) {
+      DoMessageBox(TXT_ERROR, TXT_ERRCREATEDIR, MSGBOX_OK);
+      return;
+    }
   }
```
winterheart, line 406: We can check `ec` alone here, without checking return code of `create_directories`.

PR #500, opened 2024-07-25
Use standard temporary directory instead of custom/cache

Lgt2x:
## Pull Request Type

- [ ] GitHub Workflow changes
- [x] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

Temp dir will be `/tmp/Descent3/cache` on Unix. Leave `-tempdir` option anyway, in case we want 2 clients to be run on the same host, so they don't try to lock the same file

### DIFF
--- a/Descent3/init.cpp
+++ b/Descent3/init.cpp
@@ -1387,7 +1387,6 @@ void LoadGameSettings() {
   if (Katmai && !FindArg("-nosparkles")) {
     Render_powerup_sparkles = true;
   }
-
 }
 
 struct tTempFileInfo {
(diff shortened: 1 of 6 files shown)


Descent3/init.cpp
```diff
@@ -1988,8 +1988,14 @@ void SetupTempDirectory(void) {
   if (t_arg) {
     strcpy(Descent3_temp_directory, GameArgs[t_arg + 1]);
   } else {
-    // initialize it to custom/cache
-    ddio_MakePath(Descent3_temp_directory, Base_directory, "custom", "cache", NULL);
+    ddio_MakePath(Descent3_temp_directory, std::filesystem::temp_directory_path().u8string().c_str(), "Descent3",
```
winterheart, line 1991: Need to use non-throwable version of `temp_directory_path()` and check ec error on path construction.

Lgt2x: fixed

Descent3/init.cpp
```diff
@@ -1988,8 +1986,19 @@ void SetupTempDirectory(void) {
...
+      Error("Could not find temporary directory: \"%s\"", ec.message().c_str() );
+      exit(1);
+    }
+    ddio_MakePath(Descent3_temp_directory, tempPath.u8string().c_str(), "Descent3",
+                  "cache", NULL);
+    std::filesystem::create_directories(Descent3_temp_directory, ec);
+    if (ec) {
+      Error("Could not create temporary directory: \"%s\"", Descent3_temp_directory);
+      exit(1);
+    }
```
winterheart, line 2001: Sorry, another request. This should be outside of `else` block, so we can try to create non-existent directory defined via `t_arg`.

Lgt2x: done too

winterheart:
Tested on Windows, works fine.

PR #514, opened 2024-08-04
Windows crashdumps

pzychotic:
## Pull Request Type

- [ ] GitHub Workflow changes
- [x] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [x] Other changes

### Description

Since the move to x64 Windows builds the code to manually parse callstacks in the event of unhandled exceptions wasn't used anymore. Maybe that code could have been ported to x64, I dunno. But nowadays we can just easily create a crashdump file aka MiniDump and share that for analysis.

For this to be remotely useful I also enabled the creation of debug symbols (PDB files) for all build configurations and install them, so they are available in the GitHub workflow artifacts. These shouldn't end up in the official install for the game though. For official releases it would be good to have these attached to the release on GitHub.

So when a problem comes up we can look up the version number we now already have in the executable and is recorded in the crashdump, download the matching debug symbols and analys the crash.

### Checklist

### Additional Comments

For proper releases of the game it would be nice to have the debug symbols on a symbol server, so they can be fetched automatically when analyzing a crashdump. GitHub doesn't seem to have a solution for this.

### DIFF
--- a/Descent3/CMakeLists.txt
+++ b/Descent3/CMakeLists.txt
@@ -313,8 +313,12 @@ target_link_libraries(Descent3 PRIVATE
   music networking physics renderer rtperformance sndlib ui unzip vecmat md5
   ${PLATFORM_LIBS})
 target_include_directories(Descent3 PRIVATE ${PROJECT_BINARY_DIR}/lib)
+target_link_options(Descent3 PRIVATE $<$<PLATFORM_ID:Windows>:/DEBUG:FULL>)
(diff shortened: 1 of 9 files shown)


Lgt2x (approved):
That's really nice, I didn't know that was possible. It's kind of similar to the RelWithDebugInfo CMake build type. I'll try it out on Windows to see what it looks like and  include that in the upcoming release

Lgt2x:
Very useful tool, tested with success

PR #515, opened 2024-08-05
macOS: Make sure the netcon modules end with dylib.

MaddTheSane:
## Pull Request Type

- [ ] GitHub Workflow changes
- [X] Build and Dependency changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description
This changes the extension used by the NetCon modules ("Direct TCP~IP" and "Parallax Online") to be `.dylib` on macOS.
### Checklist

### Additional Comments
An alternative would be to delete the .dylib defines in the "d3c.txt" files and just use the .so extension.

### DIFF
--- a/netcon/lanclient/CMakeLists.txt
+++ b/netcon/lanclient/CMakeLists.txt
@@ -11,6 +11,9 @@ target_link_libraries(Direct_TCP_IP PRIVATE
   ui
   $<$<PLATFORM_ID:Windows>:ws2_32>
 )
+if(CMAKE_SYSTEM_NAME STREQUAL "Darwin")
+  set_target_properties(Direct_TCP_IP PROPERTIES SUFFIX ".dylib")
+endif()
 
 add_custom_target(Direct_TCP_IP_Hog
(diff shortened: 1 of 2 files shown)


Lgt2x (approved):
The change is fine, note that we can also set `MODULE` suffix globally using `CMAKE_SHARED_MODULE_SUFFIX` if needed

Lgt2x:
@MaddTheSane can you still use Multiplayer on Mac after this PR? From my testing, MP works on Mac using the release branch artifacts but not main branch artifacts, I suspect the problem comes from here. I don't have access to a mac often, so I cannot really bisect the issue right now

MaddTheSane:
Yeah, looks like I was a bit overzealous on changing the extension to _.dylib_. I was thinking of making a generated hog work on both Linux and Mac by having _.dylib_ and _.so_ files in the same hog. Unfortunately this will require a bit of work, as the extension that `LoadMultiDLL` looks for is _.dll_ on Windows but _.so_ for everything else.

Issue #506, opened 2024-07-30
Descent 3 1.5 Release process


Lgt2x:
We have now completed the minimum requirements initially set to make a first release of Descent 3 since its public code release on April 16th.

This issue tracks the release process for the first release of the game. Anyone is free to contribute to the following tasks, just say it here what you want to help with. Any order is fine, except for the last 2 items.

- [x] Changelog/Release notes: collect, group, sort and summarize changes from all merged PRs. Weekly summaries until 20/5 are available [here](https://github.com/DescentDevelopers/Descent3/wiki/Weekly-Updates) for rerefence
- [x] Test manually `release` branch on all supported platforms: Linux (at least Ubuntu, Fedora, Arch), Windows 10 and 11, Mac OS Intel and ARM chips, Raspberry Pi 64 bits. Testing include entering and completing levels from all campaigns, creating and joining a multiplayer game using TCP/IP or PXO, binding controls
- [x] Update project issues, test those that may be fixed alread.
- [x]  Disable shipping the internal level editor in the installed packaged. The editor is broken since the SDL2 transition, and would require additional work to get working again
- [x] Write a document attached with the game, listing the instructions to get it running : how to find game data on any given OS, how to run the game from there.
- [ ] clang-format the whole code base
- [x] Create the v1.5.0 tag from `release` branch
- [ ] Outreach: post about the release on both Discord servers, Reddit, forums, and other outlets we can access

JeodC:
You've got #130 and #279 left on the 1.5 milestone list. #130 I think may already be resolved pending a test, and #279 hasn't had any activity for some time.

Lgt2x:
> You've got #130 and #279 left on the 1.5 milestone list. #130 I think may already be resolved pending a test, and #279 hasn't had any activity for some time.

Oh I did not mention it but we decided to drop #279 for this release after some investigation, it looks very complicated to get structure alignment right. #130 should be fixed with the mve revamp. I've added an item for processing remaining open issues

PR #516, opened 2024-08-09
Create changelog for Descent Open Source 1.5

Lgt2x:
## Pull Request Type

- [ ] GitHub Workflow changes
- [x] Documentation or Wiki changes
- [ ] Build and Dependency changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

Full change log of the 1.5 open source version, with references to pull requests

### Related Issues

#506

PR #521, opened 2024-08-11
Create usage file for the packaged version

Lgt2x:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

Ship simplified usage instructions with the pre-compiled binaries.

### Related Issues

Part of #506


Lgt2x:
Before I forget, we need to mention the Wayland trick

CMakeLists.txt
```diff
@@ -123,7 +123,7 @@ add_custom_command(
 )
 
 install(
-  FILES LICENSE README.md THIRD_PARTY.md
```
winterheart, line 126: README.md should be included too into package.

Lgt2x: README contains mostly build instructions and detailed usage, do you think it makes sense to include it as-is in the pre-built package? I wouldnt expect to find build instructions in a package that does not contain the source

Lgt2x:
I'll need help refining the MacOS usage procedure. Per @tophyr  , `chmod` does not work every time, and you need to run `xattr -c /path/to/Descent3.app` before being able to run the `.app` bundle to remove the need for signing. Once this is done, are the assets correctly found in the game directory when you double-click the `.app`?

CMakeLists.txt
winterheart: Probably we can move building instructions to different file. Main idea is that README.md has all info that user should read first (like project description, license information, contact information), and then go to USAGE.md for actual settings.

Lgt2x: That makes sense, I'll work towards that. Our current full usage instructions make use of HTML formatting that is only displayed correctly on Github, so that'll have to change to be shipped

Lgt2x:
README has been split between USAGE and BUILD. I removed special build instructions formatted in HTML because it's an extremely unlikely scenario, and links to a project we do not have control over. We're ready for review! (I'd appreciate a native english speaker to take a look at this and #516, I don't trust my grammar that much)

winterheart:
LGTM.

PR #522, opened 2024-08-11
Revert "macOS: Make sure the netcon modules end with dylib."

MaddTheSane:
Reverts DescentDevelopers/Descent3#515
`LoadMultiDLL` doesn't currently look for _.dylib_.


winterheart:
Probably this can be fixed by adding additional `#ifdef` in [multi_dll_mgr.cpp](https://github.com/DescentDevelopers/Descent3/blob/main/Descent3/multi_dll_mgr.cpp#L606)

MaddTheSane:
That is another way to fix this, yes.

PR #526, opened 2024-08-13
OpenGL Cleanup

tophyr:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Documentation or Wiki changes
- [x] Build and Dependency changes
- [x] Runtime changes
  - [x] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description
Cleans up OpenGL usages. Removes dead code, removes redundant or ineffective code, and consolidates implementation details.

**Game outputs should remain exactly identical. Any graphics-behavior deviation is a bug.**

### Checklist

- [x] I have tested my changes locally and verified that they work as intended.
- [x] I have documented any new or modified functionality.
- [x] I have reviewed the changes to ensure they do not introduce any unnecessary complexity or duplicate code.
- [x] I understand that by submitting this pull request, I am agreeing to license my contributions under the project's license.

PR #527, opened 2024-08-13
Modernize Renderer

tophyr:
## Pull Request Type

- [ ] GitHub Workflow changes
- [x] Runtime changes
  - [x] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description
Modernize the Descent 3 renderer to use OpenGL 3 (GLES 3, eventually) and shaders. A major caveat here is that this is a minimal modernization of **just the OpenGL calls**. Performance should be _at least_ as good, but I am not by any means promising that the renderer is now "well-designed". It just now works using non-immediate-mode and non-fixed-function concepts.

**Rendering behavior should be identical!** Any difference should be considered a bug, unless discussed, determined to be acceptable, and specifically called out here.

#### Known Issues
None

### Related Issues
* Includes https://github.com/DescentDevelopers/Descent3/pull/526, examine that first.
* Tip of the hat to https://github.com/InsanityBringer/PiccuEngine


renderer/CMakeLists.txt
```diff
@@ -15,13 +30,12 @@ set(CPPS
   lnxscreenmode.cpp
 )
 
-# These are excluded.
-#opengl.cpp
-#renderer.cpp
-#Direct3D.cpp
+find_package(glm REQUIRED)
```
winterheart, line 33: Need to move this to top-level CMakeLists.txt to other external dependencies.

renderer/ShaderProgram.h
```diff
@@ -0,0 +1,217 @@
+/*
+* Descent 3
+* Copyright (C) 2024 Parallax Software
```
winterheart, line 3: Change it to "Copyright (C) 2024 Descent Developers" since this is new addition to code.

renderer/shaders.cmake
```diff
@@ -0,0 +1,31 @@
+file(WRITE ${shaders_h} "#pragma once
```
winterheart, line 1: Probably better move this file to top-level `cmake` directory where our cmake functions resides.

renderer/shaders/fragment.glsl
```diff
@@ -0,0 +1,40 @@
+#version 330 core
```
winterheart, line 1: I'd like to add our standard copyright header here if possible.

renderer/shaders/vertex.glsl
```diff
@@ -0,0 +1,23 @@
+#version 330 core
```
winterheart, line 1: I'd like to add our standard copyright header here if possible.

renderer/dyna_gl.h
```diff
@@ -36,213 +38,72 @@
...
-#endif
-#endif
-// end ryan's adds.
-
-#ifdef DECLARE_OPENGL
-#define DYNAEXTERN(_ty, _name) _ty _name = NULL
-#else
-#define DYNAEXTERN(_ty, _name) extern _ty _name;
+inline void CheckError() {
+#if defined(_DEBUG)
```
winterheart, line 42: Why error checking only on debug? If there runtime error on release build, what happens?

winterheart (changes requested):
Some quick non-related to functionality request changes.

winterheart:
On Linux I got runtime error:

```
terminate called after throwing an instance of 'std::runtime_error'
  what():  invalid enum
SIGNAL 6 caught, aborting
```

tophyr:
> On Linux I got runtime error:
> 
> ```
> terminate called after throwing an instance of 'std::runtime_error'
>   what():  invalid enum
> SIGNAL 6 caught, aborting
> ```

interesting. run it under a debugger and it should give you a stack trace of the failing call. some GL call is failing with `INVALID_ENUM`.

renderer/dyna_gl.h
tophyr: Speed, theoretically. I'm a big fan of "fail early, fail loud" so if there's consensus that we should error-check on Release builds as well, I'd be all for it.

To your second question of "what happens if we ignore the error" - that's `main`'s behavior today. Very few GL calls are error-checked, so errors typically just go ignored. OpenGL appears to ignore failed calls, for the most part.

winterheart:
```
Thread 1 (Thread 0x7ffff7082540 (LWP 11664) "Descent3"):
#0  CheckError () at /home/winterheart/playground/Descent3/renderer/dyna_gl.h:49
        _dglGetError = 0x7ffff19df5a0
        err = 1280
#1  0x000055555587c34c in FnPtr<unsigned char const* (unsigned int)>::operator()(unsigned int) const (this=0x5555565f7658 <dglGetString>, args#0=7939) at /home/winterheart/playground/Descent3/renderer/dyna_gl.h:75
        ret = 0x0
#2  0x0000555555870da2 in opengl_GetInformation () at /home/winterheart/playground/Descent3/renderer/HardwareOpenGL.cpp:242
No locals.
#3  0x0000555555872110 in opengl_Init (app=0x7fffffffd400, pref_state=0x555555a08160 <Render_preferred_state>) at /home/winterheart/playground/Descent3/renderer/HardwareOpenGL.cpp:588
        width = 640
        height = 480
        retval = 1
        i = 0
        windowX = 0
        windowY = 0
        __FUNCTION__ = {111 'o', 112 'p', 101 'e', 110 'n', 103 'g', 108 'l', 95 '_', 73 'I', 110 'n', 105 'i', 116 't', 0 '\000'}
#4  0x0000555555874676 in rend_Init (state=RENDERER_OPENGL, app=0x7fffffffd400, pref_state=0x555555a08160 <Render_preferred_state>) at /home/winterheart/playground/Descent3/renderer/HardwareOpenGL.cpp:1302
        retval = 0
        flags = 15360
#5  0x00005555555f3f16 in SetScreenMode (sm=2, force_res_change=false) at /home/winterheart/playground/Descent3/Descent3/game.cpp:970
        scr_width = 640
        scr_height = 480
        scr_bitdepth = 32
        old_sm = 0
        rend_width = 0
        rend_height = 0
        rs = {initted = 116 't', cur_bilinear_state = 10 '\n', cur_zbuffer_state = -103 '\231', cur_texture_type = 21845, cur_color_model = (unknown: 0xffffd1e0), cur_light_state = 32767, cur_alpha_type = -16 '\360', cur_wrap_type = (WT_CLAMP | WT_WRAP_V | unknown: 0x7ffc), cur_alpha = 1431942794, cur_color = 21845, cur_texture_quality = -16 '\360', clip_x1 = 32767, clip_x2 = 1435216648, clip_y1 = 21845, clip_y2 = -140061248, screen_width = 32767, screen_height = -311065088}
#6  0x00005555555e46e1 in Descent3 () at /home/winterheart/playground/Descent3/Descent3/descent.cpp:483
        proxyarg = 0
        type = 1
#7  0x00005555557842a7 in oeD3LnxApp::run (this=0x7fffffffd400) at /home/winterheart/playground/Descent3/Descent3/sdlmain.cpp:164
No locals.
#8  0x0000555555783720 in main (argc=5, argv=0x7fffffffd5d8) at /home/winterheart/playground/Descent3/Descent3/sdlmain.cpp:365
        d3 = {<oeLnxApplication> = {<oeApplication> = {_vptr.oeApplication = 0x555555958ea0 <vtable for oeD3LnxApp+16>, m_AppActive = true}, m_WasCreated = false, m_DeferFunc = 0x5555555e4e22 <D3DeferHandler(bool)>, static os_initialized = false, static first_time = true, m_Flags = 15360, m_X = 0, m_Y = 0, m_W = 640, m_H = 480}, shutdown = false, final_shutdown = false, old_screen_mode = 0}
        dbase = {<oeLnxAppDatabase> = {<oeAppDatabase> = {_vptr.oeAppDatabase = 0x555555958e40 <vtable for oeD3LnxDatabase+16>}, database = 0x5555567078a0}, <No data fields>}
        flags = 15360
        run_d3 = true
        __PRETTY_FUNCTION__ = {105 'i', 110 'n', 116 't', 32 ' ', 109 'm', 97 'a', 105 'i', 110 'n', 40 '(', 105 'i', 110 'n', 116 't', 44 ',', 32 ' ', 99 'c', 104 'h', 97 'a', 114 'r', 42 '*', 42 '*', 41 ')', 0 '\000'}
        defaultLogger = 0x555556698500
        rc = 0
        winArg = 2
        fsArg = 0
#9  0x00007ffff78ce330 in ?? () from /usr/lib64/libc.so.6
No symbol table info available.
#10 0x00007ffff78ce3e9 in __libc_start_main () from /usr/lib64/libc.so.6
No symbol table info available.
#11 0x0000555555579a65 in _start ()
No symbol table info available.
```

tophyr:
Force-push to address @winterheart feedback

tophyr:
Aha, interesting, the `mprintf()` call on that line is no-op'ed on my build for some reason. I'll force it to execute that and adjust

tophyr:
> On Linux I got runtime error:
> 
> ```
> terminate called after throwing an instance of 'std::runtime_error'
>   what():  invalid enum
> SIGNAL 6 caught, aborting
> ```

Fixed

renderer/dyna_gl.h
tophyr: Can do! Just to be clear, though: That will cause the Release build to be much crashier. Right now, GL errors are ignored unless they are explicitly checked-for (there are *some* places that check and then `Int3`). Any ignored GL error just ends up causing a graphics degradation, at most. If I turn this on for the Release builds, it will crash on *any* GL error.

This is something I'd realistically use telemetry for, if we were ok collecting it.

tophyr: @winterheart the more I think about it, the more I actually reverse my initial position: I think we should _not_, for now, enforce these error-checks on Release. We don't have (to my knowledge) a good library capable of generating a stacktrace to print for the user, and even if we did, on Mac and Windows (likely the two most common platforms) there isn't even a console to print the error to.

I have some ideas about how to build an error-reporting mechanism that will be palatable to even the most fanatically private users, and we could tie this in to that if/when we do it. But until then, I think turning OpenGL errors into fatal errors would be worse rather than better.

winterheart: OK, but we still need same behavior both for Debug and Release builds. Let's just replace throwing with verbose logging and unify this block for all environments.

tophyr: I'm curious why we need identical behavior between Debug and Release - that's an unusual requirement. In my experience it's very common for Debug builds to include extra assertions and error checking, with those same checks compiled out for the Release version.

CMake has a RelWithDebugInfo.. is that more what you're wanting?

tophyr: On further thought, this PR is not the forum to debate that product requirement either way. I'll gate this behind a compile-flag setting instead if a Debug check and then open an issue to discuss the higher-level issue.

tophyr: https://github.com/DescentDevelopers/Descent3/issues/528 opened, and https://github.com/DescentDevelopers/Descent3/pull/527/commits/e455ed5fa58615b4d39c9eefae74a85295b5a434 added to gate this for now.

winterheart (approved):
PR is ready to merge, let's merge it after 1.5 release.

winterheart:
Please resolve conflicts on files and we ready to set.

tophyr:
Rebased onto latest main

winterheart:
Retested on Linux, no visible issues found.

PR #525, opened 2024-08-12
Add release flow to GitHub Actions

winterheart:
Initial support to automate release flow. Now after git tag of specific commit GitHub Actions will create a draft entry of release (named by tag) with source code package on it.

## Pull Request Type

- [x] GitHub Workflow changes
- [ ] Documentation or Wiki changes
- [ ] Build and Dependency changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Related Issues

Related to #506.

### Checklist

- [x] I have tested my changes locally and verified that they work as intended.
- [x] I have documented any new or modified functionality.
- [x] I have reviewed the changes to ensure they do not introduce any unnecessary complexity or duplicate code.
- [x] I understand that by submitting this pull request, I am agreeing to license my contributions under the project's license.

PR #524, opened 2024-08-12
macOS: Fix netgames not looking for dylib.

MaddTheSane:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Documentation or Wiki changes
- [X] Build and Dependency changes
- [X] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [X] Network changes
  - [ ] Other changes

### Description
Have the net games and net modules on macOS look for the _.dylib_ extension.

### Related Issues

This should replace #522

### Checklist

- [X] I have tested my changes locally and verified that they work as intended.
- [X] I have documented any new or modified functionality.
- [X] I have reviewed the changes to ensure they do not introduce any unnecessary complexity or duplicate code.
- [X] I understand that by submitting this pull request, I am agreeing to license my contributions under the project's license.

PR #523, opened 2024-08-11
Fixes for Windows build

winterheart:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Documentation or Wiki changes
- [x] Build and Dependency changes
- [x] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [x] Other changes

### Description

This PR fixes two errors:

1. Fixes build compilation on Windows x86 (while we don't support it, but project can still buildable).
2. Fixes runtime error on starting new game with new pilot.

During start of new game with new pilot (never played) game tries to inject to play `level1.mve` cutscene before training mission right in the middle of update UI frame. At same time, if game previously found `mainmenu.mve`, it will be looped in main menu screen. Since buffers for MVE playback already allocated for `mainmenu.mve`, trying to start another movie will lead to crash.

Removed movie injection from playing, now new pilot will see this sequence:

New game -> Training mission -> level1.mve cutscene -> Level 1

Removed `RELEASE` `#ifdef`'s on this part of code as this definitely masked issue until we caught it in preparation of release build.

### Checklist

- [x] I have tested my changes locally and verified that they work as intended.
- [x] I have documented any new or modified functionality.
- [x] I have reviewed the changes to ensure they do not introduce any unnecessary complexity or duplicate code.
- [x] I understand that by submitting this pull request, I am agreeing to license my contributions under the project's license.

PR #517, opened 2024-08-09
CMake: add all headers to target sources

Lgt2x:
This can help with CMake IDE integration and file indexing.

## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Documentation or Wiki changes
- [x] Build and Dependency changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

Set every missing header as source of CMake targets. This can improve IDE integration, and is generally a good practice to follow.  

### Related Issues

Mentioned in #502 https://github.com/DescentDevelopers/Descent3/pull/502#discussion_r1697584916

This should probably cover what's left to do out of #411

PR #529, opened 2024-08-14
Update release branch with latest commits

Lgt2x:
This pushes PRs #525 #523 #517 #524 #518 #519 #512 #469 #507 #516 #521 into the release branch.

To be merged before tagging the release


tophyr:
lgtm, let's shipit

Lgt2x:
> lgtm, let's shipit

so everything runs fine (scripts, multiplayer...) from the release artifact `.app`? Usage procedure is correct?

tophyr:
Ok, I see the full range of Multiplayer options on Windows. Mac shows me Direct TCP/IP only. None of them appear to actually do anything, on either platform.

Lgt2x:
> > so everything runs fine (scripts, multiplayer...) from the release artifact `.app`? Usage procedure is correct?
> 
> hadn't tested multiplayer - it shows me "Direct TCP/IP" as the only option but then when i hit "OK" it just exits the menu and goes back to main screen. doesn't seem like that's intended behavior.
> 
> scripts work fine in Stone Cutter, Pilot Training and Retribution. didn't play much into any of each level.
> 
> `USAGE.md` reads well and following its instructions (and installing the `.hog` appropriately) fixes the problems i was seeing before.

Ok so multiplayer definitely does not work. #515 broke multiplayer, but #524 should have fixed it. Make sure you copy the `online` directory from the artifact to the game data location.

tophyr:
> Make sure you copy the `online` directory from the artifact to the game data location.

yep, **works fine**, i'm just an idiot again. once i copied the artifacts correctly i was able to create a TCP/IP game between my windows and mac machines and fight each other. appears to work great.

Parallax Online tries to connect, but appears to time out on windows - likely a firewall issue. Mac asks for login and password, which I don't have so am not gonna test right now.

Lgt2x:
Sounds good, thanks for testing!

Lgt2x:
I did a little more manual testing on Windows & Linux, looks ok

Issue #537, opened 2024-08-18
Logging framework


winterheart:
Currently we have a somewhat crippled logging facility based on an old ddebug/mono submodule. It is available only on Debug and does not provide any useful information on Release build. There was another attempt to use the spdlog library (#212), but we reverted it as it added unneeded external dependency.

For our needs logging framework should have following features:

* Header-only or extremely lightweight embedded library that allows us to use logging without dllexports in our external libraries.
* GPLv3-compatible license.
* Library should support console (stderr/stdout) and file output, with configurable format of message (like timestamp, severity, location and message).
* ostream (`<<`) and `printf`-like syntax for creating messages.
* Optionally support for C++17 and STL containers.

After some research I'd like to propose to switch the plog library (https://github.com/SergiusTheBest/plog) as the most suitable solution.

Lgt2x:
ok for the lib, also noting that it uses CMake (easy integration) and allows sharing instances across libs, which is going to be handy for logging in network code and level scripts. I would prefer actual functions to macros for easier debugging, but it's fine.

A few notes/questions:
 - Header-only means we can integrate it as an embedded 3rd party. No more use for `USE_LOGGER` which should be removed
 - I wish we can still keep our own logging module wrapping it, so we only expose our logging functions to the rest of the code. This way, plog is only linked to our `log` internal library. This allows us to change logging lib without changing thousands of lines if we need to, and to limit functions used (maybe don't use all logging levels?).
 - Logging should be enabled all the time for all builds, but default level changes between release and debug
 - Should we add CLI options to control log level, and output to stdout or file? `-logfile` is already a thing, but I'm wondering if the default should be output to file rather than stdout

Jayman2000:
Why is plog a better option than spdlog?

winterheart:
> Why is plog a better option than spdlog?

Main issue of spdlog that we forced to pull extra dependencies from outside into build process like external libfmt.

Jayman2000:
> > Why is plog a better option than spdlog?
>
> Main issue of spdlog that we forced to pull extra dependencies from outside into build process like external libfmt.

I agree that it’s a good thing that plog has less dependencies than spdlog (As far as I can tell, plog only depends on the standard library which is great!). That being said, I disagree that having extra dependencies is the main issue of spdlog.

[...]

PR #546, opened 2024-08-24
Enhance logging facility

winterheart:
## Pull Request Type

- [ ] GitHub Workflow changes
- [x] Runtime changes
  - [x] Render changes
  - [x] Audio changes
  - [x] Input changes
  - [x] Network changes
  - [x] Other changes

### Description

This is first part of integrating new logging library plog into project. Currently most parts of core game files now uses new logger, external libraries and tools will be in next PR.
Logger enabled both for Release (default level INFO) and Debug (default level DEBUG), and produces output to console and (optionally) into logfile (use `-logfile` option to enable it, file `Descent3.log` be placed next to executable). Windows also can use `-winconsole` option to enable separate console window.

Logger facility can be used anywhere in core game files by including `log.h` and using `LOG_*` macros.

Initial documenting all cmd-line options in `USAGE.md` file.

Test on Linux (Release / Debug) and Windows (Release / Debug + Editor).

### Related Issues

Related to #537.

### Checklist


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "third_party/plog"]
+	path = third_party/plog
+	url = https://github.com/SergiusTheBest/plog.git
```
Lgt2x, line 3: I cant comment on the submodule itself, but why master branch and not latest release?

manage/CMakeLists.txt
```diff
@@ -30,4 +30,5 @@ target_link_libraries(manage PRIVATE
   model
   sndlib
   stream_audio
+  plog::plog
```
Lgt2x, line 33: Do you plan to create a real `log` internal library we link against instead of plog directly? It was part of my wishlist in #537 :) This lib could only be including `log.h`

CMakeLists.txt
```diff
@@ -11,6 +11,7 @@ option(ENABLE_MEM_RTL "Enable Real-time library memory management functions (dis
 option(FATAL_GL_ERRORS "Check OpenGL calls and raise exceptions on errors." OFF)
 option(FORCE_COLORED_OUTPUT "Always produce ANSI-colored compiler warnings/errors (GCC/Clang only; esp. useful with Ninja)." OFF)
 option(FORCE_PORTABLE_INSTALL "Install all files into local directory defined by CMAKE_INSTALL_PREFIX" ON)
+option(USE_SYSTEM_PLOG "Use system plog library instead bundled" OFF)
```
Lgt2x, line 14: Do any distribution actually package it? It's header-only, so the package would not contain any shared library or anything.

Descent3/Controls.cpp
```diff
@@ -805,11 +799,6 @@ void DoMovement(game_controls *controls) {
     controls->afterburn_thrust = LIMIT_FORWARD;
   if (controls->afterburn_thrust < -LIMIT_FORWARD)
     controls->afterburn_thrust = -LIMIT_FORWARD;
-
-  mprintf_at(1, 5, 30, "ch:%.2f ", controls->heading_thrust);
```
Lgt2x, line 809: no replacement?

.gitmodules
winterheart: Good point, will switch to latest available stable tag.

manage/CMakeLists.txt
winterheart: Yes, I planning to do it. In following PR I will create separate `log` submodule with some functionality from ddebug and misc.

CMakeLists.txt
winterheart: At least I saw plog library in vcpkg, and there other distributions can package it: https://repology.org/project/plog/versions

Descent3/Controls.cpp
winterheart: `mprintf_at()` is no-op, it can be safely removed. Will remove after ddebug cleanup eventually.

CMakeLists.txt
Lgt2x: it could be a good idea to use VCPKG then! So packagers only have to turn off `USE_VCPKG`, and we don't have to remind people to init submodules

winterheart: Enabled plog on vcpkg.

Lgt2x: So now there is no point keeping the `USE_SYSTEM_PLOG` option, people will set `USE_VCPKG` or not instead

Lgt2x: I think I get what you want to do:
1. If user has plog packaged in the distribution (only major distros with a plog package being Fedora/Gentoo/AUR), set `USE_SYSTEM_PLOG` and don't use the submodule
2. If user has `USE_VCPKG`, install it through VCPKG and ignore submodule
3. Otherwise, use submodule, ie for distributions that don't package it

Now that I think about it the logic is a little complicated. We have to cover 3. for distros such as Debian that cant use VCPKG and don't have a native plog package. Maybe VCPKG is not necessary after all...

winterheart: Actually there only 2 options - use system plog or use it from submodule. `USE_SYSTEM_PLOG` + vcpkg (if it enabled) makes plog system dependency, it's not important where it came from - we threats it as external library.

Lgt2x (approved):
Ok for v1! See option name change suggestion

winterheart:
Changed USE_SYSTEM_PLOG to USE_EXTERNAL_PLOG.

PR #547, opened 2024-08-25
Replace busy wait loop with a sleep

pzychotic:
## Pull Request Type

- [ ] GitHub Workflow changes
- [x] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [x] Other changes

### Description

I noticed one of my CPU cores running at 100% when in the main menu.
A quick profiling found a loop just busy-waiting for some time to pass that I replaced by a sleep for that amount of time (50ms).

### Checklist

### DIFF
--- a/ui/UISystem.cpp
+++ b/ui/UISystem.cpp
@@ -175,7 +175,7 @@
 
 #define UI_MOUSE_HOTX 2
 #define UI_MOUSE_HOTY 2
-#define UI_FRAMETIME 0.05
+constexpr int ui_FrameTimeMS = 50;
 //////////////////////////////////////////////////////////////////////////////
 //	VARIABLES
 struct tUIWindowNode {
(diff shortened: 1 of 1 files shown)


ui/UISystem.cpp
```diff
@@ -502,9 +502,7 @@ int ui_DoFrame(bool input) {
     ui_DoWindowFocus();              //	determine window with current input focus.
     res = ui_ProcessFocusedWindow(); //	process focused window
 
-    //		int64_t cur_time = timer_GetMSTime();
-    while ((timer_GetTime() - UI_input.cur_time) < UI_FRAMETIME) {
-    };
```
winterheart, line 507: I like the general idea, but old logic is sleep rest of UI_FRAMETIME with delta of current time and last timestamp of UI_input struct. I measured some timings, and `ui_DoFrame()` takes about 10 ms (on Debug), and with this we should sleep (50 - 10) ms. So we should measure, how many of 50 ms we should sleep each frame.

winterheart:
Tested on Linux, works as intended.

PR #548, opened 2024-08-25
Fix small collection of errors

pzychotic:
## Pull Request Type

- [ ] GitHub Workflow changes
- [x] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [x] Other changes

### Description

Fixed a bunch of small errors, like inverted error checks that do nothing, assignments in 'if' statements and missing va_end calls.

### Checklist

### DIFF
--- a/Descent3/room.cpp
+++ b/Descent3/room.cpp
@@ -1096,7 +1096,7 @@ int CheckTransparentPoint(const vector *pnt, const room *rp, const int facenum) 
 
 // Computes a bounding sphere for the current room
 // Parameters: center - filled in with the center point of the sphere
-//		rp - the room we’re bounding
+//		rp - the room we're bounding
 // Returns: the radius of the bounding sphere
 float ComputeRoomBoundingSphere(vector *center, room *rp) {
   // This algorithm is from Graphics Gems I.  There's a better algorithm in Graphics Gems III that
--- a/dd_grwin32/ddgrWin32Init.cpp
+++ b/dd_grwin32/ddgrWin32Init.cpp
@@ -191,7 +191,7 @@ void ddgr_FatalError(const char *fmt, ...) {
   std::vsnprintf(buf, 768, fmt, arglist);
   va_end(arglist);
 
-  if (DDGR_subsystems[i] = !-1) {
+  if (DDGR_subsystems[i] != -1) {
     strcat(buf, "\n\nSubsystem:  ");
     strcat(buf, DDGR_subsystem_names[i]);
   }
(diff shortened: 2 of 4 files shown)


winterheart:
That interesting things... I wonders that some of these bugs actually now "features" that we should reproduce as broken logic for playing levels...

jengelh (commented):
The source code is decidedly UTF-8 (as indicated by CMakeLists.txt), so the use of U+2019 does not constitute a bug per se.

pzychotic:
> The source code is decidedly UTF-8 (as indicated by CMakeLists.txt), so the use of U+2019 does not constitute a bug per se.

I actually get a MSVC compiler warning without that change:
Warning C4828 The file contains a character starting at offset 0x7241 that is illegal in the current source character set (codepage 65001).
Notepad++ shows this file to be ANSI, when saving as UTF-8 I get a change to that character and the warning is fixed.

Lgt2x:
even though it's technically changing logic, most changes seem rather innocuous. The only one I'm not sure about is the one in `physics.cpp` : some mechanics could have been applied to other types of objects not intended because the `OBJ_CLUTTER` test succeeded all the time. Do you think that it's reasonable to merge as it is @winterheart ? I don't think it will cause big problems anyway

winterheart:
I'm concerning that would change logic in level. Need to test these changes.

jengelh:
I have given this a look, by replacing the lone `OBJ_CLUTTER` by a function `XOBJ_CLUTTER(obj)`:

```c++
bool XOBJ_CLUTTER(const object *obj) {
    if (obj->type == OBJ_CLUTTER)
        return true;
    printf("otype %u\n", obj->type);
    return false;
}
```

The path to these places suggested to me that a clutter objects needs to collide with terrain, but even on level 3 (lightning tower balls) that did not seem to readily trigger.

pzychotic:
> I have given this a look, by replacing the lone `OBJ_CLUTTER` by a function `XOBJ_CLUTTER(obj)`:
> ...
> The path to these places suggested to me that a clutter objects needs to collide with terrain, but even on level 3 (lightning tower balls) that did not seem to readily trigger.

Nice approach!

If that change is too worrisome, I can split it out into a separate PR. Just didn't want to create a bunch of PRs for these little changes.

jengelh:
Yeah the last two commits seem to warrant more investigation in their own right.

pzychotic:
Updated without the possibly game play altering fixes.

Lgt2x (approved):
LGTM

Issue #528, opened 2024-08-14
Behavior differences between Debug/Release


tophyr:
OK, but we still need same behavior both for Debug and Release builds. Let's just replace throwing with verbose logging and unify this block for all environments.

_Originally posted by @winterheart in https://github.com/DescentDevelopers/Descent3/pull/527#discussion_r1716846993_
           
**I think the product requirement that Debug and Release must behave identically is detrimental.**

It's a fairly common practice to sprinkle assertions and sanity-checks throughout code that are active only in Debug builds and get compiled out in Release builds, in order to let developers catch problems and bad assumptions but still keep non-developer builds stable and performant. In many cases (including in https://github.com/DescentDevelopers/Descent3/pull/527 stemming this discussion) the error information being reported would be useless at best - and more often actively confusing - to a non-developer, but could offer a critically-important clue to someone with the tools to properly analyze it.

To wit, there are already numerous existing cases in the codebase where additional developer-oriented behavior is enabled when building in Debug mode:
* The commonly-used `DEBUG_BREAK()`, `ASSERT()`, and `Int3()` macros are only operational in Debug builds: https://github.com/DescentDevelopers/Descent3/blob/a78eb53f692520379bde498a2dd96a5781a1d04c/misc/pserror.h#L236-L261
* Chording the "delete" key with various other keys enables extra onscreen information: https://github.com/DescentDevelopers/Descent3/blob/main/ddio/key.cpp#L288-L291
* Level loading includes additional sanity checks: https://github.com/DescentDevelopers/Descent3/blob/a78eb53f692520379bde498a2dd96a5781a1d04c/Descent3/OsirisLoadandBind.cpp#L1050-L1053
* Many more examples are discoverable simply by searching for `#ifdef _DEBUG`

[...]

Issue #509, opened 2024-08-02
[Runtime Issue]: Debug windows build cannot start - missing ucrtbased.dll


JeodC:
### Build Version

current

### Operating System Environment

- [ ] Microsoft Windows (32-bit)
- [X] Microsoft Windows (64-bit)
- [ ] Mac OS X
- [ ] Linux (specify distribution and version below)

### CPU Environment

- [ ] x86 (32-bit Intel/AMD)
- [X] x86\_64 (64-bit Intel/AMD)
- [ ] ARM (32-bit)
- [ ] ARM64 (64-bit; sometimes called AArch64)
- [ ] Other (RISC V, PPC...)

### Game Modes Affected

- [ ] Single player
- [ ] Anarchy
- [ ] Hyper-Anarchy
- [ ] Robo-Anarchy
- [ ] Team Anarchy
- [ ] Capture the Flag
- [ ] Bounty
- [ ] Entropy
- [ ] Hoard
- [ ] Monsterball
- [ ] Cooperative

### Description

On attempting to run the Windows x64 debug build, a message appears about a missing `ucrtbased.dll` file. This file is not packaged with the debug artifact.

pzychotic:
Working as intended I suppose, if you don't have VS or at least the Windows SDK installed.

The debug versions of the VC runtime libraries are not supposed to be redistributed.

See https://learn.microsoft.com/en-us/cpp/windows/preparing-a-test-machine-to-run-a-debug-executable?view=msvc-170:
_Debug versions of an application aren't redistributable, and debug versions of the Visual C++ library DLLs are also not redistributable. You may deploy debug versions of applications and Visual C++ DLLs only to your other computers, for the sole purpose of debugging and testing the applications on a computer that doesn't have Visual Studio installed._

Or https://learn.microsoft.com/en-us/cpp/windows/determining-which-dlls-to-redistribute?view=msvc-170:
_You also can't redistribute debug versions of your applications, or the debug versions of DLLs or libraries supplied by Microsoft. You may only copy debug executables and libraries to machines you use for testing._

JeodC:
This hasn't been a thing for long. I've got no problem with it, but others might, especially for testing. The change that led to the requirement needs to be documented probably.

winterheart:
Probably we may to switch from full `Debug` build to `RelWithDebInfo` on artifact building (at least for Windows). That way we retain ability to get debug info from users and don't force them to install full MSVC environment for that.

[...]

PR #550, opened 2024-08-28
Improve render perf for MSVC Debug configuration

tophyr:
## Pull Request Type

- [ ] GitHub Workflow changes
- [x] Runtime changes
  - [x] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description
MSVC's STL performs extensive bounds and validity checking on iterators in Debug builds. This is desirable for correctness, but causes significant slowdowns in the current renderer's implementation due to the amount of vertex copies and transformations it performs. Until we can fully convert the renderer into more of a 'static data' model where the game does not need to ship thousands of vertices to the GPU each frame, the MSVC Debug build realistically needs to skip the per-iteration checks for its vertex iterators. This is, I think, a happy compromise between full checks and just using bare data pointers: The iterators are still used and checked once each at the start of any operation, but the memory accesses are done via pointers which should be very fast.

### Related Issues
#543, #545 

### Checklist

### DIFF
--- a/renderer/HardwareOpenGL.cpp
+++ b/renderer/HardwareOpenGL.cpp
@@ -119,7 +119,9 @@ struct Renderer {
             typename = std::enable_if_t<std::is_same_v<typename std::iterator_traits<VertexIter>::value_type, PosColorUVVertex>>>
   size_t addVertexData(VertexIter begin, VertexIter end, PosColorUVVertex_tag = {}) {
     std::array<PosColorUV2Vertex, MAX_POINTS_IN_POLY> converted;
-    std::transform(begin, end, converted.begin(), [](auto const& vtx) {
+    ASSERT(std::distance(begin, end) <= converted.size());
+    // note: funny &* here is to convert the iterators into plain pointers, so MSVC's Debug STL doesn't slow this down
+    std::transform(&*begin, &*end, &*converted.begin(), [](auto const& vtx) {
       return PosColorUV2Vertex{vtx.pos, vtx.color, vtx.uv, {}};
     });
     return shader_.addVertexData(converted.cbegin(), converted.cend());
--- a/renderer/ShaderProgram.h
+++ b/renderer/ShaderProgram.h
@@ -129,7 +129,7 @@ struct OrphaningVertexBuffer : VertexBuffer<V> {
 
     auto start = nextVertex_;
     V* mapped = reinterpret_cast<V*>(dglMapBufferRange(GL_ARRAY_BUFFER, start * sizeof(V), dist * sizeof(V), GL_MAP_WRITE_BIT | GL_MAP_UNSYNCHRONIZED_BIT));
-    std::copy(begin, end, mapped);
+    std::copy(&*begin, &*end, mapped); // note: funny &* here is to convert the iterator into a plain pointer, so MSVC's Debug STL doesn't slow this down
     dglUnmapBuffer(GL_ARRAY_BUFFER);
 
     nextVertex_ += dist;


winterheart:
I'd like just not count Debug build on MSVC as main target of our development. Users does not supposed to use debug builds, and slow performance on Debug is compromise over clean code, which runs fine on Release.

tophyr:
Agree ideologically, but this is very nearly identically-safe and should eliminate a very big bottleneck in that particular build, making debugging on Windows easier.

winterheart:
We have 10 builds for various platforms, and there only one has performance regression - Windows Debug, so I suggest either fix it alone without affecting other builds or just threat it as "Debug is not about performance".

After all, we can just disable iterator debugs on whole project for Windows Debug, I don't see any usefulness of this feature at all for us.

pzychotic:
What about CMake's RelWithDebInfo build configuration when you need that speed while still being able to debug?

winterheart:
> What about CMake's RelWithDebInfo build configuration when you need that speed while still being able to debug?

That would be good solution too.

tophyr:
`RelWithDebInfo` turns on neither the `RELEASE` nor the `_DEBUG` defines, which unfortunately means it doesn't get many of the other asserts or checks. Some are keyed on `!RELEASE` while many others are keyed on actually `_DEBUG` being true.

Normally I'm pretty strongly in agreement with the "don't alter code for special cases" line of thought, but I do at the same time think it's pretty important to keep the Debug build reasonably usable - otherwise people will never use it.

The alternatives to this change that I see are:
1. Let Windows Debug build be near-unusably slow
2. `#ifdef` this change so the conversion from `iterator` -> `T*` is only performed on Windows
3. Turn off iterator validity checks project-wide
4. Alter the method hierarchies to take `T*` instead of iterators

Option 1 is, to me, not really an option - we might as well remove the preset for it. Option 2 could be done in a semi-cleanly way using some intermediate local variables, so that significant logic lines remain the same - but this would be entirely a semantic change, because `iterator` already decays to `T*` on non-Windows platforms. Additionally, Option 2 would ignore the possibility that other platforms might implement similar functionality, making us hunt down a config-specific perf regression. Options 3 and 4 are undesirable to me because I think the iterator checks _are_ actually extremely valuable.

What's the resistance to this one - is it just that the `&*` syntax is confusing without the accompanying comment?

winterheart:
On RelWithDebInfo we can enable `_DEBUG` defines in CMakeLists.txt, it's our defined macro. That why I'm trying to not overuse build-defined logics besides ASSERTS and fatal exceptions.

Prebuilt Debug artifacts on Windows already no has any use for users - they cannot run it without MSVC itself (as pointed in #509), so Debug environment is usable only for developer. Slow performance is observed only in one of our supported environments - Windows Debug, so we should either fix our code in way that not affects other environments or fix environment itself. Using `&*` syntax is hackish looking attempt to fix code that aren't even broken. Other platforms should not suffer with such non-intuitive syntax.

I still suggest either to disable iterator checking "feature" for Windows Debug (we don't has any use of this with well written code and boundary checks where it's really needed) or left Windows Debug slow as is.

tophyr:
> On RelWithDebInfo we can enable _DEBUG defines

This would eliminate the point of the RelWithDebugInfo build - it is meant to be a Release build, just with debug info.

The iterator checks are, imo, incredibly important. They've already caught one bug in the recent past for me and will continue to catch others.

I will try to rewrite this so that it uses clearer syntax, but I don't want to target the change specifically to Windows. (In #528 I talk about how I actually want to turn _on_ similar checks for Mac and Linux.)

tophyr:
Very interesting about #509. Hadn't seen that one yet.

pzychotic:
`_DEBUG` is at least a MSVC/Windows specific define, we can't enable that when linking to the release VC Runtime libraries.
On the other hand `RELEASE` is our project specific define, `NDEBUG` would be the opposite for MSVC/Windows in Release build.

There is some inconsistency in the current code, regarding these defines.
`#ifdef _DEBUG` is used alot and most cases of `RELEASE` are used in #if**n**def.
Which basically resolves to the same when we currently only use Release builds with `RELEASE` and Debug builds with `_DEBUG` enabled.
We could break that inconsistency with converting all the `#ifdef _DEBUG` to `#ifndef RELEASE`. Or since I'm not a big fan of negated #ifdef's, change everything to our own` D3_DEBUG` or something.

Btw. I can't say that I find the current Windows Debug build to be near-unusably slow, since that is the build I'm usually working with.

tophyr:
> There is some inconsistency in the current code, regarding these defines. `#ifdef _DEBUG` is used alot and most cases of `RELEASE` are used in #if**n**def. Which basically resolves to the same when we currently only use Release builds with `RELEASE` and Debug builds with `_DEBUG` enabled. We could break that inconsistency with converting all the `#ifdef _DEBUG` to `#ifndef RELEASE`. Or since I'm not a big fan of negated #ifdef's, change everything to our own` D3_DEBUG` or something.

Agree - I talk about this a bit in https://github.com/DescentDevelopers/Descent3/issues/528#issuecomment-2324029006

> Btw. I can't say that I find the current Windows Debug build to be near-unusably slow, since that is the build I'm usually working with.

Maybe we just accept the CPU-perf cost of this stuff then? I believe enough in the importance of the iterator checks that I'm unwilling to remove them. @Lgt2x did mention in the [original report](https://github.com/DescentDevelopers/Descent3/pull/545#issuecomment-2311084526) that his test machine was pretty low-end - as long as the game runs performantly-enough on "most" systems that are going to be running Debug builds, I'm ok with leaving this as-is.

PR #553, opened 2024-08-30
Fix 13 different crashes [ASAN]

jengelh:
## Pull Request Type

- [x] Runtime changes
  - [x] Other changes

### Description

A regular build (-O2 -g) of the game crashed for me for the first time in level 4 picking up SeekerMine. At that point, I exchanged -O2 for -O0 -g -fsanitize=address,undefined and only got more program terminations 🥲

### Screenshots (if applicable)

smine

### Checklist

- [x] I have tested my changes locally and verified that they work as intended.
- [x] I have reviewed the changes to ensure they do not introduce any unnecessary complexity or duplicate code.
- [x] I understand that by submitting this pull request, I am agreeing to license my contributions under the project's license.

PR #564, opened 2024-09-02
Resolve some 6 ASAN reports

jengelh:
## Pull Request Type

- [x] Runtime changes
  - [x] Other changes

### Descriptions / 

Continue fixing ASAN reports as they come in.

### Related Issues

https://github.com/DescentDevelopers/Descent3/pull/553

### Checklist

### DIFF
--- a/Descent3/aipath.cpp
+++ b/Descent3/aipath.cpp
@@ -651,6 +651,8 @@ static inline bool AIPathAddDPathNode(ai_path_info *aip, int *slot, int *cur_nod
     status = AIPathAddDPath(aip, handle);
     if (!status)
       Int3(); // chrishack -- out of dynamic paths
+    if (aip->num_paths < 1)
+      return false;
 
     AIDynamicPath[*slot].num_nodes = MAX_NODES;
     *cur_node = 0;
@@ -660,7 +662,8 @@ static inline bool AIPathAddDPathNode(ai_path_info *aip, int *slot, int *cur_nod
   AIDynamicPath[*slot].pos[*cur_node] = *pos;
   AIDynamicPath[*slot].roomnum[(*cur_node)++] = room;
 
-  aip->path_end_node[aip->num_paths - 1] = *cur_node - 1;
+  if (aip->num_paths >= 0)
+    aip->path_end_node[aip->num_paths - 1] = *cur_node - 1;
 
   return true;
 }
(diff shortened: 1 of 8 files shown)


Descent3/terrain.cpp
```diff
@@ -293,13 +294,22 @@ void GenerateLODDeltas() {
...
+// The checksum (or rather, an affine modification of it known as
+// occlusion_checksum) is stored in the level's map files, so any modification
+// here ought to not alter the output value, for all imaginable inputs.
 int GetTerrainGeometryChecksum() {
-  int total = 0;
+  unsigned int total = 0;
 
   for (int i = 0; i < TERRAIN_WIDTH * TERRAIN_DEPTH; i++) {
-    total += (Terrain_seg[i].ypos + i);
+    float yi = Terrain_seg[i].ypos + i;
```
Lgt2x, line 305: I don't get that. Why is `int` + `uint8_t` promoted to float? Is that an obscure C++ gotcha?

Descent3/terrain.cpp
```diff
@@ -293,13 +294,22 @@ void GenerateLODDeltas() {
...
+    if (yi < 2147483648.0f || yi >= 2147483520.0f)
```
Lgt2x, line 306: Where do these value come from? Is that some IEEE754 magic I'm missing?

Descent3/terrain.cpp
jengelh: I must have mixed it up with terrain_segment::y (which is the one that is float). addressed now.

Descent3/terrain.cpp
jengelh: The next float number after 2147483520 is 2147483648, but 2147483648 is already too large to be represented in an int32. Does not matter now, code is replaced.

jengelh:
* Rebased (I hope that's ok),
* removed two leftover printfs,
* purged the excessive floating point logic from GetTerrainGeometryChecksum

Descent3/aipath.cpp
```diff
@@ -651,6 +651,8 @@ static inline bool AIPathAddDPathNode(ai_path_info *aip, int *slot, int *cur_nod
     status = AIPathAddDPath(aip, handle);
     if (!status)
       Int3(); // chrishack -- out of dynamic paths
+    if (aip->num_paths < 0)
```
Lgt2x, line 654: below we have `aip->path_id[aip->num_paths - 1]`, so we should return if `aip->num_paths < 1` instead. Same logic below

Lgt2x (commented):
no problem with rebasing branches

Lgt2x (approved):

PR #566, opened 2024-09-02
Repair non-sensical scorch iterations

jengelh:
## Pull Request Type

- [x] Runtime changes
  - [x] Render changes

### Description

Gazing at my own code/modification, I find it is embarrasingly wrong. Since ``sp`` is re-initialized at every loop iteration, setting it to (the locally-constant) ``Scorch_start`` is incorrect. Make ``sp`` really mirror ``i`` at all times; this way, it also will not be necessary anymore to update ``sp`` within the loop.

Fixes: 20ed30eef804a5a251920cd8484043ef7ea48e78

### Checklist


Lgt2x:
oh right, that seems so obviously wrong now

Lgt2x (approved):

PR #573, opened 2024-09-08
Resolve UBSAN strict aliasing reports

jengelh:
## Pull Request Type

- [x] Runtime changes
  - [x] Other changes

### Description

float z = 1.0f;
*(int *)&z = 1;

This is undefined behavior, and gccs in the past have deleted the assignment.

### Checklist

- [x] I have tested my changes locally and verified that they work as intended. [Level6]
- [x] I have tested my changes locally and verified that they work as intended. [Level15]


Lgt2x (commented):
I could not find any usage of either of these functions elsewhere in the code. It may be easier to delete them altogether.

jengelh:
So deleted.

Lgt2x (approved):
I'm fine with the changes. Tested on Linux. We're breaking the public scripting API with the import change, but that's not a concern for now.

jengelh:
I would expect mods be able to workaround build failures just by adding ``#include "osiris_vector.cpp"`` + ``#include "DallasFuncs.cpp"`` (or otherwise including them in their build). Preexisting binary builds from back then should be unaffected since User_vars is always local to the game/level dll.

scripts/DallasFuncs.h
```diff
@@ -0,0 +1,360 @@
...
+	// retains the active type, requiring instead set_type<> to change it.
+	template<typename T> void operator=(T &&v) noexcept {
+		VarT &self = *this;
+		if (std::get_if<float>(this))
+			self = static_cast<float>(v);
+		else
+			self = static_cast<int32_t>(v);
+	}
+	template<typename T> void set_type() noexcept { static_cast<VarT &>(*this) = T{}; }
+	void operator++(int) noexcept { std::visit([](auto &&uv) { ++uv; }, *this); }
```
Lgt2x, line 21: This is causing compilation problems on GCC 9.4 (Ubuntu 20.04), because [p2162r2](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p2162r2.html) is not implemented. It has only been backported [to GCC 11](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=90943). We'll need to implement our own `visit` function to circumvent that

Lgt2x: @jengelh could you please take a look at https://github.com/DescentDevelopers/Descent3/pull/592 ?

PR #568, opened 2024-09-04
Fix discovered memory leaks

winterheart:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Documentation or Wiki changes
- [ ] Build and Dependency changes
- [x] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [x] Other changes

### Description

Fixed some memory leaks discovered with Valgrind.

### Checklist

- [x] I have tested my changes locally and verified that they work as intended.
- [x] I have documented any new or modified functionality.
- [x] I have reviewed the changes to ensure they do not introduce any unnecessary complexity or duplicate code.
- [x] I understand that by submitting this pull request, I am agreeing to license my contributions under the project's license.

PR #574, opened 2024-09-08
Resolve 2 UBSAN warnings about uninitialized vars; establish allocation/type check along with it

jengelh:
## Pull Request Type

- [x] Runtime changes
  - [x] Other changes

### Description

UBSAN evaluates use of undefined variable contents at runtime; it's more effective against unintialized heap allocations, whereas compile-time is more effective to mark code paths that are perhaps not exercised during gameplay at all.
mem_malloc is rewritten to catch potentially allocating nontrivial class types, the latter of which is a result of adding initializers. This supersedes https://github.com/DescentDevelopers/Descent3/pull/568/commits/7777720695dc9b294ba01e22fbcd53032dfe5b38 which only fixed a few struct members.

### Related Issues

#568

### Checklist

### DIFF
--- a/mem/mem.h
+++ b/mem/mem.h
@@ -73,6 +73,19 @@
 
 #ifndef MEM_H
 #define MEM_H
+#include <cstdlib>
+#include <type_traits>
+
+template<typename T> static inline T *mem_rmalloc()
+{
+	static_assert(std::is_trivially_constructible_v<T> && std::is_trivially_destructible_v<T>);
+	return static_cast<T *>(std::malloc(sizeof(T)));
+}
+template<typename T> static inline T *mem_rmalloc(std::size_t nelem)
+{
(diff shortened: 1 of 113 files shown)


Lgt2x (approved):
Very clean!

mem/mem.h
```diff
@@ -73,6 +73,19 @@
 
 #ifndef MEM_H
 #define MEM_H
+#include <cstdlib>
+#include <type_traits>
+
+template<typename T> static inline T *mem_rmalloc()
+{
+	static_assert(std::is_trivially_constructible_v<T> && std::is_trivially_destructible_v<T>);
+	return static_cast<T *>(::malloc(sizeof(T)));
```
Lgt2x, line 82: I'm curious about this practice, why using `::malloc` rather than `std::malloc`?

jengelh: old bad habits 🥲

Lgt2x:
Ok to merge when trivial conflict with logging module is solved

PR #582, opened 2024-09-09
ASAN/UBSAN fixes

jengelh:
## Pull Request Type

- [x] Runtime changes
  - [x] Other changes

### Description

lesser UBSAN fixes

### Checklist

### DIFF
--- a/lib/grdefs.h
+++ b/lib/grdefs.h
@@ -50,7 +50,7 @@
 #ifndef GRDEFS_H
 #define GRDEFS_H
 
-#include "pstypes.h"
+#include <cstdint>
 
 //	bit depth info
 #define BPP_TO_BYTESPP(x) (((x) + 7) >> 3)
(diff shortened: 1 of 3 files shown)


lib/grdefs.h
```diff
@@ -50,6 +50,7 @@
 #ifndef GRDEFS_H
 #define GRDEFS_H
 
+#include <cstdint>
 #include "pstypes.h"
```
winterheart, line 54: Including `cstdint` usually makes `pstypes.h` redundant.

lib/grdefs.h
```diff
@@ -82,9 +83,9 @@ static const ddgr_color GR_NULL = 0xffffffff, // don't do a thing with this.
 #define GR_COLOR_CHAR 1 // ASCII 1 and (r,g,b) changes current text color in string.
 
 //	MACROS
-static inline ddgr_color GR_RGB(int r, int g, int b) { return ((r << 16) + (g << 8) + b); }
+static inline ddgr_color GR_RGB(uint8_t r, uint8_t g, uint8_t b) { return ((r << 16) + (g << 8) + b); }
 
-static inline uint16_t GR_RGB16(int r, int g, int b) { return (((r >> 3) << 10) + ((g >> 3) << 5) + (b >> 3)); }
+static inline uint16_t GR_RGB16(uint8_t r, uint8_t g, uint8_t b) { return (((r >> 3) << 10) + ((g >> 3) << 5) + (b >> 3)); }
```
winterheart, line 88: Probably is better change r,g,b to `uint32_t`? All calls defines r,g,b as `int`.

winterheart: Is there any changes are planned?

jengelh: (Not sure where Github lost my comment, but here goes again: )

Changing to `uint32_t` is a *big* problem. The callsite effectively uses `char b = 255; GR_RGB16(r,g,b);`, which causes b=0xffffffff, and all those blue bits spill over into `r` and `g`, making the result of GR_RGB useless.
So no change planned, uint8_t is just right. (A function probably should not take types whose range is not handled anyway.)

winterheart:
Tested on Linux, no issues found.

PR #590, opened 2024-09-16
Add ui network settings

ccfly42:
## Pull Request Type

- [ ] GitHub Workflow changes
- [x] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [x] Other changes

### Description
The commit includes a new settings window where you can change the connection speed setting.
Connection Speed ​​sets the client-side pps for multiplayer games and is currently set to default 7 pps.

I only found one useful free space in the settings under "general". If more settings are to be added in the future, 
space can be created by reducing the spacing between existing settings.

I added the settings "Lan" and "Lan+" and removed "28k", "33k" and "Cable". These are no longer up to date and would take up space in the menu.
```
old list  	new list
name pps 	name pps
28k   5		-
33k   6		-
56k   7		56k  7
ISDN  8		ISDN 8
Cable 9		-
Fast  12	Fast 12
-		Lan  20
-		Lan+ 30
```

general
conspeed

### Checklist

### Additional Comments
-the maximum pps for the dedicated has been increased to 40.
-standard pps increased from 7 to 8 pps when not configured.


Lgt2x:
I'm wondering, is there any situation in 2024+ where you would not want the fastest out of them all? Broadband has come a long way in 25 years, I don't know how you can get your hands on 56k or ISDN nowadays! Even "Lan" and "Lan+" are not very evocative in terms of speed

ccfly42:
Yes, you're right. Other possibilities would be a slider, input field or we can hard code some value and close the pull.

Lgt2x:
I'd say let's default to the highest current possible value, and we can revisit to add settings for competitive play to increase throughput/PPS if people want that!

Lgt2x:
As a side note we could reuse the slider widget from the mouse sensitivity selection to make a PPS selector

ccfly42:
Current suggestion:
-default pps: 30 (max)
-max dedicated pps: 40
slider

Lgt2x:
Some servers run 40PPS, maybe the client-side maximum should be higher? I suppose modern broadband can support that. What happens when client PPS is higher than the server's?

ccfly42:
I chose 30 pps because I know that it works without any problems (1.4 30pps mods, piccu). 
We can increase the maximum to 40 pps or more (I would be happy about it), 
but I'm not sure if 40 pps is fine as a default value, as it hasn't been tested.

The client PPS are limited per session by the server PPS. 
When connecting, the client is informed about the server PPS. 
If the client PPS is higher than the server PPS, the client is limited to the server PPS. 
So the client cannot send more PPS than the server.

If a less conservative approach is OK, we can set the default PPS to 40 or 60 and adjust it again (increase, decrease) 
in a later release if problems occur. 
If problems happens in the meantime, the server administrators can enforce lower PPS based on the server PPS.

It should be noted that the PPS should not exceed the FPS.
Furthermore, when looking through the netcode, I discovered functions that temporarily reduce the client's PPS 
in the event of problems. However, I cannot say how well these work.

Lgt2x (approved):
Tested again on Linux, works as intended. Thanks for the great contribution! Default and maximum value can be adjusted in the future if needed.

Lgt2x:
> Furthermore, when looking through the netcode, I discovered functions that temporarily reduce the client's PPS
in the event of problems. However, I cannot say how well these work.

Definitely worth investigating, this area could also be documented and improved

PR #592, opened 2024-09-18
DallasFunc user_var object: fix GCC 9.4 compilation

Lgt2x:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

GCC 9/Ubuntu 20.04 that does not implement P2162R2, allowing to use std::visit on classes inheriting from std::variant. This addresses the issue by implementing our own visit function performing a static cast to the parent variant type

### Related Issues

Fixup #573

### Checklist

### DIFF
--- a/scripts/DallasFuncs.h
+++ b/scripts/DallasFuncs.h
@@ -8,6 +8,15 @@ class user_var : public std::variant<float, int32_t> {
 	using VarT = std::variant<float, int32_t>;
 
 	public:
+	// Required for compatibility with GCC 9/Ubuntu 20.04 that does not implement P2162R2,
+	// using std::visit on classes inheriting from std::variant
+	// (see https://gcc.gnu.org/bugzilla/show_bug.cgi?id=90943)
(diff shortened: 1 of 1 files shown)


winterheart:
If we going to support Ubuntu 20.04 LTS toolchain, then we probably need add it our GitHub CI matrix.

Lgt2x:
> If we going to support Ubuntu 20.04 LTS toolchain, then we probably need add it our GitHub CI matrix.

You're right. It will be EOL by April 2025, but apparently many people still use it as a daily driver

jengelh:
Mentioning P2162 was helpful in optimizing your 592. I have a "zero-cost" version in #593. (We can do without the extra visit() member function.)

Lgt2x:
Replaced by #593

PR #598, opened 2024-09-20
[Init] Fixed Default_pilot initialization

pzychotic:
When we have no config file yet, e.g. on first start of the game, or the entry for "Default_pilot" is missing in the config, the variable Default_pilot will be filled with random memory. This might later crash when we try to interpret this string as std::filesystem::path in PilotSelect() in pilot.cpp.

## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Documentation or Wiki changes
- [ ] Build and Dependency changes
- [x] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [x] Other changes

### Description

When we have no config file yet, e.g. on first start of the game, or the entry for "Default_pilot" is missing in the config, the variable Default_pilot will be filled with random memory. This might later crash when we try to interpret this string as std::filesystem::path in PilotSelect() in pilot.cpp.

### Related Issues

Fixes #594

### Checklist

- [x] I have tested my changes locally and verified that they work as intended.
- [x] I have documented any new or modified functionality.
- [x] I have reviewed the changes to ensure they do not introduce any unnecessary complexity or duplicate code.
- [x] I understand that by submitting this pull request, I am agreeing to license my contributions under the project's license.

Issue #594, opened 2024-09-19
[Runtime Issue]: Launching the game causes “Debug Assertion Failed! `_CrtlsValidHeapPointer(block)`”


Jayman2000:
### Build Version

v1.6 770e9e8f22db0b7596f23d8fdd4d0ff67e422737

### Operating System Environment

- [ ] Microsoft Windows (32-bit)
- [X] Microsoft Windows (64-bit)
- [ ] Mac OS X
- [ ] Linux (specify distribution and version below)

### CPU Environment

- [ ] x86 (32-bit Intel/AMD)
- [X] x86\_64 (64-bit Intel/AMD)
- [ ] ARM (32-bit)
- [ ] ARM64 (64-bit; sometimes called AArch64)
- [ ] Other (RISC V, PPC...)

### Game Modes Affected

- [X] Single player
- [X] Anarchy
- [X] Hyper-Anarchy
- [X] Robo-Anarchy
- [X] Team Anarchy
- [X] Capture the Flag
- [X] Bounty
- [X] Entropy
- [X] Hoard
- [X] Monsterball
- [X] Cooperative

### Game Environment

I’m using game data from the US Windows Dual-Jewel version of Descent 3 v1.4.

### Description

When I launch the game on Windows, I get this error:

![Microsoft Visual C++ Runtime Library&NewLine;&NewLine;Debug Assertion Failed!&NewLine;&NewLine;Program: …ous\Documents\Descent3\builds\win\Descent3\Debug\Descent3.exe&NewLine;File: minkernel\crts\ucrt\src\appcrt\heap\debug_heap.cpp&NewLine;Line: 904&NewLine;&NewLine;Expression: _CrtlsValidHeapPointer(block)&NewLine;&NewLine;For information on how your program can cause an assertion failure, see the Visual C++ documentation on asserts.&NewLine;&NewLine;(Press Retry to debug the application)&NewLine;&NewLine;Choose one of these options: Abort, Retry, Ignore](https://github.com/user-attachments/assets/b269ee88-6a30-43ca-9c5e-f860f7a11c20)

### Regression Status

I don't know

### Steps to Reproduce

1. Compile a debug build of Descent 3 on Windows.
2. Delete the `%APPDATA%\Outrage Entertainment` directory if it exists.
3. Make sure that all of the required game files are in the same directory as `Descent3.exe`.
4. Change directory into that directory.
5. Run `Descent3.exe -nointro`.

[...]

PR #599, opened 2024-09-21
Disallow exiting pilot selection without selecting a pilot

Lgt2x:
## Pull Request Type

- [ ] GitHub Workflow changes
- [x] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

Message box did not display correctly when cancelling (ESC) the pilot selection menu, allowing entering the game without a valid pilot

### Related Issues

Completes #598 

Same vein as #594 

### Checklist

### DIFF
--- a/Descent3/pilot.cpp
+++ b/Descent3/pilot.cpp
@@ -964,14 +964,12 @@ void PilotSelect() {
   NewPltUpdate(select.pilot_list, 0, pfilename);
 
   // if we get here than there is at least one pilot already
-  char old_file[_MAX_FNAME];
+  std::filesystem::path old_file;
 
   // use this in case they cancel out
   pfilename = Current_pilot.get_filename();
   if (cfexist(pfilename) != CFES_NOT_FOUND) {
-    strcpy(old_file, pfilename.c_str());
-  } else {
-    old_file[0] = '\0';
+    old_file =  std::filesystem::u8path(pfilename);
   }
 
   DoWaitMessage(false);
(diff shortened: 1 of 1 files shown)


Lgt2x:
maybe `cfexist` should return `CFES_NOT_FOUND` when filename is null, needs testing to see if it does not break anything else

Descent3/pilot.cpp
```diff
@@ -1023,7 +1023,7 @@ void PilotSelect() {
 
     case UID_CANCEL: {
       // Cancel out
-      bool found_old = (cfexist(old_file) != CFES_NOT_FOUND);
+      bool found_old = (cfexist(old_file) != CFES_NOT_FOUND && *old_file);
```
winterheart, line 1026: I'd rewrite whole logic to get rid `found_old` variable. On `old_file` initialization we already checked that is exists or not. So instead of using `found_old` we may check `old_file[0] == '\0'`.
We can even change `old_file` to `std::fs::path` and use `.empty()` for checks.

Lgt2x: you're right, I was lazy on that one :P fixed

Descent3/pilot.cpp
```diff
@@ -964,14 +964,12 @@ void PilotSelect() {
...
-  char old_file[_MAX_FNAME];
+  std::filesystem::path old_file;
 
   // use this in case they cancel out
   pfilename = Current_pilot.get_filename();
   if (cfexist(pfilename) != CFES_NOT_FOUND) {
-    strcpy(old_file, pfilename.c_str());
-  } else {
-    old_file[0] = '\0';
+    old_file =  pfilename;
```
winterheart, line 972: For Windows we need here `pfilename.u8string()` conversion.

winterheart:
Tested on Linux, no issues found.

PR #600, opened 2024-09-22
Support controller analog triggers

### LINKED ISSUE #496: [Runtime Issue]: XBox Controller registers triggers as [-1, 1] axes instead of [0, 1]
### Build Version

87f6ef00

- [X] Mac OS X

### CPU Environment

- [ ] x86 (32-bit Intel/AMD)
- [ ] ARM (32-bit)
- [X] ARM64 (64-bit; sometimes called AArch64)
- [ ] Other (RISC V, PPC...)

### Game Modes Affected

- [X] Single player
- [ ] Anarchy
- [ ] Hyper-Anarchy
- [ ] Robo-Anarchy
- [ ] Team Anarchy
- [ ] Capture the Flag
- [ ] Bounty
- [ ] Entropy
- [ ] Hoard
- [ ] Monsterball
- [ ] Cooperative

### Game Environment

_No response_

### Description

XBox Controller's "trigger buttons" are continuous-value controls, which are detected (somewhat sensibly) as joystick axes in the game. However, the game assumes that the axis range is both positive and negative, when in reality the trigger should be mapped as a positive-only axis. The result of this is that an untouched trigger is always detected as being the maximally-negative value.

### Regression Status

_No response_

### Steps to Reproduce

_No response_

Lgt2x:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [x] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

Add support for analog triggers, such as those used in the xbox controllers. They are detected as axis by SDL2, but are unidirectional, so they should be mapped to functions expecting a [-1;1] range. We allow mapping them to other buttons. Button controls have also been made analog when it made sense, so triggers can have fine control over movement.

Left to do in this MR: implement `ctTime` and `ctDownCount` for analog triggers so they can be used for weapons, refine detection threshold for mappings

The controller code is a mess, this will need other passes to make it better. 

### Related Issues

Close #496 

### Checklist

### DIFF
--- a/Descent3/Controls.cpp
+++ b/Descent3/Controls.cpp
@@ -942,7 +942,7 @@ void DoControllerMovement(game_controls *controls) {
 
   //	do button pitch
   if (ctl_pub.value > 0.0f)
-    controls->pitch_thrust -= ctl_pub.value;
+    controls->pitch_thrust -= ctl_pub.value; 
   if (ctl_pdb.value > 0.0f)
     controls->pitch_thrust += ctl_pdb.value;
 
--- a/Descent3/Controls.cpp
+++ b/Descent3/Controls.cpp
@@ -1,5 +1,5 @@
 /*
-* Descent 3 
+* Descent 3
 * Copyright (C) 2024 Parallax Software
 *
 * This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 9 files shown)


Descent3/args.cpp
```diff
@@ -114,12 +114,10 @@ int FindArg(const char *which) {
 
   for (int i = 1; i <= TotalArgs; i++) {
     if (which_matches(GameArgs[i])) {
-      LOG_INFO.printf("FindArg: Found [%s] at argument index (%d).", which, i);
```
winterheart, line 117: I would like to retain logging supplied cmd-line options from user for diagnostic purposes. This may useful to identify non-standard user environments.

Descent3/args.cpp
```diff
@@ -128,7 +126,6 @@ int FindArgChar(const char *which, char singleCharArg) {
   for (int i = 1; i <= TotalArgs; i++) {
     char *str = GameArgs[i];
     if (str[0] == '-' && str[1] == singleCharArg && str[2] == '\0') {
-      LOG_INFO.printf("FindArg: Found [-%c] at argument index (%d).", singleCharArg, i);
```
winterheart, line 131: Same here.

winterheart:
I've tested on Xbox Series controller on Linux, there some improvements compared to current main branch, no uncontrolled rotation anymore, but now my ship is constantly moving up, like left trigger constantly pushed. If I push right trigger, ship goes down, but after release ship again goes up.

Descent3/args.cpp
Lgt2x: With the check for "-nosound" I introduced it was spamming logging non-stop, it's a quick and dirty way to make it stop. I'll do it more properly using a global cache variable to disable sound, no worries

Lgt2x:
> I've tested on Xbox Series controller on Linux, there some improvements compared to current main branch, no uncontrolled rotation anymore, but now my ship is constantly moving up, like left trigger constantly pushed. If I push right trigger, ship goes down, but after release ship again goes up.

You've mapped triggers to something like pitch? It's possible that default deadzone is too high, how does it go when you increase it here ? https://github.com/DescentDevelopers/Descent3/pull/600/files#diff-7f049cd870d2530a44058475c2746ae9ba2f1bddf3995756061cf046e72fc2cfR1332

And the mapping part goes smoothly?

winterheart:
Just retested on new pilot with default configs, same thing happens.

Lgt2x:
> Just retested on new pilot with default configs, same thing happens.

Ok then that's the analog trigger detection threshold that's not set correctly. Migrating to the SDL2 GameController API will solve that

Lgt2x:
I could reproduce the problem on an actual Xbox series controller, it's a problem with the default mapping that uses a trigger as an axis. While it can be remapped easily, short-term solution is to disable this default mapping. I'll update the PR later.

Descent3/args.cpp
Lgt2x: commit dropped

Lgt2x:
Alright, finally came around to finishing this one! I'm not done at all with controllers, but at least I'm building familiarity. Triggers can be mapped to direction functions, and default mappings should not force them to be real axis anymore. XBox series controller should now work properly. Firing with trigger buttons is not supported yet.

Ready for anyone to test & review!

Lgt2x:
@winterheart could you please review and test this branch? thanks!

ddio/sdlcontroller.cpp
```diff
@@ -628,6 +627,8 @@ int sdlgameController::assign_function(ct_function *func) {
     case ctKey:
       elem.ctl[i] = 0;
       break;
+    case ctAnalogTrigger:
+      std::cout << "test";
```
winterheart, line 631: Test leftovers.

ddio/sdlcontroller.cpp
```diff
@@ -441,6 +407,37 @@ ct_config_data sdlgameController::get_controller_value(ct_type type_req) {
...
+                                                  ct_config_data *val) {
+
+  if ((m_ControlList[controllerId].flags & axis_ctf_flag) &&
+      !(m_ControlList[controllerId].axis_is_trigger & axis_ctf_flag)) {
+    float limit = (m_ControlList[controllerId].sens[axis_ct_flag - 1] > 1.5f) ? 0.95f
+                  : (m_ControlList[controllerId].sens[axis_ct_flag - 1] > 1.0f)
+                      ? 0.80f
+                      : (m_ControlList[controllerId].sens[axis_ct_flag - 1] / 2);
+    float pos = get_axis_value(controllerId, axis_ct_flag, ctAnalog);
+    if (fabs(pos) > limit)
```
winterheart, line 420: Need to use `std::abs`.

lib/joystick.h
```diff
@@ -66,6 +66,7 @@
 #define JOYSTICK_H
 
 #include <cstdint>
+#include <vector>
```
winterheart, line 69: vector declaration is unused in header.

winterheart (commented):
New code eliminates uncontrolled rotation and elevate movements. There no forward / back, up / down movements on default joystick configuration, but I think we can do more polishing in `SDL_GameController()` enabling iteration.

Lgt2x:
> New code eliminates uncontrolled rotation and elevate movements. There no forward / back, up / down movements on default joystick configuration, but I think we can do more polishing in SDL_GameController() enabling iteration.

Indeed, that's the very next steps, but out of scope for this PR. Movement mapped manually is always better than uncontrolled movement that you cannot remap. Comments fixed otherwise. Thanks for testing!

Descent3/Controls.cpp
```diff
@@ -935,40 +935,40 @@ void DoControllerMovement(game_controls *controls) {
...
 
   //	do button heading
-  if (ctl_hlb.value)
-    controls->heading_thrust += (-1.0f);
-  if (ctl_hrb.value)
-    controls->heading_thrust += (1.0f);
+  if (ctl_hlb.value > 0.0f)
+    controls->heading_thrust += ctl_hlb.value;
+  if (ctl_hrb.value > 0.0f)
+    controls->heading_thrust -= ctl_hlb.value;
```
pzychotic, line 941: Is this a copy-paste error? Shouldn't it be ctl_hrb.value?

Descent3/Controls.cpp
```diff
@@ -935,40 +935,40 @@ void DoControllerMovement(game_controls *controls) {
...
 
   //	do button banking
-  if (ctl_blb.value)
-    controls->bank_thrust += (1.0f);
-  if (ctl_brb.value)
-    controls->bank_thrust += (-1.0f);
+  if (ctl_blb.value > 0.0f)
+    controls->bank_thrust += ctl_blb.value;
+  if (ctl_brb.value > 0.0f)
+    controls->bank_thrust -= ctl_blb.value;
```
pzychotic, line 956: Same for ctl_brb.value?

pzychotic:
I tried with a PS5 controller on Windows 11.

Analog triggers are detected as normal axis. The `initialVal` in 
https://github.com/DescentDevelopers/Descent3/blob/deb9749d1f467f2202603177cb91fb43a270859f/ddio/sdljoy.cpp#L167
is always `0` for **all** axis/triggers.

If I try to force the code to go into that `if` clause for analog triggers, they don't work at all. I can't even assign them in config menu.
If I don't do that, they always fire and trying to assign a control for an axis always assigns the left trigger/ U-axis.

I haven't found anything yet, to distinguish an analog trigger from an axis for the PS5 controller.

Lgt2x:
Ugh that's a pain... No choice but to use the SDL2 game controller database to identify axes then. I'll update this PR for that.

Lgt2x:
Closing this, I'll rework the PR after we have SDL3 (#663) with the new game controller API

PR #591, opened 2024-09-17
Update DDIO and related to it's code OSIRIS

winterheart:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Documentation or Wiki changes
- [ ] Build and Dependency changes
- [x] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [x] Other changes

### Description

Remove unused functions ddio_GetParentPath() and ddio_CleanPath(). Implement new crossplatform ddio_GetTmpFileName() to replace ddio_GetTempFileName().

Optimize module loading in OSIRIRS: use std::filesystem::path on most parts, change OSIRIS_Extracted_scripts to std::map that removes some fixed size restrictions and optimize search of extracted modules.

### Checklist

- [x] I have tested my changes locally and verified that they work as intended.
- [x] I have documented any new or modified functionality.
- [x] I have reviewed the changes to ensure they do not introduce any unnecessary complexity or duplicate code.
- [x] I understand that by submitting this pull request, I am agreeing to license my contributions under the project's license.

PR #607, opened 2024-09-25
Fix debug break when starting missions with scripts

Jayman2000:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Documentation or Wiki changes
- [ ] Build and Dependency changes
- [x] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [x] Other changes

### Description

This pull request prevents this error from happening when you start a new single player game on Windows:

Descent 3 Debug Break&NewLine;&NewLine;Int3 in C:\Users\Anon Y. Mous\Documents\Descent3-backup\Descent3\OsirisLoadandBind.cpp at line 1005.&NewLine;&NewLine;It's probably safe to continue.  Continue?&NewLine;&NewLine;Choose one of these options: Yes or No.

### Related Issues

Fixes #605.

### Checklist

- [x] I have tested my changes locally and verified that they work as intended.
- [x] I have documented any new or modified functionality.
- [x] I have reviewed the changes to ensure they do not introduce any unnecessary complexity or duplicate code.
- [x] I understand that by submitting this pull request, I am agreeing to license my contributions under the project's license.

Issue #605, opened 2024-09-24
[Runtime Issue]: Debug break when starting Pilot Training


Jayman2000:
### Build Version

v1.6 c5630268c244efb8cc1f1d967c09923915958433

### Operating System Environment

- [ ] Microsoft Windows (32-bit)
- [X] Microsoft Windows (64-bit)
- [ ] Mac OS X
- [ ] Linux (specify distribution and version below)

### CPU Environment

- [ ] x86 (32-bit Intel/AMD)
- [X] x86\_64 (64-bit Intel/AMD)
- [ ] ARM (32-bit)
- [ ] ARM64 (64-bit; sometimes called AArch64)
- [ ] Other (RISC V, PPC...)

### Game Modes Affected

- [X] Single player
- [ ] Anarchy
- [ ] Hyper-Anarchy
- [ ] Robo-Anarchy
- [ ] Team Anarchy
- [ ] Capture the Flag
- [ ] Bounty
- [ ] Entropy
- [ ] Hoard
- [ ] Monsterball
- [ ] Cooperative

### Game Environment

I’m using game data from the US Windows Dual-Jewel version of Descent 3 v1.4.

### Description

Starting the Pilot Training mission on Windows will give you this error:

![Descent 3 Debug Break&NewLine;&NewLine;Int3 in C:\Users\Anon Y. Mous\Documents\Descent3-backup\Descent3\OsirisLoadandBind.cpp at line 1005.&NewLine;&NewLine;It's probably safe to continue.  Continue?&NewLine;&NewLine;Choose one of these options: Yes or No.](https://github.com/user-attachments/assets/7cf0393f-2972-4e53-93c4-78a1a35cc0f6)

### Regression Status

Newest commit that doesn’t have the problem: 2d29dd4f7c76bb9474103435d587dae61b922dfb
Oldest commit that does have the problem: 09ab370821ee37cc3bea66e43a34bb21f792d503

### Steps to Reproduce

1. Start a debug build of Descent 3 on Windows.

2. Once you get to the main menu, click “New Game”.

3. If you’re prompted to select a mission, then select “Pilot Training”.

winterheart:
This Int3() happens on inability to load level module. Did you copied platform-dependent `d3-win.hog` into Base_directory (where Descent3.exe is placed)?

Jayman2000:
[...]

PR #603, opened 2024-09-23
cfile: Replace cf_LibraryFindFirst() with cf_DoForeachFileInLibrary() 

winterheart:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Documentation or Wiki changes
- [x] Build and Dependency changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

* Remove unused `cf_ChangeFileAttributes()`
* Replace bulky `cf_LibraryFindFirst()` / `cf_LibraryFindNext` with `cf_DoForeachFileInLibrary()`, which simplifies code.

### Checklist

- [x] I have tested my changes locally and verified that they work as intended.
- [x] I have documented any new or modified functionality.
- [x] I have reviewed the changes to ensure they do not introduce any unnecessary complexity or duplicate code.
- [x] I understand that by submitting this pull request, I am agreeing to license my contributions under the project's license.

PR #609, opened 2024-09-26
Fix level loading

pzychotic:
## Pull Request Type

- [ ] GitHub Workflow changes
- [x] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [x] Other changes

### Description

I believe some changes from #591 broke loading levels or anything that loads dynamic libs from hog files (at least on Windows). Instead of loading the new x64 DLLs from d3-win.hog, the old x86 DLLs from the original game data are loaded, which fails spectacularly.

There is this comment still, that isn't true anymore: https://github.com/DescentDevelopers/Descent3/blob/3a7c86a2d257718bc9c6071fc6007a839ee6f3b7/Descent3/init.cpp#L1504-L1505

`Osiris_ExtractScriptsFromHog()` will happily overwrite existing entries of foo.dll in `OSIRIS_Extracted_scripts` with the last loaded temp filename from either d3.hog or any .mn3 file loaded at level start.

Trying to fix the problem, I asked myself: Why do we even load the old dynamic libs at all? We extract both old and new libs to the temp directory, only to ignore the old ones and only load the new ones. We can skip all that and only extract and load the new ones. Which this PR tries to do.

On Linux/Mac we might have a subtle difference. Since the original builds are already x64 and loading the old dynamic libs might just work, without us realizing that we don't load our new/fixed code from d3-linux|mac.hog.

### Checklist

### DIFF
--- a/Descent3/Mission.cpp
+++ b/Descent3/Mission.cpp
@@ -1811,7 +1811,6 @@ bool mn3_Open(const char *mn3file) {
   char pathname[_MAX_PATH];
   char filename[PSFILENAME_LEN + 1];
   char ext[PSFILENAME_LEN];
-  int mn3_handle;
   // concatanate the mn3 extension if it isn't there.
   char tempMn3File[_MAX_PATH];
   if (!IS_MN3_FILE(mn3file)) {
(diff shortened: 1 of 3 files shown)


winterheart:
This issue addressed by #603, one of commits removes extraction from outdated hogs and prevents extraction for already extracted scripts.

pzychotic:
Oh, I wasn't aware that we already have a related issue #605 for this and even a fix in #607.

@winterheart Can you please incorporate my 2nd commit from here (https://github.com/DescentDevelopers/Descent3/pull/609/commits/a9f335a0121e9aae386a99f20a2934c6c480315a), since the MAX_LOADED_MODULES limit was increased only to account for the loading old **and** new libs from hog files, which we don't do anymore.

PR #613, opened 2024-09-29
Enhance mission levels select window

winterheart:
## Pull Request Type

- [ ] GitHub Workflow changes
- [x] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [x] Other changes

### Description

This PR was inspired by #485. Now missions that contains more that one level, shows enhanced level select window (see screenshot). This enhancement allows user to select available level instead of input number of mission manually. Additional refactoring made for filename paths used in related functions.

Descent3_mission_levels

### Checklist

### DIFF
--- a/Descent3/LoadLevel.cpp
+++ b/Descent3/LoadLevel.cpp
@@ -3588,6 +3588,86 @@ static inline char* GetCurrentSumString() {
   return output_buf;
 }
 
+bool LoadLevelInfo(const std::filesystem::path &filename, level_info &info) {
+  CFILE *ifile = cfopen(filename, "rb");
+  bool found = false;
+
+  if (!ifile) {
+    LOG_ERROR.printf("Failed to open mission file %s", filename.u8string().c_str());
+    return false;
+  }
+
+  try {
+    // catch cfile errors
+
+    // Read & check tag
+    char tag[4];
+    cf_ReadBytes((uint8_t *)tag, 4, ifile);
+    if (strncmp(tag, LEVEL_FILE_TAG, 4)) {
+      LOG_ERROR.printf("%s is not a level file (tag %c%c%c%c)",
+                       filename.u8string().c_str(), tag[0], tag[1], tag[2], tag[3]);
+      cfclose(ifile);
+      return false;
+    }
+
+    // Read & check version number
+    int version = cf_ReadInt(ifile);
+
+    // Check for too-new version
+    if (version > LEVEL_FILE_VERSION) {
+      LOG_ERROR.printf("Mission file %s too new (version %d)", filename.u8string().c_str(), version);
+      cfclose(ifile);
+      return false;
+    }
+    if (version < LEVEL_FILE_OLDEST_COMPATIBLE_VERSION) {
+      LOG_ERROR.printf("Mission file %s too old (version %d)", filename.u8string().c_str(), version);
+      cfclose(ifile);
+      return false;
+    }
+
+
+    // Init level info
+    strcpy(info.name, "Unnamed");
+    strcpy(info.designer, "Anonymous");
+    strcpy(info.copyright, "");
+    strcpy(info.notes, "");
+
+
+    while (!cfeof(ifile)) {
+      char chunk_name[4];
+
+      cf_ReadBytes((uint8_t *)chunk_name, 4, ifile);
+      chunk_start = cftell(ifile);
+      chunk_size = cf_ReadInt(ifile);
+
+      if (ISCHUNK(CHUNK_LEVEL_INFO)) {
+        cf_ReadString(info.name, sizeof(info.name), ifile);
+
+        // Localize level name here...
+        strcpy(info.name, LocalizeLevelName(info.name));
+
+        cf_ReadString(info.designer, sizeof(info.designer), ifile);
+        cf_ReadString(info.copyright, sizeof(info.copyright), ifile);
+        cf_ReadString(info.notes, sizeof(info.notes), ifile);
+        found = true;
+        break;
+      } else {
+        cfseek(ifile, chunk_start + chunk_size, SEEK_SET);
+      }
+    }
+
+  } catch (cfile_error *cfe) {
+    LOG_FATAL.printf("Error reading: file = <%s>, error = \"%s\"", cfe->file->name, cfe->msg);
+    ASSERT(cfe->read_write == CFE_READING);
+    cfclose(ifile);
+    return false;
+  }
+
+  cfclose(ifile);
+  return found;
+}
+
+
 extern bool Disable_editor_rendering;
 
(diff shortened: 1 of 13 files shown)


JeodC:
Nice work, it looks great! Tested in Windows. Do you mind if the core missions go to the top of the list? See [here](https://github.com/JeodC/Descent3/commit/071f0182b2f7b4ee88399cf67c823e1a7de2219a) for reference. I also really like the idea of the `LEVEL SELECT` header being the name of the current mission, as a reference for the user. This was done in the old PR that inspired this current one.

image

winterheart:
> Nice work, it looks great! Tested in Windows. Do you mind if the core missions go to the top of the list? See [here](https://github.com/JeodC/Descent3/commit/071f0182b2f7b4ee88399cf67c823e1a7de2219a) for reference. I also really like the idea of the `LEVEL SELECT` header being the name of the current mission, as a reference for the user. This was done in the old PR that inspired this current one.

I'm planning to rewrite "New Game" UI flow and divide Campaign and Mission select to different screens. Regarding "Level select" title I'll add name of mission.

winterheart:
Changed "Select Level" with mission name.

Descent3/LoadLevel.cpp
```diff
@@ -3588,6 +3588,86 @@ static inline char* GetCurrentSumString() {
   return output_buf;
 }
 
+bool GetLevelInfo(const std::filesystem::path &filename, level_info &info) {
```
Lgt2x, line 3591: `GetLevelInfo` is also a DMFC function with a different signature that returns a `level_info` object. I would suggest renaming this function to avoid possible confusion to something like `LoadLevelInfo`, as this function loads and retrieves the level info from file.

Descent3/LoadLevel.cpp
```diff
@@ -3588,6 +3588,86 @@ static inline char* GetCurrentSumString() {
...
+  CFILE *ifile = cfopen(filename, "rb");
+  bool found = false;
```
Lgt2x, line 3593: variable is only used much later, can be declared before the `while` loop below

Descent3/LoadLevel.cpp
```diff
@@ -3588,6 +3588,86 @@ static inline char* GetCurrentSumString() {
...
+    return false;
+  }
+
+  try {
+    // catch cfile errors
+
+    // Read & check tag
+    char tag[4];
+    cf_ReadBytes((uint8_t *)tag, 4, ifile);
+    if (strncmp(tag, LEVEL_FILE_TAG, 4)) {
```
Lgt2x, line 3606: can be extracted to a new function. This operation is also done for `LoadLevel`. Same for level file version check and initialization of level info

Descent3/LoadLevel.h
```diff
@@ -634,6 +635,15 @@
 // 130->131 Added the powerups ignore gravity checkbox
 // 131->132 Added another ff bounce texture for Dan (his last day wish)
 
+/**
+ * Get information from level such as level name, description, author and notes.
```
Lgt2x, line 639: description is part of `level_info` struct

Descent3/Mission.cpp
```diff
@@ -948,11 +944,12 @@ bool LoadMission(const char *mssn) {
   srclinenum = 1;
   numlevels = -1;
   curlvlnum = 0;
-  indesc = 0;
+  indesc = false;
   cur_objective = -1;
   msn->multiplayable = true;
   msn->singleplayable = true;
   msn->training_mission = false;
+  msn->levels.reserve(100); // Reserve some place before we know actual number of levels
```
Lgt2x, line 952: maybe 100 is a little too much, base game has much fewer

Descent3/Mission.cpp
```diff
@@ -994,16 +991,16 @@ bool LoadMission(const char *mssn) {
...
       } else if (!stricmp(command, "URL")) {
         if (curlvlnum != 0) {
           strcpy(errtext, TXT_MSN_LVLCOMMAND);
           goto msnfile_error;
         } else {
-          for (int a = 0; a < MAX_MISSION_URL_COUNT; a++) {
-            if (Net_msn_URLs.URL[a][0] == '\0') {
-              strncpy(Net_msn_URLs.URL[a], operand, MAX_MISSION_URL_LEN - 1);
-              Net_msn_URLs.URL[a][MAX_MISSION_URL_LEN - 1] = '\0';
+          for (auto &a : Net_msn_URLs.URL) {
```
Lgt2x, line 1000: rename `a` -> `url`

Descent3/Mission.h
```diff
@@ -308,13 +322,18 @@ bool LoadMissionLevel(int level);
...
-void ShowProgressScreen(const char *str, const char *str2 = NULL, bool flip = true);
+void ShowProgressScreen(const char *str, const char *str2 = nullptr, bool flip = true);
 
 // See if a mission file is multiplayer playable.
 bool IsMissionMultiPlayable(const char *mission);
 
-//	return information about a mission
-bool GetMissionInfo(const char *msnfile, tMissionInfo *msn);
+/**
+ * Returns information about a mission
```
Lgt2x, line 331: "Fill mission information" as we don't return the mission info object

Descent3/menu.cpp
```diff
@@ -1090,25 +1073,23 @@ static inline int count_missions(const std::vector<std::filesystem::path> &missi
...
-            }
-          }
-        });
+          if (stricmp(path.filename().u8string().c_str(), "d3_2.mn3") == 0)
+          return;
+        if (GetMissionInfo(path.filename(), &msninfo) && msninfo.name[0] && msninfo.single) {
+          filelist.push_back(path.filename());
+          lb->AddItem(msninfo.name);
+          c++;
+          if (!(c % 2))
```
Lgt2x, line 1089: I know this code was there before, I don't see why do we display wait message only for even `c`

Lgt2x (commented):
Love this new feature! Tested with success on Linux. I also have a request, not necessarily for this MR: it could be great to display the number of levels of a mission in the "info" dialog below the "notes" header when selecting the mission.

Descent3/LoadLevel.cpp
winterheart: There try-catch block, we cannot declare variable inside it.

Descent3/menu.cpp
winterheart: Just saving CPU cycles on fetching mission list.

winterheart:
> I also have a request, not necessarily for this MR: it could be great to display the number of levels of a mission in the "info" dialog below the "notes" header when selecting the mission.

That would be kinda spoiler for player, he will know how many missions are left.

Lgt2x (approved):
Tested again on Linux, I think we're good now!

Issue #618, opened 2024-10-05
There should only be one list of command-line options


Jayman2000:
At the moment, both [`USAGE.md`](https://github.com/DescentDevelopers/Descent3/blob/7ca92bd1bd31db1500baca1b1423435f2db5ecaa/USAGE.md#command-line-options) and [the wiki](https://github.com/DescentDevelopers/Descent3/wiki/Running-Descent-3#command-line-options) have lists of command-line options for Descent 3. It would be better if there was just one list instead of two.

In the PR description for #471, I said that I would open a pull request that adds documentation for `-additionaldir` if #471 got merged. Now that #471 has been merged, it’s not clear to me if I should add the documentation to `USAGE.md` , the wiki or both. If there was only one list of command-line options, then it would be clear where documentation for new command-line options should go.

winterheart:
`USAGE.md` is main doc for users, cmd-line options included. New documentation related to user interaction should be reflected in `USAGE.md`.

Lgt2x:
The wiki lists the command-line options available when the project started. Since then, several have been removed and added. The source of truth should be `USAGE.md`, and the wiki should be used for developer docs.

It could be its own issue, but the game itself should know what options it supports and be able to display a help menu when launched with `--help`/`-h`.

Jayman2000:
This issue was fixed when the patch from #624 was accepted.

PR #623, opened 2024-10-06
Copy command-line documentation into `USAGE.md` & add documentation for `-additionaldir`

Jayman2000:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

This pull request takes [the table of command-line options that’s on the wiki](https://github.com/DescentDevelopers/Descent3/wiki/Running-Descent-3#command-line-options) and merges it with [the one in `USAGE.md`](https://github.com/DescentDevelopers/Descent3/blob/48c87750f49b8f9cf4a24d3911fb70218e60a0cc/USAGE.md#command-line-options). It also adds documentation for the `-additionaldir` command-line option.

### Related Issues

This is related #618, but it doesn’t fully fix that issue.

### Checklist

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -80,6 +80,37 @@ This error means that game data could not be found. Make sure you copied all
 game files to the `D3-open-source` folder, and that you're running the game
 from this same folder.
 
+## Base directories
+
+A base directory is a directory that Descent 3 expects game files to be in. When you run Descent 3, it will try to access many different files. Most of those files need to be stored in a base directory. There are two different types of files that are stored in base directories:
+
+- Read-write files are files that can change while you play Descent 3. Examples: `<you name>.plt` and files in the `savegame/` directory.
+- Read-only files are files that do not change while you play Descent 3. Examples: `d3.hog` and files in the `movies/` directory.
+
+Descent 3 has two types of base directories:
+
+- The writable base directory can contain both read-write files and read-only files. There is only one writeable base directory. By default, the writable base directory gets set to the current working directory.
+- The read-only base directories can only contain read-only files. There can be any number of read-only base directories. By default, Descent 3 uses zero read-only base directories.
+
+You can set the writable base directory and the list of read-only base directories using the `-setdir`, `-useexedir` and `-additionaldir` command-line options (see [the next section](#command-line-options)).
+
+When Descent 3 tries to find a read-only file, then it will look through the list of base directories in this order:
+
+- the last read-only base directory that was specified on the command-line,
+- the second-to-last read-only base directory that was specified on the command-line,
+- the third-to-last read-only base directory that was specified on the command-line,
+- …
+- the first read-only base directory that was specified on the command-line and, finally,
+- the writable base directory.
+
+Files that are in base directories that are higher on that list will override files that are in base directories that are lower on that list. For example, lets say that you run Descent 3 like this:
+
+```
+Descent3 -setdir /home/user/my-writable-base-directory -additionaldir /home/user/my-read-only-base-directory
+```
+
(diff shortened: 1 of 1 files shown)


USAGE.md
```diff
@@ -35,17 +35,395 @@ Descent 3 Message(Error: Couldn't find the string table.)
...
+Descent3 -setdir /home/user/my-writable-base-directory -additionaldir /home/user/my-read-only-base-directory
+```
+
+Let’s also say that both `my-writable-base-directory` and `my-read-only-base-directory` contain a file named `d3.hog`. In this example, Descent 3 will load `/home/user/my-read-only-base-directory/d3.hog` because read-only directories have a higher precedence than the writable base directory. Descent 3 will ignore `/home/user/my-writable-base-directory/d3.hog`.
+
+## Command-Line Options
+
+The following command-line options are available in Descent 3. You can set command-line options on the Misc. tab of the Setup section of the Descent 3 launcher or by creating a shortcut to `Descent3.exe`. Case is not significant in command-line options, and `-`, `--`, and `+` are all accepted.
+
+<table>
```
winterheart, line 73: Please use markdown-style syntax for table. `USAGE.md` file is included to source tarball, available to users as file and must be easy to read.

Jayman2000: Do you really think that a Markdown table would be easy to read? I’m looking at [the table in `BUILD.md` right now](https://github.com/DescentDevelopers/Descent3/blob/48c87750f49b8f9cf4a24d3911fb70218e60a0cc/BUILD.md?plain=1#L161-L171), and it’s kind of difficult to read because the lines are over 400 characters long. The table in this PR would be even harder to read because the lines would be even longer. Plus, both of the text editors that I tried using defaulted to using word wrap, so I had to disable word wrap before I could start trying to read the table that was in `BUILD.md`.

Lgt2x: I agree that HTML formatting for a file we ship in releases is not ideal, it's likely to be opened raw in a text editor. To be honest I'm not fond of md tables either, a good old list `- option: description` is as easy to read on a simple text editor :D

USAGE.md
```diff
@@ -35,17 +35,395 @@ Descent 3 Message(Error: Couldn't find the string table.)
...
+  </tr>
+  <tr>
+    <td><code>-height &lt;height&gt;</code></td>
+    <td>integer</td>
+    <td>480 unless you select a different value in the options menu</td>
+    <td>all</td>
+    <td>Sets the screen resolution to the specified height, if possible.</td>
+  </tr>
+  <tr>
+    <td><code>-himem</code></td>
```
Lgt2x, line 106: Outside of the scope of this PR, but some of these old options could be cleaned up, they really are not relevant to the modern-age player

USAGE.md
Jayman2000: @winterheart

> Do you really think that a Markdown table would be easy to read?

winterheart: We can reorder table into plain bullet list like in Linux man pages.

Jayman2000: > We can reorder table into plain bullet list like in Linux man pages.

OK, I just pushed a new version of this PR that turns the HTML table into a few Markdown bullet lists. How do you think it looks now?

USAGE.md
Jayman2000: I agree that it would be good to clean up some of the old command-line options, and I agree that it’s out of scope for this PR. To make sure that we don’t forget about that idea, I opened #638.

Jayman2000:
I pushed a new version of this PR that resolves merge conflicts. The new version also adds a new commit titled “Add missing &lt;placeholder&gt; for -display”.

Lgt2x (approved):
I like the options displayed this way, thanks for undertaking this documentation update. I'll check again the `Base directories` section when we want to merge #628

PR #626, opened 2024-10-07
Fix undefined behavior in memory allocation

sebholt:
## Pull Request Type

- [ ] GitHub Workflow changes
- [x] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [x] Other changes

### Description

Currently `mem_rmalloc` uses `std::malloc` for memory allocation and the deallocation is performed in `mem_free`
which calls `::free`. This does seem to work but is **undefined behavior** because memory allocated with `std::malloc`
should be freed with `std::free` not `::free`.

We should use either 
- the C interface: `::malloc` and `::free`, or
- C++ interface: `std::malloc` and `std::free`.

The `mem_malloc/mem_free` functions already use the C interface so it makes sense to let `mem_rmalloc` simply call
`mem_malloc` instead of `std::malloc` directly. There's also no speed difference because everything is inline or a macro. 

This changes the `mem_rmalloc` function to call `mem_malloc` instead of `std::malloc` to match the `mem_free` call.

### Checklist

### DIFF
--- a/mem/mem.h
+++ b/mem/mem.h
@@ -76,17 +76,6 @@
 #include <cstdlib>
 #include <type_traits>
 
-template<typename T> static inline T *mem_rmalloc()
-{
-	static_assert(std::is_trivially_constructible_v<T> && std::is_trivially_destructible_v<T>);
-	return static_cast<T *>(std::malloc(sizeof(T)));
-}
-template<typename T> static inline T *mem_rmalloc(std::size_t nelem)
-{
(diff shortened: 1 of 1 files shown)


pzychotic:
LGTM
But since this code was just recently introduced by @jengelh, I would like to get his opinion before merging.

sebholt:
Ok. I've updated the PR description to make the issue more clear.

jengelh:
It's a sound argument.

pzychotic:
Thanks both of you!

PR #627, opened 2024-10-08
Fix pilot cpp

sebholt:
## Pull Request Type

- [ ] GitHub Workflow changes
- [x] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [x] Other changes

### Description

This fixes access to uninitialized string buffers in pilot.cpp in two cases that are very similar.
From earlier commits an uninitialized string buffer remained but was used nonetheless in logging functions where
a `std::string`, that was introduced meanwhile, should have been used instead.

This fixes both cases and removes the now unused string buffer.

### Checklist


Descent3/pilot.cpp
```diff
@@ -2816,20 +2815,20 @@ void ShipSelectDeleteLogo(newuiListBox *lb) {
...
   // Get the filename
-  std::filesystem::path p = Custom_images[selected_index - 1];
+  std::filesystem::path custom_filename = Custom_images[selected_index - 1];
 
   // delete custom_filename, we don't want it....
   char buffer[512];
   snprintf(buffer, sizeof(buffer), TXT_PLTOKDEL, custom_logoname);
   if (DoMessageBox(TXT_PLTDELCONF, buffer, MSGBOX_YESNO, UICOL_WINDOW_TITLE, UICOL_TEXT_NORMAL)) {
-    LOG_INFO.printf("Deleting pilot logo %s (%s)", custom_logoname, custom_filename);
+    LOG_INFO.printf("Deleting pilot logo %s (%s)", custom_logoname, custom_filename.c_str());
```
pzychotic, line 2824: Please use `custom_filename.u8string().c_str()` in all log messages, otherwise Windows `wchar_t` based `std::fs::path`'s will not be displayed correctly.

sebholt: Thanks for the feedback. I've added `u8string()` in the logging calls.

pzychotic: Thanks for for your fix.

PR #629, opened 2024-10-12
Re-enable building the editor

pzychotic:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

`CMAKE_SYSTEM_NAME` isn't defined before the `project` call, so moved the option below it. Actually moved all options to keep them together and hopefully prevent similar problems in the future when adding new stuff.
Also put all the editor related projects into a single `if` scope.

This broke already a while ago, when the vcpkg changes where introduced and the `project` call moved down.
I just noticed it now after deleting the whole build directory and starting fresh, since the option was still in my CMakeCache all that time.

### Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,19 +5,9 @@ if(NOT CMAKE_BUILD_TYPE AND NOT DEFINED ENV{CMAKE_BUILD_TYPE})
   set(CMAKE_BUILD_TYPE "Debug" CACHE STRING "default build type")
 endif()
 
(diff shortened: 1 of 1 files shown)


winterheart:
I remember fixing this once before, probably in one of my countless branches that didn't make it into the main branch...

PR #635, opened 2024-10-20
Hogmaker: Only rebuild the hog file if any input file changed

sebholt:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

This lets the hogmaker (re)build the hogfile only if
- the hog doesn't exist
- the hogmaker executable is newer than the hog
- the inputs source file is newer than the hog
- any of the input files is newer than the hog

otherwise only "Already up to date" is printed.

In the build process this
- reduces the build log spamming by hogmaker, especially in repeated builds where the hog doesn't change
- reduces the filesystem io because the hog isn't overwritten with the same content all the time
- keeps the hog timestamp as old as possible

Also some unused code is removed from the hogmaker.

### Checklist

### DIFF
--- a/tools/HogMaker/HogMaker.cpp
+++ b/tools/HogMaker/HogMaker.cpp
@@ -25,19 +25,13 @@
 #include "d3_version.h"
 #include "HogFormat.h"
 
-// TODO: To our descendants from the future: remove it when code will support C++20
-template <typename TP> std::time_t to_time_t(TP tp) {
-  auto sctp = std::chrono::duration_cast<std::chrono::system_clock::duration>(tp - TP::clock::now()) +
-              std::chrono::system_clock::now();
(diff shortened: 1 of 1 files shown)


winterheart:
I would like to retain ability to generate hog file no matter of what in cmd-line. Still, I understand your intentions and can propose alternative based on properly cmake dependencies detection. In result all hog files will be generated only if any of packed files was changed.

sebholt:
> Still, I understand your intentions and can propose alternative based on properly cmake dependencies detection.

That'd be great. I'd prefer a CMake based solution which integrates into the build system more nicely.

sebholt:
I'm closing this PR because we have a better CMake based solution in #637.

Issue #631, opened 2024-10-14
[Runtime Issue]: Support for Windows XP


own3mall:
### Build Version

1.5 latest release

### Operating System Environment

- [X] Microsoft Windows (32-bit)
- [X] Microsoft Windows (64-bit)
- [ ] Mac OS X
- [ ] Linux (specify distribution and version below)

### CPU Environment

- [X] x86 (32-bit Intel/AMD)
- [X] x86\_64 (64-bit Intel/AMD)
- [ ] ARM (32-bit)
- [ ] ARM64 (64-bit; sometimes called AArch64)
- [ ] Other (RISC V, PPC...)

### Game Modes Affected

- [X] Single player
- [ ] Anarchy
- [ ] Hyper-Anarchy
- [ ] Robo-Anarchy
- [ ] Team Anarchy
- [ ] Capture the Flag
- [ ] Bounty
- [ ] Entropy
- [ ] Hoard
- [ ] Monsterball
- [ ] Cooperative

### Description

I tried to get the Descent3.exe binary to work in Windows XP by following the instructions in a reddit comment from here (https://www.reddit.com/r/windowsxp/comments/1dckc7b/is_not_a_valid_win32_application/):

Specifically:

> If you want to try editing it, open the EXE in some hexeditor, find bytes 50 45 (PE) near the beginning of the file, and exactly 62 bytes after that you'll probably see sequence 06 00 00 00 (meaning Vista), then some 4 bytes (probably 00) and then again 06 00 00 00. Change both those occurrences to 05 00 01 00 (meaning XP).

That gets me past the "is not a valid Win32 application" error, but then I get :  

> This application has failed to start because the application configuration is incorrect.  Reinstalling the application max fix this problem.

Yet, I've installed Visual C++ 2015-2019 (last supported XP version).

Could you please support XP?  The original Descent 3.exe build does...

### Steps to Reproduce

Try to run on XP 32-bit and XP x64 (64-bit).  Doesn't work.  Original binary works in both versions.

winterheart:
Hello.

This task may be difficult to achieve. Currently we fully support only Windows 64bit target. 32bit target still buildable (as last I checked some time ago), but not tested. Main difficulties here is using correct toolchain. You definitely need MSVC 2019 as project requires C++17 capable compiler. Other things is remember to install MS Visual C++ redistributable runtime package.

Lgt2x:
[...]

PR #640, opened 2024-10-27
Fix Windows x86 build

pzychotic:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

This fixes compilation of Windows x86 builds when creating template instantiations of `FnPtr` with functions using `__stdcall`.
The x64 case works because the `decltype` of `void __stdcall Fun(...)` is the same as `void Fun(...)` so the missing `GLFUNCCALL` in the declaration does not matter.
With this change I was able to run a local x86 build of the game.

### Related Issues

Related to #631

### Checklist

### DIFF
--- a/renderer/dyna_gl.h
+++ b/renderer/dyna_gl.h
@@ -63,7 +63,7 @@ inline void CheckError() {
 
 template <typename T> struct FnPtr;
 template <typename Ret, typename... Args>
-struct FnPtr<Ret(Args...)> {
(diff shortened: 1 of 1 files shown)


Lgt2x (approved):
Good catch :)

PR #637, opened 2024-10-20
Optimize making HOG files

winterheart:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Documentation or Wiki changes
- [x] Build and Dependency changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

Optimizing HOG files building by using `add_custom_command` instead of `add_custom_target`, which not enforces always rebuilding HOGs. In result this reduces incremental build time and optimize building process.

### Related Issues

Fixes #635.

### Checklist

- [x] I have tested my changes locally and verified that they work as intended.
- [x] I have documented any new or modified functionality.
- [x] I have reviewed the changes to ensure they do not introduce any unnecessary complexity or duplicate code.
- [x] I understand that by submitting this pull request, I am agreeing to license my contributions under the project's license.

PR #642, opened 2024-10-29
Introduce cross-compiled ARM64 build

Lgt2x:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

Using the HogMaker build changes of #637 , introduce an example of cross-compilation to ARM64. This also serves as non-regression automated build for this type of compilation.

The changes to `vcpkg.json` are due to a [VCPKG issue](https://github.com/microsoft/vcpkg/issues/40031) causing dbus cross-compilation to fail. The best solution would be to report the issue upstream to dbus, but for the time being, we simply disable the optional SDL2 dependency to dbus.

`find_package(HogMaker)` is now in top-level `CMakeLists.txt`, because otherwise the exported targets would not be found.

The changes
### Checklist


.github/workflows/build.yml
```diff
@@ -83,7 +92,17 @@ jobs:
         with:
           vcpkgJsonGlob: vcpkg.json
 
+      - name: Configure cross-compiled build
+        if: ${{ matrix.os.preset == 'linux-cross-arm64' }}
+        run: |
+          cmake --preset linux
+          pushd builds/linux/
+          ninja -f build-${{ matrix.build_type }}.ninja HogMaker
+          popd
```
winterheart, line 101: This can be simpified with
```
ninja -f build-${{ matrix.build_type }}.ninja -C builds/linux/ HogMaker
```

Lgt2x: fixed

PR #645, opened 2024-11-03
Resolve 3301 compiler warnings

jengelh:
## Pull Request Type

### Description

Compiler warnings point out qualitative issues with the code. gcc emits roughly 3860 warnings on Linux-amd64.
This batch addresses the top two categories,

```
scripts/Y2K.cpp:467:5: warning: deleting object of polymorphic class type "LevelScript_0000" which has non-virtual destructor might cause undefined behavior [-Wdelete-non-virtual-dtor]
      467 |     delete ((LevelScript_0000 *)ptr);
scripts/LEVEL0.cpp:1036:27: warning: unused variable "event_data" [-Wunused-variable]
```

### Checklist


Descent3/OsirisLoadandBind.cpp
```diff
@@ -2683,7 +2683,7 @@ bool Osiris_RestoreSystemState(CFILE *file) {
     return false;
   }
 
-  uint8_t version = (uint8_t)cf_ReadByte(file);
+  // uint8_t version = (uint8_t)cf_ReadByte(file);
```
sebholt, line 2686: Is it safe to remove the byte read? All subsequent reads will at an other position than before.

jengelh: Good catch, I fixed it now.

Lgt2x (approved):
Thanks for taking the time to solve warnings, that's very appreciated. Splitting in multiple commits will help if we ever need to revert something that unexpectedly causes mayhem. In the meantime, I did not see any change that could be a problem. Runtime is fine too.

PR #651, opened 2024-11-26
MacOS CI builds: use MacOS-13 instead of the deprecated MacOS-12

Lgt2x:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Related Issues

See https://github.com/actions/runner-images/issues/10721

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,7 +26,7 @@ jobs:
             cc: cl
             cxx: cl
             name: Windows-x64
-          - runner: macos-12 # This is supposed to be Intel for now, and what macos-latest is defaulting to for some reason (as of 04/2024)
+          - runner: macos-13 # Should be Intel
             preset: mac
             cc: cc
             cxx: c++


.github/workflows/build.yml
```diff
@@ -26,7 +26,7 @@ jobs:
             cc: cl
             cxx: cl
             name: Windows-x64
-          - runner: macos-12 # This is supposed to be Intel for now, and what macos-latest is defaulting to for some reason (as of 04/2024)
+          - runner: macos-3 # Should be Intel
```
halprin, line 29:
```suggestion
          - runner: macos-13 # Should be Intel
```

Lgt2x: thanks, fixed

halprin (approved):
Checks out.  Both the debug and release versions marked as Intel say they are x86_64 applications.

PR #653, opened 2024-12-02
ddio/lnxmouse.cpp: Code compression using lambdas

sebholt:
## Pull Request Type

- [ ] GitHub Workflow changes
- [x] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [x] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

In ddio/lnxmouse.cpp there's a lot of code duplication that differs only by an integer or two. The given patches introduce lambdas that that are called instead of using duplicated inline code. This reduces the code size and helps to avoid errors in duplicated code.

### Checklist

### DIFF
--- a/ddio/lnxmouse.cpp
+++ b/ddio/lnxmouse.cpp
@@ -192,73 +192,44 @@ void ddio_MouseSetVCoords(int width, int height) { ddio_MouseSetLimits(0, 0, wid
 int sdlMouseButtonDownFilter(SDL_Event const *event) {
   ASSERT(event->type == SDL_MOUSEBUTTONDOWN);
 
-  const SDL_MouseButtonEvent *ev = &event->button;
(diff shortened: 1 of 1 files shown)


Lgt2x:
Does this actually fix anything? I tend to be a little conservative when it comes to refactoring sections of the code that have no issues in the first place, it can sneakily introduce hard to notice problems without real benefits other than DRY-er code.

sebholt:
It doesn't fix any runtime issue, it's purely code cosmetics. I see how this might introduce unforeseen issues although the changes are not too intrusive. The long term motivation for the changes was to make mouse wheel events first class citizens in the engine and this was a first code cleanup step. But actually I'm too busy to work on that atm so I'm closing this PR.

PR #588, opened 2024-09-16
netcon update

winterheart:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Documentation or Wiki changes
- [x] Build and Dependency changes
- [x] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [x] Network changes
  - [ ] Other changes

### Description

Replace old inetgetfile with new httpclient that allows download missions and other files via cpp-httplib library. That fixes thread safety errors in Descent3 Online netcon module. Reactivated ability download missing mission files from game browser.

### Checklist

- [x] I have tested my changes locally and verified that they work as intended.
- [x] I have documented any new or modified functionality.
- [x] I have reviewed the changes to ensure they do not introduce any unnecessary complexity or duplicate code.
- [x] I understand that by submitting this pull request, I am agreeing to license my contributions under the project's license.

PR #662, opened 2025-02-14
Netcon update

Lgt2x:
Rebased and fixed version of #588


Lgt2x:
self-approving this as this is mostly not my code

PR #663, opened 2025-02-14
Updgrade to SDL3

Lgt2x:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

Now that SDL3 has a stable release, it's about time we upgrade our main dependency ! This will unlock [new features](https://wiki.libsdl.org/SDL3/NewFeatures) that we could use, such as the GPU API or the Dialog API.

This upgrade was performed using the [migration guide](https://github.com/libsdl-org/SDL/blob/main/docs/README-migration.md) and Python renaming script for functions and macros that changed name.

Due to an issue in upstream dbus (which is now required for SDL3) when cross-compiling, I had to create an overlay port inspired by vcpkg#40038.

I also tried to make third-party management more manageable so that developers only have 2 options to get non-vendored dependencies: either use vcpkg or install dependencies themselves so that `find_package` finds them automatically. There used to be a 3rd option for plog that could use the code from a submodule. For consistency, it was removed as well as the `USE_EXTERNAL_PLOG` cmake option.

Gamma correction was also removed because SDL3 does not support it.


ddio/sdljoy.cpp
```diff
@@ -318,19 +318,22 @@ void joy_GetPos(tJoystick joy, tJoyPos *pos) {
 static int joyGetNumDevs(void) {
   int found = 0;
 
+  int joyCount = 0;
+  SDL_GetJoysticks(&joyCount);
```
pzychotic, line 322: We might have a little mem leak here. [SDL_GetJoysticks](https://wiki.libsdl.org/SDL3/SDL_GetJoysticks) states that the returned array "should be freed with [SDL_free](https://wiki.libsdl.org/SDL3/SDL_free)() when it is no longer needed."

renderer/HardwareOpenGL.cpp
```diff
@@ -430,24 +422,39 @@ int opengl_Setup(oeApplication *app, const int *width, const int *height) {
         LOG_WARNING << "No parameter for -display given";
       } else {
         int arg_index = atoi(arg_index_str);
-        int display_count = SDL_GetNumVideoDisplays();
+        int display_count = 0;
+        SDL_GetDisplays(&display_count);
```
pzychotic, line 426: We might have a little mem leak here. [SDL_GetDisplays](https://wiki.libsdl.org/SDL3/SDL_GetDisplays) states that the array returned "should be freed with [SDL_free](https://wiki.libsdl.org/SDL3/SDL_free)() when it is no longer needed".

ddio/lnxkey_sdl.cpp
```diff
@@ -339,38 +339,38 @@ static inline uint8_t sdlkeycode_to_keycode(uint32_t sdlkeycode) {
...
   uint8_t kc = 0;
 
-  if ((event->type != SDL_KEYUP) && (event->type != SDL_KEYDOWN))
-    return (1);
+  if ((event->type != SDL_EVENT_KEY_UP) && (event->type != SDL_EVENT_KEY_DOWN))
+    return true;
 
-  switch (event->key.state) {
-  case SDL_PRESSED:
+  switch (event->key.down) {
```
pzychotic, line 348: Since `event->key.down` is a bool now, the `switch` seems odd and might be changed to an `if`.

pzychotic (commented):
Good job, really like the removed submodules :)

ddio/sdljoy.cpp
Lgt2x: indeed, it should be fixed now!

ddio/lnxkey_sdl.cpp
Lgt2x: reworked to use if/else logic

pzychotic:
Two things I noticed testing on Win 11:
- The game is a lot darker now and the brightness slider in settings doesn't do anything. Maybe related to the removed gamma correction?!
- I have a 4K display with 150% display scaling in Windows settings. Previously the game window was scaled accordingly, now it's not. Might be related to SDL3s new [high DPI support](https://wiki.libsdl.org/SDL3/README/highdpi).

Lgt2x:
> Two things I noticed testing on Win 11:
> 
>     * The game is a lot darker now and the brightness slider in settings doesn't do anything. Maybe related to the removed gamma correction?!
> 

Definitely related to gamma correction. The SDL functions related to gamma correction did not work great on Linux in my experience, and they were applied to the entire screen, which is less than ideal. This is part of the reason why they were [eventually deleted](https://github.com/libsdl-org/SDL/pull/6617). In my opinion in-game brightness selection should be implemented at the shader level if we want it, or we can let the user configure their monitor settings themselves using the grey box displayed in the brightness settings, and remove the slider (long gone are the CRT days...). I'm unsure currently how to handle this properly.


>     * I have a 4K display with 150% display scaling in Windows settings. Previously the game window was scaled accordingly, now it's not. Might be related to SDL3s new [high DPI support](https://wiki.libsdl.org/SDL3/README/highdpi).

I did not notice that because I run my monitor at 100% scaling, but it should be fixed now. We're using `SDL_GetDisplayContentScale` to scale the window according to the display scaling factor.

Lgt2x:
Gamma correction is now applied by shaders. This is also better solution than using the native SDL function because it is now applied only to the game window and not the whole screen. The setting is now enabled on Unix platforms, and the "apply" button really works. I'm quite satisfied by this new solution. Can you please test again @pzychotic ? :)

PR #667, opened 2025-02-20
Add HD resolution & FOV settings

### LINKED ISSUE #205: [Feature Request] Support for 1080p and Higher Resolutions
Please make sure that you guys add support for 1080p and higher resolutions.

Lgt2x:
## Pull Request Type

- [ ] GitHub Workflow changes
- [x] Runtime changes
  - [x] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

This adds support for setting higher resolutions, as well as the FOV angle. The projection math has been corrected so the aspect ratio is automatically computed given window dimensions.

Possible improvements (probably not in this PR):
 - scale the HUD text to the resolution
 - scale the reticle
 - scale menu

This work is derived from the [Piccu Engine](https://github.com/InsanityBringer/PiccuEngine) that has done this before.

### Related Issues

Closes #205 

Descent 3 running in glorious 2540*1440
res1

New video configuration menu
res2

res3

### DIFF
--- a/Descent3/config.cpp
+++ b/Descent3/config.cpp
@@ -747,6 +747,13 @@ struct video_menu {
     if (GetScreenMode() == SM_GAME) {
       Render_preferred_state.bit_depth = Render_preferred_bitdepth;
       rend_SetPreferredState(&Render_preferred_state);
+
+      SetScreenMode(GetScreenMode(), true);
+
+      int temp_w = Video_res_list[Game_video_resolution].width;
+      int temp_h = Video_res_list[Game_video_resolution].height;
+      Current_pilot.set_hud_data(NULL, NULL, NULL, &temp_w, &temp_h);
+  
     }
 
     Render_FOV_setting = static_cast<float>(fov[0]) + D3_DEFAULT_FOV;
@@ -791,18 +798,6 @@ struct video_menu {
         int newindex = resolution_list->GetCurrentIndex();
         if (static_cast<size_t>(newindex) < Video_res_list.size()) {
           Game_video_resolution = newindex;
-          int temp_w, temp_h;
-          int old_sm = GetScreenMode();
-
-          if (old_sm == SM_GAME) {
-            SetScreenMode(SM_NULL);
-            SetScreenMode(old_sm, true);
-          }
-
-          temp_w = Video_res_list[Game_video_resolution].width;
-          temp_h = Video_res_list[Game_video_resolution].height;
-          Current_pilot.set_hud_data(NULL, NULL, NULL, &temp_w, &temp_h);
-
           std::string res = Video_res_list[Game_video_resolution].getName();
           snprintf(resolution_string, res.size() + 1, res.c_str());
         }
--- a/Descent3/config.cpp
+++ b/Descent3/config.cpp
@@ -303,6 +303,7 @@
 #include "sounds.h"
 #include "ctlconfig.h"
 #include "d3music.h"
+#include "gameloop.h"
 
 #include <algorithm>
 #include <vector>
@@ -336,6 +337,7 @@ std::vector<tVideoResolution> Video_res_list = {{512, 384},
 
 const int DEFAULT_RESOLUTION = 7; // 1280x720
 int Game_video_resolution = DEFAULT_RESOLUTION;
+float Render_FOV_setting = 72.0f;
 
 tDetailSettings Detail_settings;
 int Default_detail_level = DETAIL_LEVEL_MED;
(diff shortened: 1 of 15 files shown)


JeodC:
Tested with 2560x1440 native. Level 5 still renders the reactor percentages at Y-center, which is what I considered first on seeing this. The resolution picker has options not available on my device, however, can they be retrieved from the OS?

Descent3/config.cpp
```diff
@@ -303,14 +303,41 @@
...
+#include <algorithm>
+#include <vector>
 
-tVideoResolution Video_res_list[N_SUPPORTED_VIDRES] = {
-    {512, 384}, {640, 480}, {800, 600}, {960, 720}, {1024, 768}, {1280, 960}, {1600, 1200}, {640, 480} // custom
-};
+#define STAT_SCORE STAT_TIMER
 
-int Game_video_resolution = 1;
+std::vector<tVideoResolution> Video_res_list = {{512, 384},
```
JeodC, line 313: I think ideally these options should be retrieved from the operating system and populated as such, using the custom list as a fallback. Thoughts?

halprin:
Works overall on macOS.  I set it to 2560x1440 and looks lovely.

I run into a non-ideal situation when I subsequently change the resolution while in game, though.  This is from starting out in 2560x1440.

1. Press F2 to bring up the video settings.
2. Change the resolution to 1280x720.
3. See image 1 below.  Notice that the settings window moves to the corner.  Not ideal.
4. Thankfully, I can still click on "Change" for the resolution.
5. See image 2 below.  Thankfully, the window for the resolutions are centered.
6. I change the resolution back to 2560x1440.
7. See image 3 below.  Yay, the video settings window is back being centered.


Image 1.
Screenshot 2025-02-20 at 5 42 02 PM

Image 2.
Screenshot 2025-02-20 at 5 42 16 PM

Image 3.
Screenshot 2025-02-20 at 5 42 30 PM

Descent3/config.cpp
Lgt2x: that's a good idea: default resolution is the "native" screen resolution, and the user can otherwise choose from the pre-defined list.

Lgt2x:
the menu placement is less than ideal indeed, I'll see what I can do.

Thank you guys for testing!

Descent3/config.cpp
JeodC: Well, no, I mean than at the moment I can choose resolutions beyond what my monitor will output, and I assume in some cases choosing one of those options might end up causing a user to get "stuck" with a black screen or a blown-up viewport that puts several items offscreen where they may not be able to correct their mistake.

I mean literally populate the list by querying the OS/GPU for the available resolutions and use the pre-defined list as a fallback if the query fails.

pzychotic: [SDL_GetFullscreenDisplayModes()](https://wiki.libsdl.org/SDL3/SDL_GetFullscreenDisplayModes) seems to be the way in SDL3 to query all the supported display resolutions.

Lgt2x: From my testing, going to a higher resolution than your screen does not do anything unexpected... What I'm unsure about is using this approach with multiple monitors that do not have the same resolutions. What happens if the game window is created on the "wrong" monitor and the game does not suggest the right resolutions?

Lgt2x:
@halprin that should be fixed, we now only change resolution when exiting the config menu

Descent3/config.cpp
Lgt2x: I've iterated a little more on this, and ended with this solution (implemented in last commit): in a new function run at game launch, `ConfigureDisplayResolutions`, we iterate display modes with `SDL_GetFullscreenDisplayModes()` on all available displays, only keep unique (width;height) pairs, and use the current display mode of the default display as the default resolution. `-width` and `-height` arguments create a new resolution if it does not exist yet and select it. The current resolutions list is used as a fallback only.

Let me know what you think!

halprin (approved):
> @halprin that should be fixed, we now only change resolution when exiting the config menu

I tested on a standard 1440p monitor and Apple Retina monitor (HiDPI).  Both work well on the Mac!  The resolution switching works fine too.

Descent3/config.cpp
Lgt2x: @JeodC @pzychotic did you have the change to test this new version? =)

Lgt2x: Good catch, I'll update that. Thanks for testing!

Lgt2x:
We have 2 approvals, let's merge this

PR #669, opened 2025-02-24
macOS Code Signing and Notarization

### LINKED ISSUE #540: [Runtime Issue]: MacOS signing configuration
### Build Version

v1.5

- [X] Mac OS X

### CPU Environment

- [ ] x86 (32-bit Intel/AMD)
- [ ] ARM (32-bit)
- [X] ARM64 (64-bit; sometimes called AArch64)
- [ ] Other (RISC V, PPC...)

### Game Modes Affected

- [ ] Single player
- [ ] Anarchy
- [ ] Hyper-Anarchy
- [ ] Robo-Anarchy
- [ ] Team Anarchy
- [ ] Capture the Flag
- [ ] Bounty
- [ ] Entropy
- [ ] Hoard
- [ ] Monsterball
- [ ] Cooperative

### Game Environment

_No response_

### Description

MacOS builds are unsigned. In order to sign them, we need an Apple Developer account to associate with the project and sign them with. This entails both a financial and legal responsibility for the project. Additionally, having only one contributor control the signing certificate would pose a significant operational risk: We should strive to always have at least two contributors/maintainers with access to certificate management.

### Regression Status

_No response_

### Steps to Reproduce

_No response_

### LINKED ISSUE #559: [Runtime Issue]: Unable to run on MacOS ARM - “Descent3” is damaged and can’t be opened. You should move it to the Trash.
### Build Version

v1.5.0

- [X] Mac OS X

### CPU Environment

- [ ] x86 (32-bit Intel/AMD)
- [ ] ARM (32-bit)
- [X] ARM64 (64-bit; sometimes called AArch64)
- [ ] Other (RISC V, PPC...)

### Game Modes Affected

- [X] Single player
- [ ] Anarchy
- [ ] Hyper-Anarchy
- [ ] Robo-Anarchy
- [ ] Team Anarchy
- [ ] Capture the Flag
- [ ] Bounty
- [ ] Entropy
- [ ] Hoard
- [ ] Monsterball
- [ ] Cooperative

### Game Environment

_No response_

### Description

Setting the executable bit seems to break the binary signature, so the came can't run.

### Regression Status

_No response_

### Steps to Reproduce

Download the release
Install as per instructions
Set executable bit as per readme with `chmod 755 Descent3.app/Contents/MacOS/Descent3`
Try to run, either via terminal or Finder. Get “Descent3” is damaged and can’t be opened. You should move it to the Trash.

halprin:
## Pull Request Type

- [x] GitHub Workflow changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

The macOS binaries are now code signed and notarized.

There are two ways this is accomplished.
1. The hog files that contain dynamic libraries.  These libraries must be signed before packaging into the hog file (because you can't just sign the hog file), so they are signed during the build.  This includes `d3-osx.hog` and everything in the `online` folder.
2. The main application and raw binaries in the `netgames` folder.  These binaries are signed after the build using a helper GitHub action.  This helper GitHub action also packages everything into a DMG disk image and notarizes the disk image.  A DMG is an established package format and natively supports notarization, unlike a zip file.

Code signing and notarization only happens on push to the `main` branch, not in a pull request.  [Secrets can't be accessed in a pull request from a fork](https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions#using-secrets-in-a-workflow), but I also chose to refactor the GitHub action workflows to explicitly _not_ pass in secrets into the pull request workflow to make it obvious. Let me know if we don't like this, but I thought the explicit approach was better than implicit.  We can also leverage this approach in the future for versioned releases.

Updated the documentation to remove the work-arounds needed when the application was unsigned.

If you want to inspect a test build, you can see [this build in my fork](https://github.com/halprin/Descent3/actions/runs/13488119174).  You can't depend on the build on this PR because, as mentioned above, we aren't signing and notarizing for PR builds.

### Related Issues

Fixes #540.
Fixes #559.

### Checklist


.github/workflows/build.yml
```diff
@@ -50,6 +57,9 @@ jobs:
 
     runs-on: ${{ matrix.os.runner }}
 
+    env:
+        MACOS_SIGNING: ${{ secrets.MACOS_SIGNING_IDENTITY != '' }}
```
tophyr, line 61: nit: might be better to include the os preset check in here as well? then the `if` clauses in the following changes could check just one master variable. thoughts?
```suggestion
        MACOS_SIGNING: ${{ matrix.os.preset == 'mac' && secrets.MACOS_SIGNING_IDENTITY != '' }}
```

.github/workflows/build.yml
```diff
@@ -103,12 +113,19 @@ jobs:
           cp ./arm64-osx.cmake ./community/universal-osx.cmake
           sed -i '' 's/^set(VCPKG_OSX_ARCHITECTURES.*$/set(VCPKG_OSX_ARCHITECTURES "arm64;x86_64")/' ./community/universal-osx.cmake
 
+      - name: Import macOS code signing certificate
+        if: ${{ matrix.os.preset == 'mac' && env.MACOS_SIGNING == 'true' }}
+        uses: apple-actions/import-codesign-certs@v3
```
tophyr, line 118: worth noting just for diligence reasons: `apple-actions` appears to be a non-Apple-owned repository, and compromise of the actions coming from that repository (and thus compromise of our certificates) is something we should consider in our threat model. in the long term, it might be better to fork that repository into one controlled exclusively by @DescentDevelopers/d3-maintainers and use that.

.github/workflows/build.yml
```diff
@@ -121,6 +138,25 @@ jobs:
         # There no cmake install presets so install in traditional way
         run: cmake --install builds/${{ matrix.os.preset }}/ --config ${{ matrix.build_type }}
 
+      - name: Sign, package, and notarize for macOS
+        if: ${{ matrix.os.preset == 'mac' && env.MACOS_SIGNING == 'true' }}
+        uses: halprin/macos-sign-package-notarize@v2
```
tophyr, line 143: same as above, passing secrets into an action not hosted exclusively by @DescentDevelopers/d3-maintainers theoretically poses a risk. long-term we should probably fork this and take control of it.

.github/workflows/build.yml
```diff
@@ -1,18 +1,25 @@
...
+      MACOS_APP_STORE_CONNECT_KEY:
+        required: false
+      MACOS_APP_STORE_CONNECT_KEY_ID:
+        required: false
+      MACOS_APP_STORE_CONNECT_ISSUER_ID:
+        required: false
+
 
 jobs:
+
```
tophyr, line 22: style nit: the extra newline here seems odd to me. is this a yaml thing i'm not familiar with?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,18 @@
...
+      - '**/README.md'
+      - '**/LICENSE'
+
+
+jobs:
+
+  build:
+    name: Build for PR
+    uses: ./.github/workflows/build.yml
+    # explicitly not passing secrets into the build
```
tophyr, line 18: double-checking: pull requests originating from outside the `Descent3` repository will never get access to secrets, no matter what - correct? ie, even if the PR modifies this file, or even if someone _merges_ a modification to this file?

_(is it even possible to submit a PR to yourself? suppose i could try that out, but kinda irrelevant either way)_

USAGE.md
```diff
@@ -63,12 +63,7 @@ conflicts.
     - On Linux, `cd` to `D3-open-source` and run `./Descent3`. Wayland users
       may need to set environment variable `SDL_VIDEODRIVER=wayland` before
       launching the game.
-    - On macOS, the `.app` bundle is currently not signed, so your operating
-      system will not let you run it by double-clicking it. To remediate that,
-      open your terminal and `cd` to `D3-open-source`. Run
-      `xattr -c ./Descent3.app`, `xattr -c ./netgames/*`,
-      `chmod +x ./Descent3.app/Contents/MacOS/Descent3`, and then run the game
-      using `./Descent3.app/Contents/MacOS/Descent3`
+    - On macOS, open the terminal, `cd` to `D3-open-source`, and run `./Descent3.app/Contents/MacOS/Descent3`.
```
tophyr, line 66: 😍

tophyr (approved):
Thank you, thank you THANK YOU!!! This will really help the user experience on Mac. I noted some threat-model risks that we should mitigate, but I don't think they're critical to do before merging this.

.github/workflows/build.yml
Lgt2x: I forked both under the DescentDevs namespace, @halprin  should have access to both when they accept the organization invite.

https://github.com/DescentDevelopers/import-codesign-certs
https://github.com/DescentDevelopers/macos-sign-package-notarize

Is there anything more we need to do to make the repositories a github action?

.github/workflows/cd.yml
```diff
@@ -0,0 +1,23 @@
+name: Continuous Delivery
+
+
+on:
+  push:
+    branches:
+      - main
```
Lgt2x, line 7: good idea to totally separate the signed and not signed workflows

.github/workflows/build.yml
```diff
@@ -103,12 +113,19 @@ jobs:
...
+          p12-file-base64: ${{ secrets.MACOS_SIGNING_CERTIFICATE_P12 }}
+          p12-password: ${{ secrets.MACOS_SIGNING_CERTIFICATE_P12_PASSWORD }}
+
       - name: Configure CMake
         if: ${{ matrix.os.preset != 'linux-cross-arm64' }}
         env:
           CC: ${{ matrix.os.cc }}
           CXX: ${{ matrix.os.cxx }}
-        run: cmake --preset ${{ matrix.os.preset }} -DBUILD_TESTING=ON -DENABLE_LOGGER=ON -DFORCE_PORTABLE_INSTALL=ON -DBUILD_EDITOR=ON -DUSE_EXTERNAL_PLOG=ON
+        run: cmake --preset ${{ matrix.os.preset }} -DCODESIGN_IDENTITY=${{ secrets.MACOS_SIGNING_IDENTITY }} -DBUILD_TESTING=ON -DENABLE_LOGGER=ON -DFORCE_PORTABLE_INSTALL=ON -DBUILD_EDITOR=ON -DUSE_EXTERNAL_PLOG=ON
```
Lgt2x, line 128: The `CODESIGN_IDENTITY` CMake cache variable should be declared as such in the top-level `CMakeLists.txt`: `set(CODESIGN_IDENTITY ... CACHE STRING ...)` like we do for `USE_VCPKG` for example. It should also be documented in `BUILD.md`

netcon/descent3onlineclient/CMakeLists.txt
```diff
@@ -27,6 +27,11 @@ target_link_libraries(Descent3_Online_TCP_IP PRIVATE
 target_include_directories(Descent3_Online_TCP_IP PRIVATE ${SDL3_INCLUDE_DIRS})
 if(CMAKE_SYSTEM_NAME STREQUAL "Darwin")
   set_target_properties(Descent3_Online_TCP_IP PROPERTIES SUFFIX ".dylib")
+  if(DEFINED CODESIGN_IDENTITY AND NOT "${CODESIGN_IDENTITY}" STREQUAL "")
```
Lgt2x, line 30: We could make that couple lines a CMake function or macro that takes the target name as argument

Lgt2x (commented):
I tested your artifact, it works great! That's some seriously cool work you did there :)

I invited you to the github org so you can push the secrets you need

.github/workflows/build.yml
tophyr: @tophyr : "sOmEoNe ShOuLd Do ThIs"
also @tophyr : _doesn't do it himself_
@Lgt2x : "...ok"

@halprin i'm gonna request changes on this PR to use the @DescentDevelopers versions of those actions, but the actions themselves look fine

tophyr (changes requested):
Please use the https://github.com/DescentDevelopers/import-codesign-certs and https://github.com/DescentDevelopers/macos-sign-package-notarize actions instead of external ones, now that @Lgt2x has forked them.

.github/workflows/build.yml
```diff
@@ -1,18 +1,25 @@
 name: Descent 3 Build
 
+
 on:
-  workflow_dispatch:
```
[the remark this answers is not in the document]
halprin: Yeah, the `ci.yml` and `cd.yml` independently call to `build.yml`.

.github/workflows/build.yml
halprin: We can't lump them together because sometimes we need to build for non-Mac, build for the Mac but not sign, and build for the Mac and sign.  If we combined the variables, then we couldn't sometimes build for the Mac but not sign or sometimes build for the Mac and sign.

.github/workflows/build.yml
halprin: I just added an extra newline.  If you like, I can delete it.

.github/workflows/build.yml
halprin: Fixed.

.github/workflows/build.yml
halprin: It worked, surprisingly.  Also, that `halprin` guy is pretty shady who made that https://github.com/halprin/macos-sign-package-notarize action.  Anyway, fixed!

.github/workflows/build.yml
tophyr: It would just be `MACOS_SIGNING=false` in those cases, wouldn't it? I'm just noticing that all three times `MACOS_SIGNING` is referenced, it's also and'ed with the `os.preset == 'mac'` check. If we move the `os.preset == 'mac'` check into the `MACOS_SIGNING` variable, we can remove it from the three usage sites and prevent mixups in the future if someone alters this and forgets to check it.

.github/workflows/build.yml
tophyr: no worries, it's consistent, that's really all that matters

.github/workflows/build.yml
halprin: Oooo, maybe not fixed.  In fact, the build is failing.  Sad.

halprin: I tried futzing with it but couldn't get it working.  I'll try again tomorrow.

.github/workflows/build.yml
halprin: Oh, I see.  Hmmm.  But even with this change we would continue to use `if: ${{ matrix.os.preset == 'mac' }}` like when we call `Install macOS Rosetta 2` and a couple others places, right?

.github/workflows/ci.yml
halprin: I'll do an actual live test of this tomorrow.

.github/workflows/build.yml
Lgt2x: I'll look into it tonight and try to help you out

CMakeLists.txt
```diff
@@ -5,6 +5,8 @@ if(NOT CMAKE_BUILD_TYPE AND NOT DEFINED ENV{CMAKE_BUILD_TYPE})
   set(CMAKE_BUILD_TYPE "Debug" CACHE STRING "default build type")
 endif()
 
+set(CODESIGN_IDENTITY "" CACHE STRING "Sets the macOS code signing identity.  If set to something besides the empty string "", then the dynamic libraries put into the hog files will be signed using this identity.")
```
Lgt2x, line 8: this does not work because of the double quotes inside of the documentation string `""`, CMake does not like that

netcon/descent3onlineclient/CMakeLists.txt
Lgt2x: You can define a function like this in the top level CMakelists:
```cmake
set(CODESIGN_IDENTITY "" CACHE STRING "Sets the macOS code signing identity. If set to something besides the empty string, then the dynamic libraries put into the hog files will be signed using this identity.")
function (macos_sign target)
  message ("target: ${target}")
  if(DEFINED CODESIGN_IDENTITY AND NOT "${CODESIGN_IDENTITY}" STREQUAL "")
  message(STATUS "Code signing ${target}")
  add_custom_command(TARGET ${target} POST_BUILD
         COMMAND codesign --verbose --sign "${CODESIGN_IDENTITY}" --force --timestamp --deep -o runtime $<TARGET_FILE:${target}>)
  endif()  
endfunction()
```

and use it as `macos_sign("Parallax_Online")` for instance

.github/workflows/build.yml
Lgt2x: see suggestions below

halprin: Thanks for the help!  Fixed.

netcon/descent3onlineclient/CMakeLists.txt
halprin: Thanks for the help!  Fixed.

CMakeLists.txt
halprin: Thanks for the help there!

.github/workflows/ci.yml
halprin: The live test was a success.  I confirmed GitHub's documentation to be correct.  A PR from a fork has _no_ access to the secrets in the base repository.

.github/workflows/build.yml
halprin: Fixed.

Lgt2x (approved):

Lgt2x:
Good for me, are the secrets all set on this repository so we can merge @halprin ?

halprin:
We're ready to merge!

Issue #664, opened 2025-02-14
Linux (MX23.5, Debian 12 base, Xfce DE): Fn, Alt Fn keyboard shortcuts


kliamr:
For example, F4 opens a dropdown terminal in MX Linux Xfce, so I cannot access the Guidebot menu as D3 does not 'intercept' this and some other Fn keys. Likewise the Alt Fn keys to save / load a game. I can however use F1 and F2 menus, and F9 quick save.

So it seems that some Fn keys are 'intercepted' and others are not. A workaround might be to delete all the Xfce keyboard shortcuts but this seems extreme when other games do correctly intercept those keys and not pass them through to the OS. 

I'm probably using the wrong terminology but hopefully what I'm describing is clear enough and there is a solution.

Lgt2x:
Hey, thanks for the report! I could reproduce the issue after installing XFCE and setting the F4 key to trigger an action at the desktop-environment level. We're probably missing an event filter somewhere, I'll try to find a fix.

kliamr:
Thank you Louis for your fast response, I hope it's not too much work to fix!

Lgt2x:
@kliamr #665 should fix it, can you please test it? :) DXX-rebirth uses [the same trick](https://github.com/dxx-rebirth/dxx-rebirth/blob/master/common/arch/sdl/event.cpp#L281)

kliamr:
Hi Louis, although I have installed the required dependencies including **libcpp-httplib-dev** I am still getting the following error:

kliamr@mx:~/Descent/Descent3-keyboard-grab
$ cmake --preset linux
-- The C compiler identification is GNU 12.2.0
-- The CXX compiler identification is GNU 12.2.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
CMake Error at CMakeLists.txt:164 (find_package):
  By not providing "Findhttplib.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "httplib", but
  CMake did not find one.

  Could not find a package configuration file provided by "httplib" with any
  of the following names:

    httplibConfig.cmake
[...]

PR #675, opened 2025-02-28
Miscellaneous build & CI improvements

### LINKED ISSUE #673: [Build Failure]: In-Source `CMake` build fails with `HogMaker` binary/directory name conflict
### Build Version

d48b0f1d7f46888c62b7c8876f92abae8a2cbf77

- [x] Mac OS X
- [x] Linux (specify distribution and version below)

### CPU Environment

- [x] x86 (32-bit Intel/AMD)
- [x] x86\_64 (64-bit Intel/AMD)
- [x] ARM (32-bit)
- [x] ARM64 (64-bit; sometimes called AArch64)
- [x] Other (RISC V, PPC...)

### Description

HogMaker output binary conflicts with existing HogMaker directory

### Regression Status

_No response_

### Steps to Reproduce

```sh
cd Descent3 && cmake . && make
```

Lgt2x:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

This PR fixes a few build and CI problems reported over time:
 - Explicitly forbid in-source builds (close #673 )
 - Find `cpp-httplib` using only the header of the (header-only) library if the CMake config file is not found, for distributions that don't package it (looking at you, Debian), reported in #664 by @kliamr
 - Get wider glibc compatibility on Linux by building from ubuntu 20.04 containers instead of 24.04. To get an up-to-date CMake > 3.20, I had to use a specialized action. Descent3 artifact built by CI can now run in Debian Buster and Ubuntu 20.04, which was not possible before (reported in #664)
 - Remove the clang build, that cannot run properly with ubuntu 20.04. It is less useful to have this build now, with MacOS also using a variation of clang. If we wanted to keep it, we would need to use ubuntu 22.04, and the artifact would have a different glibc compatibility version, which could be confusing.
 - Remove clang format check, which was actually never used


Lgt2x:
Waiting for sdl 3.2.6 update in vcpkg that should fix the arm64 build

Lgt2x:
Okay build is passing, this is ready for review for anyone willing to!

Now also closes #676 mouse issue, we were missing a few X11 development libs that VCPKG did not indicate. Also updating SDL3 to 3.2.6 to fix the arm64 build; in short, a problem with the signal.h header was fixed recently.

Lgt2x:
Let's not keep the Linux CI build broken on `main` for too long, merging this.

Issue #620, opened 2024-10-06
Restore the ability to run 32-bit level scripts


Lgt2x:
This has been mentioned multiple times, so here is a master issue to talk about it and suggest possible solutions.

The problem is the following: Descent 3 relies on pre-compiled scripts inside of levels to run the logic: open doors, pick-up objects, etc. For a long time now, we're compiling the main campaign's level scripts and put them in an OS-dependent file: `d3-linux.hog`/`d3-osx.hog`/`d3-win.hog`. This means that level scripts should be compiled again when the game ABI is updated.

Outside of the main campaigns, many user levels have been created over the years using the Descent 3 level editor, generating level scripts binary blobs that are now incompatible with the current version of the game. Especially, the game is now compiled exclusively to 64-bit architectures, instead of the original 32-bit it originally ran on. This means that user-created level scripts that have not been compiled recently against the 64-bit Descent 3 game cannot be played at all.

So far, it has been proposed to use a virtual machine able to run 32-bit black-box scripts/blobs, in particular using the [Box86](https://github.com/taviso/loadlibrary) or [loadlibrary](https://github.com/taviso/loadlibrary) projects, but no proof of concept has been made. More investigation is needed to find the best solution we can get to get old user-created levels to run again.

Related to #583 among others

JeodC:
Since you have the osiris source code, even from the 32bit era, is it possible to create a conversion tool to decompile and recompile scripts? If so, I would keep such a thing separate from the game itself, so external websites can perform the conversions and host levels, and the game will not be bogged down with the process during runtime.

ccfly42:
The best option is of course to load the 32-bit versions. 
Whether a web service or the level designers themselves 
update their scripts to 64-bit, we should set conditions for this so that older versions (32-bit) are not affected. 
We should make sure that the 64-bit script files do not replace the 32-bit files, 
otherwise we will create confusion and incompatibilities.

I don't know exactly whether the script file is loaded via the .d3l file name or whether the
script file name is entered into the .d3l file. If the former is the case,
a postfix "e.g. {filename}_x64.dll/.so" should be used.

Until the problem is fixed, I recommend making a 32-bit version available on a regular basis.
According to #631 , this seems to work. On the one hand, singleplayer players could play their levels again
until the problem is fixed, and on the other hand, server administrators could specifically use 32-bit versions
for multiplayer levels that contain 32-bit scripts.

SiriusTR:
[...]

PR #678, opened 2025-03-05
Osiris: only load first-party scripts

Lgt2x:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [x] Other changes

### Description

Disable loading scripts (.dll,.so,.dylib) from .mn3 files, to avoid 3rd-party levels abusing the scripting system. These scripts, compiled for Win32, would not work anyway. The only scripts we want to load are first-party scripts that we compile as part of 'd3-<platform>.hog'.

### Related Issues

Related to #620 (but not quite loading anything new)

### DIFF
--- a/Descent3/Mission.cpp
+++ b/Descent3/Mission.cpp
@@ -670,6 +670,7 @@
 #include "BOA.h"
 #include "terrain.h"
 #include "multi.h"
+#include "module.h"
 #include "hud.h"
 #include "localization.h"
 #include "levelgoal.h"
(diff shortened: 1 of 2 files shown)


Lgt2x:
@Jayman2000 can you please test this against your custom mission and confirm 3rd party code is not loaded anymore?

Jayman2000 (approved):
I can confirm that third-party party code is not loaded anymore.

Lgt2x:
Thanks for testing. I've tested that 1st party level scripts still run properly on Windows and Linux, let's merge this.

PR #681, opened 2025-03-10
Add Fullscreen toggle

Lgt2x:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Runtime changes
  - [x] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

Add a fullscreen toggle button, inspired by Piccu Engine.

Also fix a minor issue with #667 where the resolution read from the database would not be applied.

Remove unused lnxscreenmode. Some noise in the diff due to clang-format.

d3fs


halprin (commented):
There's a bug when the game resolution is not set to your display's native resolution and you enter into a game.  The viewport is surrounded by black.

# Prerequisites
1. Your game resolution setting is something smaller than your display's resolution.  E.g. your display is set to 2560x1440, but the game is set to 1280x720.

# Steps to Reproduce
1. Launch game.  It opens in fullscreen by default.
2. Choose a new game from the main menu.
3. Select a mission and start.
4. The viewport is surrounded by black even though I have fullscreen enabled.  This is incorrect.

Screenshot 2025-03-12 at 7 48 13 PM

5. If I quit the mission at this point, even the main menu viewport is "shrunk" and surrounded by black now.
6. This step doesn't matter whether I quit to the main menu or stayed in-game.  I go into options and disable fullscreen, the game goes into window mode and fills the entire window (which is correct).  If I then go back into options and re-enable fullscreen, the viewport now takes up the entire display and doesn't have the black surround (which is correct).

# Expected Behavior

I expected the viewport to take up the entire display when in fullscreen mode all the time.  There should never be any black surrounding the viewport while in fullscreen.

Lgt2x:
Hey @halprin , I could not reproduce the issue very reliably, I'm not sure where that comes from... I'm now forcing resolution change when entering a level, is it any better?

halprin:
It is not any better.  :(

halprin:
If it helps, I'm happy to jump into Discord and share my screen to show what I'm experiencing.

Lgt2x:
I had only tested on Linux and Windows, and it turns out it's a MacOS specific bug... I'll setup a Mac debugging environment and try to fix it.

Unrelated, but do you also get systematic assertions in the sound mixer on mac in debug builds? I get
```
Assertion failure at Initialize (/Users/louis/dev/Descent3/sndlib/mixer.cpp:54), triggered 1 time:
 'false'
```

I never get those on other platforms

halprin:
> Unrelated, but do you also get systematic assertions in the sound mixer on mac in debug builds? I get
> 
> ```
> Assertion failure at Initialize (/Users/louis/dev/Descent3/sndlib/mixer.cpp:54), triggered 1 time:
>  'false'
> ```
> 
> I never get those on other platforms
> 
> [EDIT] probably my junky VM mac setup

Just to confirm, I do not get this assertion error.

Lgt2x:
Okay it should be better now, I fixed a few bugs related to resolution change and full-screen mode, in game and on the menu.

halprin:
> Okay it should be better now, I fixed a few bugs related to resolution change and full-screen mode, in game and on the menu.

Works for me!

Albeit, I found one more bug.  I don't know if it is Mac specific.  While in mission, change the resolution _and_ the full screen option.  Click Done.  It looks like the resolution changes, but the full screen choice doesn't take effect.

Lgt2x:
Nice! I fixed the problem you mentionned

halprin:
> Nice! I fixed the problem you mentionned

Almost there!  The resolution and fullscreen changes together now, but the window size (when in window mode) does not.  If I disable fullscreen and reduce the resolution a lot (just so it is easy to see a difference), the window does not shrink to match the resolution.

Lgt2x:
There is something weird with the MacOS windowing system, again... That's the only system where I could reproduce the problem you describe. When setting to windowed from fullscreen and lowering the resolution, my laggy VM seems to first show the shadow of a 800x600 window (the resolution I selected), but then it grows back to its original size. Then, if I enter the menu and exit it, the window is rescaled properly.

If I add a delay between switching to windowed mode and changing the resolution, to give it the time to set up properly, the resolution of the window is correct. Maybe there is something to do to make sure that the fullscreen mode is correctly exited before changing resolution?

Lgt2x:
This is a very "RTFM" moment :smile: just had to sync the window. Fixed! I hope that does not break anything else. I'll do the 3-OS thorough testing tomorrow.

halprin (approved):
Everything works!

halprin:
Thanks for working out the Mac exclusive issues!  I super appreciate it.

PR #683, opened 2025-03-16
[MacOS] Brewfile: add missing 'pkg-config' dependency

Lgt2x:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

VCPKG will not install dependencies if missing

### DIFF
--- a/Brewfile
+++ b/Brewfile
@@ -2,3 +2,4 @@
 
 brew "cmake"
 brew "ninja"
+brew "pkg-config"


halprin (approved):
What does `pkg-config` get us?

Lgt2x:
It is a [collection of compiler metadata,](https://github.com/pkgconf/pkgconf) that can be used with CMake. VCPKG requires it to build the deps:
```
CMake Error at scripts/cmake/vcpkg_find_acquire_program.cmake:166 (message):
  Could not find pkg-config. Please install it via your package manager:

        brew install pkg-config
````

Something probably installs it automatically for the CI builds.

PR #686, opened 2025-03-22
use aligned version of vector for aligned_vector

jopadan:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [x] Other changes
    - [x] Math changes
 
### Description

- use portable align for aligned_vector and make it vector compatible

### Checklist


grtext/grtext.cpp
```diff
@@ -230,10 +230,13 @@ void grtext_Init() {
 }
 
 //	macro to get character width
+#undef CHAR_WIDTH
```
Lgt2x, line 233: where would that definition come from? If it is from a system include, it could be a good idea to change the name of our functions to something else instead

lib/vecmat_external.h
```diff
@@ -87,42 +97,50 @@ constexpr static inline const vector id(ssize_t i = -1)
...
-}
-constexpr static inline const aligned_vector ne()
-{
-  return aligned_vector{ (scalar)0, (scalar)0, (scalar)0 };
-}
-};
+// zero's out a vector
+static inline void vm_MakeZero(vector *v) { *v = vector::ne(); }
 
+using vector_array = vector;
```
Lgt2x, line 139: why do we need 2 names for the same thing?

Lgt2x (commented):
good job!

grtext/grtext.cpp
jopadan: See [Numeric limits](https://en.cppreference.com/w/c/types/limits) since C23
Yes, probably using #ifndef should be sufficient.

lib/vecmat_external.h
jopadan: vector_array is now combined with vector containing xyz and x,y,z and a typedef/use of vector to stay compatible with existing code that uses it.

grtext/grtext.cpp
jopadan: Probably renaming to GRCHAR_WIDTH should do the job

Descent3/terrainrender.cpp
```diff
@@ -2035,13 +2035,13 @@ void DrawSky(vector *veye, matrix *vorient) {
...
       tempvec = Terrain_sky.horizon_vectors[t][i];
-      tempvec.y -= veye->y * 0.5f;
+      tempvec.y() -= veye->y() * 0.5f;
       vm_MatrixMulVector(&Temp_sky_vectors[t][i], &tempvec, vorient);
 
       tempvec = Terrain_sky.horizon_vectors[t][i];
-      tempvec.x += veye->x;
-      tempvec.z += veye->z;
-      tempvec.y += veye->y * 0.5f;
+      tempvec.x() += veye->x();
```
Lgt2x, line 2042: why is `x` a functor now? This is all getting very complicated, is there a way to keep the same API as before?

lib/vecmat_external.h
```diff
@@ -43,118 +43,250 @@
...
+{
+	none     = 0 << 0,
+	scalar   = 1 << 1,
+	vector   = 2 << 1,
+	matrix   = 3 << 1,
+	adaptive = 4 << 1,
 };
-constexpr static inline angvec id(ssize_t i = -1)
+
+template<typename T, size_t N, enum align A = align::adaptive, size_t N_POW2 = VM_BIT_CEIL(N)>
```
Lgt2x, line 84: not following why we're doing all of that, what was the issue with what we had before your latest changes?

Descent3/terrainrender.cpp
jopadan: The only way to achieve it is to copy std::array implementation, add
```cpp
union {
      typename __array_traits<_Tp, _Nm>::_Type _M_elems;
      struct { T x,y,z,w };
      struct { T h,p,b };
};
```
and use modified array as a base class

jopadan: probably with requires or optional members

jopadan: Another possibility might be CLANG only Vector Extensions but they don't allow scalar aligned 3 component types
or CUDA types which allow 3 component scalar/vector aligned types.

lib/vecmat_external.h
jopadan: It allows types to be based on scalar/vector aligned std::array which allows use of [static extents/mdspan](https://en.cppreference.com/w/cpp/container/mdspan/extents/static_extent) and compatibility with other C++ container interfaces.

jopadan: Use of SIMD optimization for vector aligned types for -O3

tophyr:
Based on the title alone I am philosophically *super* on board, but this PR is going to be a bear to review. GitHub even chokes trying to render the diff.

This will also cause widescale merge issues with contributor repos. @Lgt2x, I know we've talked in the past about doing sweeping changes like these immediately-after big "cut" releases. Do we have one coming up that might be good to use? Otherwise, right now is actually seeming like kind of a good point to do this stuff, since we don't appear to have a ton of PRs or other activity inflight.

Descent3/terrainrender.cpp
Lgt2x: ok, the user side of our vector API is fine as it stands, we can keep it this way

jengelh:
```
constexpr inline vec<T,3,A> operator-() const     { vec<T,N,A> neg = {}; neg.fill((T)-1); return (*this) * neg; }
```
Is that a bug? Should it be <T,N,A> instead of <T,3,A>?

jengelh:
Ugh this commit also broke missile view, which is now rendered in a whacky position.

d32

Lgt2x:
how did you get this issue to occur?

PR #659, opened 2024-12-29
Add `DEFAULT_ADDITIONAL_DIRS` CMake option

Jayman2000:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Documentation or Wiki changes
- [x] Build and Dependency changes
- [x] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [x] Other changes

### Description

This pull request adds a `DEFAULT_ADDITIONAL_DIRS` CMake option. This new CMake option can be used by packagers in order to help ensure that Descent 3 automatically finds game files that were installed by one or more packages. See the commit message for details.

### Related Issues

In the pull request description for #471, I had promised that I would open three additional pull requests if #471 got merged. This is one of those pull requests.

### Checklist

- [x] I have tested my changes locally and verified that they work as intended.
- [x] I have documented any new or modified functionality.
- [x] I have reviewed the changes to ensure they do not introduce any unnecessary complexity or duplicate code.
- [x] I understand that by submitting this pull request, I am agreeing to license my contributions under the project's license.

PR #628, opened 2024-10-08
Enhance multiple base directory usage

winterheart:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Documentation or Wiki changes
- [x] Build and Dependency changes
- [x] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [x] Other changes

### Description

This is further enhancement of #471 feature.

Priority of base directory changed to following list (top is higher priority):

* Writable preference path
* User defined paths (cmd-line and configuration)
* Platform defined paths (such as `/usr/share/Descent3` on Linux or Steam installation paths on all platforms (TBD))
* Directory of executable

User defined path can be set by `-additionaldir` or `GAME_base_directory` in config file.

Added `ddio_GetPrefPath()` and `ddio_GetBasePath()` helper functions for building base directory entries.

Make `cf_LocatePathCaseInsensitiveHelper()` to iterate over all path items (previously it's only iterated only on filename of relative path).

Update functions related to writable base directory functionality: save / load games, demo writing / playback, logo and audiotaunts, screenshots.

Screenshots now are saved into `<writable base path> / screenshots` with timestamp in name.

If cmake project configured with `-DFORCE_PORTABLE_INSTALL=OFF`, Descent3 will search data files in platform defined path (`${CMAKE_INSTALL_DATADIR}/Descent3` or `/usr/share/Descent3`).

### Related Issues

Fixes #326
Fixes #373
Fixes #534

### Checklist

- [x] I have tested my changes locally and verified that they work as intended.
- [x] I have documented any new or modified functionality.
- [x] I have reviewed the changes to ensure they do not introduce any unnecessary complexity or duplicate code.
- [x] I understand that by submitting this pull request, I am agreeing to license my contributions under the project's license.

PR #691, opened 2025-03-25
Look for game data in usual Steam/GOG installation paths

Lgt2x:
## Pull Request Type

- [ ] GitHub Workflow changes
- [x] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [ ] Input changes
  - [ ] Network changes
  - [x] Other changes

### Description

Add standard platform paths to look for game data. This makes the game automatically look for not included proprietary game data in classic vendor installation location on all platforms. The objective is not to cover all possible cases, but make the experience seamless for the majority of people.

### Related Issues

Follow up to #628 , also related to #659 

cc @Jayman2000 @winterheart

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -194,10 +194,10 @@ find_package(plog REQUIRED)
 
 if(CMAKE_SYSTEM_NAME STREQUAL "Linux" OR CMAKE_SYSTEM_NAME STREQUAL "OpenBSD")
   message("Building for Linux")
-  add_compile_definitions(POSIX __LINUX__ _USE_OGL_ACTIVE_TEXTURES PRIMARY_HOG=\"d3-linux.hog\")
+  add_compile_definitions(POSIX __LINUX__ PRIMARY_HOG=\"d3-linux.hog\")
 elseif(CMAKE_SYSTEM_NAME STREQUAL "Darwin")
   message("Building for macOS")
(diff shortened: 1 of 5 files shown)


winterheart:
I was intended to implement some run-time discovery for installation paths of GOG / Steam since in user environment they may be literally anywhere. For example, mine Steam installation in Windows is placed in `C:\Games\Steam` directory. Hardcoded paths are OK, but there is much room to improvement.

Lgt2x:
I could not really think of a better approach than put the default installation path if the user keeps the defaults when installing steam of GOG, and hope for the best. Looking up online, most people seemed to also have games installed at the same location as I did on Windows, did you change the default (or use an older system)? 

And how would one dynamically discover steam path without scanning the whole drive, which I don't think is a good idea? 

If that works for some/most users that keep the defaults, it's already a valuable improvement in my opinion.

As a fallback, I plan to have a nice GUI using Sdl3 dialog (next PR) to pick up manually the installation path.

winterheart:
Well, at least for Windows we can read some Registry paths. Both Steam and GOG (and even legacy standalone installation) has own settings in Registry including installation path. For Linux and macOS it's little more complicated.

And here lies second caveat of hardcoded values: even in Steam with known installation we cannot predict where data actually is: user can add/change storage location for game data. So for Steam case we need make two actions: retrieve installation path and parse `libraryfolders.vdf` file (it's looks like JSON, but this is not a JSON, we need custom parser) in order to retrieve actual path. Yeah, that's heavy.

Lgt2x:
That indeed looks tedious, I'm not sure we want to go that far...

Issue #634, opened 2024-10-19
[Runtime Issue]: Mouse input code misses multiple mouse events per timestep leading to jerky mouselook


SiegeLord:
### Build Version

398699e04588cf154a189f783e5cde308f78bc99

### Operating System Environment

- [ ] Microsoft Windows (32-bit)
- [X] Microsoft Windows (64-bit)
- [ ] Mac OS X
- [ ] Linux (specify distribution and version below)

### CPU Environment

- [ ] x86 (32-bit Intel/AMD)
- [X] x86\_64 (64-bit Intel/AMD)
- [ ] ARM (32-bit)
- [ ] ARM64 (64-bit; sometimes called AArch64)
- [ ] Other (RISC V, PPC...)

### Game Modes Affected

- [X] Single player
- [X] Anarchy
- [X] Hyper-Anarchy
- [X] Robo-Anarchy
- [X] Team Anarchy
- [X] Capture the Flag
- [X] Bounty
- [X] Entropy
- [X] Hoard
- [X] Monsterball
- [X] Cooperative

### Description

Mouselook is very jerky regardless of sensitivity settings. The issue is that the code in lnxmouse.cpp sets the dx/dy values based on the xrel/yrel values, which discards the xrel/yrel values coming from earlier mouse events in the same timestep. Here's how I fixed it:

```diff
diff --git a/ddio/lnxmouse.cpp b/ddio/lnxmouse.cpp
index 558d2519..8afa54df 100644
--- a/ddio/lnxmouse.cpp
+++ b/ddio/lnxmouse.cpp
@@ -395,10 +395,10 @@ int sdlMouseMotionFilter(SDL_Event const *event) {
   } // if
   else {
     if (ddio_mouseGrabbed) {
-      DDIO_mouse_state.dx = event->motion.xrel;
-      DDIO_mouse_state.dy = event->motion.yrel;
-      DDIO_mouse_state.x += DDIO_mouse_state.dx;
-      DDIO_mouse_state.y += DDIO_mouse_state.dy;
+      DDIO_mouse_state.dx += event->motion.xrel;
+      DDIO_mouse_state.dy += event->motion.yrel;
+      DDIO_mouse_state.x += event->motion.xrel;
+      DDIO_mouse_state.y += event->motion.yrel;
     } // if
     else {
       DDIO_mouse_state.dx = event->motion.x - DDIO_mouse_state.x;
```

[...]

PR #695, opened 2025-04-12
Fix mouselook

tophyr:
## Pull Request Type

- [ ] GitHub Workflow changes
- [x] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [x] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

Likely with the change to SDL's 'captured mouse' mode, mouse movements during fullscreen became deltas *from the last movement*, not from the last sampled position or frame. In order to correctly measure the mouse movement, the deltas must be summed up and then reset when 'read' at input-processing time.

Moving to summing the deltas threw many other mouse-related calibration factors out of whack, so these factors are re-determined empirically across a range of mice, from 400dpi 125hz to 30,000dpi 8khz. Reasonable defaults are chosen targeting an 'average' mouse of 1000dpi, and the mouse sensitivity precision is increased to allow finer user tuning.

### Checklist

### DIFF
--- a/Descent3/ctlconfig.cpp
+++ b/Descent3/ctlconfig.cpp
@@ -303,7 +303,7 @@
 #define IDV_OPTIONS 18
 
 // used for adjusting settings.
-#define CFG_AXIS_SENS_RANGE 20
+#define CFG_AXIS_SENS_RANGE 200
 const int16_t UID_JOYCFG = 0x1000;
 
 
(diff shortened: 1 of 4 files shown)


tophyr:
Converting to draft... this breaks _non_ mouselook. Will fix tomorrow, shouldn't be hard.

tophyr:
Ok, this is ready.

Lgt2x (approved):
Good job, this is **much better** than it was before. Your fix looks correct to me. This breaks no mouse grab mode (enabled with `-nomousegrab` that I sometimes use for debug, but no problem addressing that later.

I don't seem to notice any difference depending on the game resolution as mentioned in https://github.com/DescentDevelopers/Descent3/issues/634#issuecomment-2440103762 either

PR #702, opened 2025-05-13
Improve -nomousegrab mouse input handling

Lgt2x:
## Pull Request Type

- [ ] GitHub Workflow changes
- [ ] Runtime changes
  - [ ] Render changes
  - [ ] Audio changes
  - [x] Input changes
  - [ ] Network changes
  - [ ] Other changes

### Description

Minor fixes to #695 (@tophyr ):
 - `event->motion.[x/y]rel` is float, so store `DDIO_mouse_state.dx` as float to avoid losing precision. This is especially needed when mouse is not grabbed by the window (-nomousegrab option) where xrel/yrel does not only have integer values.
 - Remove broken legacy code supporting `-nomousegrab` mode. Position is still not the same between the actual OS cursor and the in-game cursor, but it is better than it was before.

### DIFF
--- a/ddio/lnxmouse.cpp
+++ b/ddio/lnxmouse.cpp
@@ -95,7 +95,8 @@ static bool DDIO_mouse_init = false;
 static struct mses_state {
   int fd;                   // file descriptor of mouse
   int t, l, b, r;           // limit rectangle of absolute mouse coords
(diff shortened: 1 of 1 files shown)


winterheart:
Mouse handling on -nomousegrab in windowed mode much better, but still not accurate.

PR #705, opened 2025-05-21
Fix link failure

jengelh:
## Pull Request Type

### Description

Fixes

```
    /usr/include/c++/14/bits/unique_ptr.h:1077:(.text+0x7ae9): undefined reference to `httplib::Client::Client(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)'
    /usr/lib64/gcc/x86_64-suse-linux/14/../../../../x86_64-suse-linux/bin/ld: /usr/include/c++/14/bits/unique_ptr.h:93:(.text+0x7b0b): undefined reference to `httplib::Client::~Client()'
```

### Related Issues

#675 

### Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -180,6 +180,7 @@ endif()
 
 find_package(SDL3 REQUIRED CONFIG REQUIRED COMPONENTS SDL3)
 find_package(glm REQUIRED)
+find_package(httplib QUIET)
 if (TARGET httplib::httplib)
   add_library(httplib ALIAS httplib::httplib)
 else()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -186,9 +186,16 @@ if (TARGET httplib::httplib)
 else()
   # Some Linux distributions do not package httplibConfig.cmake, so find the header manually
   find_file(CPP_HTTPLIB_HEADER httplib.h REQUIRED)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -180,14 +180,17 @@ endif()
...
 find_package(SDL3 REQUIRED CONFIG REQUIRED COMPONENTS SDL3)
 find_package(glm REQUIRED)
-if (TARGET httlib::httplib)
+find_package(httplib QUIET)
+if (TARGET httplib::httplib)
   add_library(httplib ALIAS httplib::httplib)
 else()
   # Some Linux distributions do not package httplibConfig.cmake, so find the header manually
   find_file(CPP_HTTPLIB_HEADER httplib.h REQUIRED)
+  find_library(CPP_HTTPLIB_LIB cpp-httplib REQUIRED)
```
Lgt2x, line 189: Some distributions, for instance fedora, won't ship the shared library file (see https://packages.fedoraproject.org/pkgs/cpp-httplib/cpp-httplib-devel/fedora-42.html#files), because it is supposed to be header-only. This library packaging across distribution is a mess...

Lgt2x:
Lesson learned: never assume that "header-only" libraries are actually header-only :smile: 

Still builds correctly on debian:bookorm & Fedora 42 w/o VCPKG, thanks!

PR #708, opened 2025-05-21
Clang warning fixes

Lgt2x:
### Description

Resolve a bunch of clang warnings. The core of the leftover warnings are "unused-X" that need evaluation, char subscripts and C++ function signatures exported with C linkage.

For removed code, I kept relevant logic when it made sense, while keeping execution logic unchanged.


Descent3/GameCheat.cpp
```diff
@@ -523,7 +523,7 @@ void DemoCheats(int key) {
 
     int i;
     for (i = 0; i < MAX_PLAYER_WEAPONS; i++) {
-      if (HAS_FLAG(i)) {
+      if (HAS_FLAG(i) != 0) {
```
Lgt2x, line 526: looking at it again this test makes no sense, may as well remove it

PR #720, opened 2025-06-06
Resolve 160 instances of -Wchar-subscript

jengelh:
## Pull Request Type

### Description

`char` is unsigned on some platforms, which breaks the assumption it
can hold value -1. The simple fix is to replace it by int8_t, which
gives compatibility with x86 platforms where Descent3 usually was
delivered on.

```
Descent3/attach.cpp:668:47: warning: array subscript has type "char" [-Wchar-subscripts]
```

### Checklist

### DIFF
--- a/Descent3/AIGoal.cpp
+++ b/Descent3/AIGoal.cpp
@@ -463,6 +463,7 @@
  * $NoKeywords: $
  */
 
+#include <cstdint>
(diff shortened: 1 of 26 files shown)


Lgt2x (approved):
Thanks for taking care of that!

